PR #8, opened 2017-05-09
Cdt 9 2 Bug 516340 - Error "Could not create the hostShellProcess." Connection is not open

JawaharN:
I did some workaround here. My problem got solved. I did know this's clean solution or not.


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=516340

PR #17, opened 2020-11-03
Bug 568397: Prevent builder setup and end user entry(ies) mix

cartu38:
Some builder option(s) is(are) hold per dedicated builder setting(s).
As part of non default builder command, end user was able to override such option(s).
Was leading to jeopardized builder command plus some mismatches thinking UI promoted data(s).
Now get rid of end user entry(ies) if supported by dedicated builder option(s).
All such is applying today only to parallel build and stop on error options.

Change-Id: Icaefc80f0a5f08e64a2597461df7e3b10d316d50
Signed-off-by: cartu38 opendev <cartu38.opendev@gmail.com>


jonahgraham:
I believe this is the same change provided https://git.eclipse.org/r/c/cdt/org.eclipse.cdt/+/171735 - thanks for providing the change in gerrit.

PR #24, opened 2022-08-03
Add org.apache.commons.io to CDT's p2 site

### LINKED ISSUE #21: maven bundle for org.apache.commons.io has a different symbolic name
When trying to use a newer Eclipse SDK integration build with CDT 10.7, [plugindependencies](https://github.com/iloveeclipse/plugindependencies) detects the following problem:

```
Error: [org.eclipse.cdt.cmake.is.core 1.0.200.202106071842] plugin not found: org.apache.commons.io 2.6.0
```

This seems to be due to the ```Bundle-SymbolicName``` for the Maven bundle being different:

Orbit:
```
Automatic-Module-Name: org.apache.commons.io
Bundle-SymbolicName: org.apache.commons.io
```

Maven:
```
Automatic-Module-Name: org.apache.commons.io
Bundle-SymbolicName: org.apache.commons.commons-io
```

I assume this will lead to problems at runtime (we can't build with the error above, so we can't test).

Using the Maven bundle is coming from: https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/426

jonahgraham:
Fixes #21

### DIFF
--- a/releng/org.eclipse.cdt.repo/category.xml
+++ b/releng/org.eclipse.cdt.repo/category.xml
@@ -218,6 +218,7 @@
    <bundle id="org.yaml.snakeyaml" version="0.0.0"/>
    <bundle id="com.sun.jna" version="0.0.0"/>
    <bundle id="com.sun.jna.platform" version="0.0.0"/>
+   <bundle id="org.apache.commons.io" version="0.0.0"/>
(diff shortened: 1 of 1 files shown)


jonahgraham:
Verified on main branch with https://github.com/eclipse-cdt/cdt/pull/25

PR #34, opened 2022-08-11
Bug 580505 - Fix and improve CDT.setup

merks:
Don't install org.eclipse.pde.api.tools.ee.feature.feature.group because
it's no longer in the latest update site, and that causes problems like
this one as noticed by a user:

https://www.eclipse.org/forums/index.php?t=msg&th=1111320&goto=1854139&#msg_1854139

Move all the tasks up to the project level.

Include a * requirement on the CDT targlet so that all projects are
imported by the targlet task with no need for a separate import task.
Probably many of the explicitly listed requirements are no longer needed
because all requirements of all projects in the clone are resolved now
automatically.  Also ensure that all platforms are resolved so that all
of CDT fragments are imported.

Move the Git clone task before the modular targlet task so that the
clone is available during the first resolution.

Add an EclipseIni task so that the CDT.setup is redirected into the
local clone's version such that you can modify that setup and test it
locally via Help -> Perform Setup tasks (and so that Navigate -> Open
Setup -> CDT opens the editable workspace version).

https://bugs.eclipse.org/bugs/show_bug.cgi?id=580505

### DIFF
--- a/releng/CDT.setup
+++ b/releng/CDT.setup
@@ -7,18 +7,15 @@
     xmlns:jdt="http://www.eclipse.org/oomph/setup/jdt/1.0"
     xmlns:pde="http://www.eclipse.org/oomph/setup/pde/1.0"
     xmlns:predicates="http://www.eclipse.org/oomph/predicates/1.0"
-    xmlns:projects="http://www.eclipse.org/oomph/setup/projects/1.0"
     xmlns:setup="http://www.eclipse.org/oomph/setup/1.0"
     xmlns:setup.p2="http://www.eclipse.org/oomph/setup/p2/1.0"
     xmlns:setup.targlets="http://www.eclipse.org/oomph/setup/targlets/1.0"
     xmlns:setup.workingsets="http://www.eclipse.org/oomph/setup/workingsets/1.0"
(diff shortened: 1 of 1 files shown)


ruspl-afed:
> Move all the tasks up to the project level.

Is it a recommended way to formulate .setup?

> Include a * requirement on the CDT targlet so that all projects are
imported by the targlet task with no need for a separate import task.

For me an explicit import is better, but I don't have so deep knowledge of internals

merks:
> > Move all the tasks up to the project level.
> 
> Is it a recommended way to formulate .setup?
> 
> > Include a * requirement on the CDT targlet so that all projects are
> > imported by the targlet task with no need for a separate import task.
> 
> For me an explicit import is better, but I don't have so deep knowledge of internals

It only makes a difference if there were multiple streams and you wanted tasks to perform tasks for one specific but not the others.  But yes, unless there is a specific reason to put in the stream, better to put in the project.

The significant advantage of using the targlet to also do the import is that it actually ensures that all the requirements of all the projects are resolved into the target platform, so nothing can possibly be missing afterwards.  This also means that you don't have to list (and maintain in parallel) so many requirements because most are just duplicates of what's already in the projects.  But I did not do that simplification, yet.

The entire Eclipse Platform SDK's projects are all expressed in this way...

ruspl-afed (approved):
Thank you Ed @merks for improving CDT.setup

jonahgraham:
Thanks @merks I will test out and merge soon.

jonahgraham:
Brilliant, works great.

merks:
Here's an example from the Platform SDK's setups:

image

Here the first Tarlgets even have no requirements and so their source locator's projects will be resolved only for the IUs which are required by the platform projects.  So lots of wildcard requirements relative to source locators and very few explicit requirements, except for some high level root features that are desired to be always in the target platform even not available as a source project.

@jonahgraham 

Would you like a Configuration so that you can have a button like this?

https://github.com/eclipse-pde/.github/blob/main/CONTRIBUTING.md#create-an-eclipse-development-environment

jonahgraham:
> Would you like a Configuration so that you can have a button like this?

Yes please :+1:

PR #35, opened 2022-08-11
Test pipeline-github-plugin

jonahgraham:


jonahgraham:
The plug-in is installed now so we can add better/extra feedback from Jenkins to GitHub. Here is a simple example:

```
            script {
              if (env.CHANGE_ID) {
                // Use github-pipeline to interact with GitHub: https://github.com/jenkinsci/pipeline-github-plugin
                pullRequest.comment('The Jenkins build of this PR has has started. See details at ' + env.BUILD_URL)
              }
            }
```

PR #41, opened 2022-08-13
Update CONTRIBUTING.md

merks:
Add a badge-styled link to the CDT configuration and simplify the setup description.


merks:
I've added the badge-styled link to the configuration.  I've commented out the section with detailed steps because the configuration automates all those steps and because the page opened by the link also has all these instructions.

The description on that page has a description specific to CDT.

image

It comes from here:

https://github.com/eclipse-cdt/cdt/blob/455f7fc25a2567587d6091bf615741a484822db4/releng/CDTConfiguration.setup#L37-L49

Best to use the properties view to edit this content so that the XML is properly escaped.

image

You can preview what it looks like in the hover:

image

jonahgraham (approved):
LGTM - I also deleted the commented out stuff.

jonahgraham:
Thanks @merks

PR #46, opened 2022-08-16
Docker: Allow setting a mapping for docker paths

cmorty:
[Bug 579944](https://bugs.eclipse.org/bugs/show_bug.cgi?id=579944) introduced using Windows paths, to allow putting your project into the WSL's file system. 
While Docker for Windows (the commercial version) supports Windows paths, this does not work with a OSS-docker installed in the WSL. 
[See discussion in Bug 579944]] (https://bugs.eclipse.org/bugs/show_bug.cgi?id=579944#c5)
It is now possible to provide a path mapping to work around this.

### DIFF
--- a/launch/org.eclipse.cdt.docker.launcher/src/org/eclipse/cdt/internal/docker/launcher/ContainerPropertyTab.java
+++ b/launch/org.eclipse.cdt.docker.launcher/src/org/eclipse/cdt/internal/docker/launcher/ContainerPropertyTab.java
@@ -170,83 +170,97 @@ public class ContainerPropertyTab extends AbstractCBuildPropertyTab
 
 		usercomp.setLayoutData(gd);
 
-		enableButton = new Button(usercomp, SWT.CHECK);
-		enableButton.setText(Messages.ContainerPropertyTab_Enable_Msg);
+		//Enable button
+		{
+			enableButton = new Button(usercomp, SWT.CHECK);
+			enableButton.setText(Messages.ContainerPropertyTab_Enable_Msg);
 
-		iCfg = getCfg();
-		iCfgd = getResDesc().getConfiguration();
+			iCfg = getCfg();
+			iCfgd = getResDesc().getConfiguration();
 
-		gd = new GridData(GridData.FILL_HORIZONTAL);
-		gd.horizontalSpan = 5;
-		enableButton.setLayoutData(gd);
+			gd = new GridData(GridData.FILL_HORIZONTAL);
+			gd.horizontalSpan = 5;
+			enableButton.setLayoutData(gd);
+		}
+		// Connection
+		{
+			Label connectionSelectorLabel = new Label(usercomp, SWT.NULL);
+			connectionSelectorLabel.setText(Messages.ContainerTab_Connection_Selector_Label);
+			gd = new GridData(GridData.FILL_HORIZONTAL);
+			gd.horizontalSpan = 1;
+			gd.grabExcessHorizontalSpace = false;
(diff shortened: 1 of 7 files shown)


launch/org.eclipse.cdt.docker.launcher/src/org/eclipse/cdt/internal/docker/launcher/ContainerPropertyTab.java
```diff
@@ -170,83 +173,140 @@ public void createControls(Composite parent) {
 
 		usercomp.setLayoutData(gd);
 
-		enableButton = new Button(usercomp, SWT.CHECK);
-		enableButton.setText(Messages.ContainerPropertyTab_Enable_Msg);
+		//Enable button
+		{
```
jjohnstn, line 177: Not sure why you are breaking up the code with blocks.  Comments are ample.

jjohnstn (approved):
Just a minor comment about unnecessary blocks, but it should not block merging.  Don't know why the build keeps failing, but will keep trying.

akurtakov:
These tests are in unrelated plugins so ignore if you want this in 2022-09 release.

jonahgraham:
> These tests are in unrelated plugins so ignore if you want this in 2022-09 release.

BTW Current Main branch is for 2022-12 release. For 2022-09 we have already released [10.7.1](https://github.com/eclipse-cdt/cdt/releases/tag/CDT_10_7_1) - of course we can do a 10.7.2 if we want to backport things. Please let me know.

launch/org.eclipse.cdt.docker.launcher/src/org/eclipse/cdt/internal/docker/launcher/ContainerPropertyTab.java
jonahgraham: FWIW I am fond of these extra blocks, I would go further an locally declare stuff like gd in each block as that makes it easier to detect when a gd is being reused mistakenly (especially when stuff is copied/pasted in the future). However to me none of these things are blocking changes and as @cmorty is actively maintaining this, I am pleased to see it in a style he enjoys reading and that @jjohnstn has merged these changes.

jjohnstn:
I didn't know that 11.0.0 wasn't 2022-09.  The fix can't be back-ported as it does a major version bump in a plug-in so needs to be part of 11.0.0.

launch/org.eclipse.cdt.docker.launcher/src/org/eclipse/cdt/internal/docker/launcher/ContainerPropertyTab.java
cmorty: @jonahgraham : Yes. I added the blocks because I lost orientation. As for the gd: I thought about that, but then was to lazy. ;) I first had to learn, that I'm not able to reuse gd, which IMO is a bummer. :-)

jonahgraham: Thanks @cmorty for the comment. I too remember having to learn that you can't reuse gd.  The other way to do this is to not make a `gd` variable at all and use the factory, something like this:

```java
GridDataFactory.fillDefaults().grab(false, false).applyTo(label);
```

Its not really worth changing old code to that style, but it is useful to know for new code.

See [`GridDataFactory` javadoc](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/jface/layout/GridDataFactory.html) for more examples of how to use it.

PR #48, opened 2022-08-16
Fix widget is disposed error during shutdown

Torbjorn-Svensson:
After building a project and shuting down eclipse, the build console
will have a propertyChange event fired that will attempt to modify the
stream and background color that will trigger a use-after-dispose error
from SWT. This changeset ensures that the action is only taken on
non-disposed widgets.

Change-Id: I872c28da1b2fca7ec39f5b12d9d1ec398e559814


Torbjorn-Svensson:
@jonahgraham: I think this changeset is a candidate for 10.7.x

Torbjorn-Svensson:
This is an import of https://git.eclipse.org/r/c/cdt/org.eclipse.cdt/+/194492

jonahgraham (approved):
Yes, good for 10.7.1 too.

PR #51, opened 2022-08-16
Fix widget is disposed error during shutdown

Torbjorn-Svensson:
After building a project and shuting down eclipse, the build console
will have a propertyChange event fired that will attempt to modify the
stream and background color that will trigger a use-after-dispose error
from SWT. This changeset ensures that the action is only taken on
non-disposed widgets.


Torbjorn-Svensson:
This is a cherry-pick of #48.

jonahgraham (approved):

jonahgraham (approved):

PR #54, opened 2022-08-19
Fix some flaky managedbuilder tests

cwalther:
This fixes some tests that sometimes fail for me when I run them in the Eclipse GUI on Linux.

* CProjectDescriptionSerializationTests.testPersistentProperties contained a hardcoded sleep time to wait for an asynchronous operation that was often insufficient. Wait for the actual condition instead.

* BuiltinSpecsDetectorTest.testAbstractBuiltinSpecsDetector_EnvChangesGlobal did not properly clean up after itself, which caused it, and some of its siblings, to fail when run for the second time in the same session.

(Running tests twice in the same session is not done deliberately, but it happens in Eclipse with a JUnit Plug-in Test launch configuration set to run all tests in _org.eclipse.cdt.managedbuilder.core.tests/tests_ – once directly and once through _AllLanguageSettingsProvidersMBSTestSuite_. I have not found any way of running every test exactly once in the GUI – the [recommended](https://github.com/eclipse-cdt/cdt/blob/main/FAQ/README.md#how-do-i-run-cdt-junit-test-suite) way seems to be to select _org.eclipse.cdt.managedbuilder.tests.suite.AutomatedIntegrationSuite_, but that leaves out some tests.)

### DIFF
--- a/build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/language/settings/providers/tests/BuiltinSpecsDetectorTest.java
+++ b/build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/language/settings/providers/tests/BuiltinSpecsDetectorTest.java
@@ -851,6 +851,10 @@ public class BuiltinSpecsDetectorTest extends BaseTestCase {
 		assertEquals(true, provider.isExecuted());
 		assertEquals(ENV_SAMPLE_VALUE_2, provider.getSampleEnvVar());
 
+		// clean up
+		vars.deleteAll();
+		fUserSupplier.setWorkspaceEnvironment(vars);
+
 		// unregister listeners
 		provider.unregisterListener();
 	}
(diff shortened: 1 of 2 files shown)


jonahgraham (approved):

jonahgraham:
Thanks @cwalther for fixing some of these tests.

> The [recommended](https://github.com/eclipse-cdt/cdt/blob/main/FAQ/README.md#how-do-i-run-cdt-junit-test-suite) way seems to be to select org.eclipse.cdt.managedbuilder.tests.suite.AutomatedIntegrationSuite, but that leaves out some tests.)

I think this is out of date now - the suites are still in there but we now run (most) tests using the standard patterns in Tycho Surefire. Some of the bundles have exceptions like this one: https://github.com/eclipse-cdt/cdt/blob/0158c6aa6ef16cdae29283aa04cbf76a1aea50e9/core/org.eclipse.cdt.core.tests/pom.xml#L57-L64

I'll try an update to the readme.

cwalther:
Thanks! This is working much more conveniently than Bugzilla+Gerrit.

PR #58, opened 2022-08-22
Update FAQ instruction on running test suite

jonahgraham:
as a result of the conversation on PR #54 we identified the FAQ
on this item was out of date.

https://github.com/eclipse-cdt/cdt/pull/54#issuecomment-1222596313


jonahgraham:
@cwalther if you have thoughts you can review my update to the FAQ entry we discussed in #54.

jonahgraham:
Fully rendered link to preview before merge: https://github.com/jonahgraham/cdt/tree/run_junit_tests/FAQ#how-do-i-run-cdt-junit-test-suite

cwalther:
The update is fine, thanks. However, it does not really address the problem I was having, which was: Is there any way of obtaining a _JUnit Plug-in Test_ launch configuration in the GUI that will run exactly the same set of tests as Tycho Surefire would?

I failed to find one – if I select a suite class, that leaves out the tests that are not part of any suite, if I select a project or package that contains both tests and suites, the tests that are part of a suite will run twice because the launch configuration selects both the individual tests and the suites containing them again. I couldn’t find a way of specifying include patterns like the ones Tycho Surefire uses.

jonahgraham:
Sorry @cwalther - I don't know of anyway to do that. It sounds like a good feature request for eclipse-jdt to be able to do test discovery in that way (i.e. glob patterns).

PR #63, opened 2022-08-23
Improve dark theme on Windows

Torbjorn-Svensson:
Switch to CTabFolder/CTabItem for newui.
This changeset preserved binary API (as per tooling) but essentially
changes the meaning of some fields.

On a review of existing adopters/extenders of CDT no one is actually
accessing these fields.

Change-Id: Ic1ef0f242c1d2932726f3a6d4c9df9558312764a
Also-by: Jonah Graham <jonah@kichwacoders.com>


Torbjorn-Svensson:
This is an import of https://git.eclipse.org/r/c/cdt/org.eclipse.cdt/+/192656

Torbjorn-Svensson:
@Jonah, how do you want to proceed on this one?
Should we keep the old interface or should we just do the breaking change directly without caring about implementations that might extend the public interlace (that was never intended to be public...)?

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractPage.java
```diff
@@ -204,11 +206,13 @@ public abstract class AbstractPage extends PropertyPage implements IPreferencePa
 	 */
 	@Deprecated(forRemoval = true)
 	protected TabFolder folder;
```
jonahgraham, line 208: I think we can remove the `@Deprecated` fields now.

jonahgraham:
> @jonah, how do you want to proceed on this one? Should we keep the old interface or should we just do the breaking change directly without caring about implementations that might extend the public interlace (that was never intended to be public...)?

I think as we should just do the breaking change.

PS I am @jonahgraham on GitHub (not `@jonah`)

jonahgraham (changes requested):

jonahgraham (commented):
This looks good to me - please also update https://github.com/eclipse-cdt/cdt/blob/main/doc/org.eclipse.cdt.doc.isv/guide/deprecated_API_removals.html

Torbjorn-Svensson:
> I think as we should just do the breaking change.

Done in the last push.
I reverted to the old names to reduce the number of lines changed.

> PS I am @jonahgraham on GitHub (not `@jonah`)

Oops!

jonahgraham:
@Torbjorn-Svensson I added a couple of extra commits, one to remove the rest of the fields from API that we had previously announced and one to update the removals doc. Please review and let me know what you think.

jonahgraham:
:-( Sadly because you created this PR initially, I cannot add you as a reviewer. (BTW, note that I pushed my changes to the branch in your fork)

jonahgraham (approved):
@Torbjorn-Svensson assuming this passes the CI can you approve this.

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractPage.java
```diff
@@ -168,44 +168,22 @@ public abstract class AbstractPage extends PropertyPage implements IPreferencePa
...
+	 * @noreference This field is not intended to be referenced by clients.
 	 */
-	@Deprecated(forRemoval = true)
 	protected boolean noContentOnPage = false;
 	/**
-	 * @deprecated This field was never meant to be API. This field will be made private.
+	 * @noreference This field is not intended to be referenced by clients.
 	 */
-	@Deprecated(forRemoval = true)
 	protected boolean displayedConfig = false;
```
Torbjorn-Svensson, line 177: Shouldn't this field be private?

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractPage.java
```diff
@@ -168,44 +168,22 @@ public abstract class AbstractPage extends PropertyPage implements IPreferencePa
 	 * Bookeeping variables
 	 */
 	/**
-	 * @deprecated This field was never meant to be API. This field will be made private.
+	 * @noreference This field is not intended to be referenced by clients.
 	 */
-	@Deprecated(forRemoval = true)
 	protected boolean noContentOnPage = false;
```
Torbjorn-Svensson, line 173: Shouldn't this field be private?

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractPage.java
jonahgraham: Its accessed from AbstractSinglePage so I marked it as `@noreference` instead which gives us freedom to refactor in the future.

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractPage.java
jonahgraham: Its accessed from AbstractSinglePage so I marked it as `@noreference` instead which gives us freedom to refactor in the future.

Torbjorn-Svensson:
> @Torbjorn-Svensson assuming this passes the CI can you approve this.

The PR creator can't approve this the normal way, but LGTM.

PR #71, opened 2022-09-13
make the cmake error parser public API

15knots:


jonahgraham (commented):
LGTM - just missing `@since` tags for the API tooling. IIRC you had trouble getting that to work properly in the past, if that is still the case I can add the tags, let me know.

PR #72, opened 2022-09-13
make the cmake error parser public API

15knots:


jonahgraham (approved):
This PR looks a bit cleaner and won't need to be squashed. Thanks for recreating it.

15knots:
test failures do not come from changes in this PR.

PR #74, opened 2022-09-26
Remove old m2e connectors

jonahgraham:
With m2e upgrading to 2.0 in 2022-09, the old m2e connectors don't work anymore and generally aren't needed either.


jonahgraham:
@ruspl-afed - you are the resident Oomph expert in the CDT community. I have merged this change as it does lead to a working Eclipse CDT now that we are using 2022-09. But I don't know if there is a better way or something else that should be done instead. So if you have any thoughts post this merge I would be grateful to receive them.

PR #85, opened 2022-10-04
Remove double blanks from working set names

Bananeweizen:
Be aware this is a whitespace only change. 
image


jonahgraham (approved):
Thank you @Bananeweizen

PR #89, opened 2022-10-06
Fix broken working set references in the CDT.setup.

merks:
The previous "Remove double blanks from working set names" broke the references because there a space is encoded as %20.


jonahgraham (approved):

jonahgraham:
Thanks @merks

PR #90, opened 2022-10-07
Bug 580873: Remove error checking on dprintf parameters

jonahgraham:
The parser for dprintf parameters was much too simplistic andprevents real uses cases from operating.

It is not necessary to remove the , between the arguments to dprintf, so don't try to split on that anymore. That also means we can't check for and error on mismatch between format specifiers and number of arguments.

e.g. `"===> XML_EVENT_TEXT(%s)\n", (char *)strtok(Text,"\n")` should be permitted.

The alternative would be to write a more complete parser for both the printf specification and the arguments. Well out of scope and unnecessary.

Some inputs may now be accepted by the GUI and then when inserted with GDB fail. These errors are displayed in the GUI already anyway.


jonahgraham:
This is a draft because it needs:

- [ ] tests updated
- [ ] API notes updated due to removal of Messages from API
- [ ] N&N entry(?)

Issue #32, opened 2022-08-11
Migrate Eclipse CDT from Gerrit/Bugzilla/etc to GitHub


jonahgraham:
The Eclipse CDT project is moving to GitHub. We are using this issue to track all the items that need to be completed. Many of the items in this list need to be submitted to webmaster when we are ready at the [helpdesk issue](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/1562)

*This issue was migrated from https://github.com/eclipse-cdt/cdt-infra/issues/62 that had been used to track the move ahead of us having the new repo*

TODO list for the move:

- [x] Create https://github.com/eclipse-cdt/.github
  - Do not enable Issues, Discussions, Wiki - these will all be handled by the main repo https://github.com/eclipse-cdt/cdt
- [x] Create a minimal README.md file in eclipse-cdt/.github that redirects to eclipse-cdt/cdt
- [x] Archive https://github.com/eclipse-cdt/cdt-new-managedbuild-prototype
- [x] Archive https://github.com/eclipse-cdt/cdt-vscode
- [x] Migrate CDT's git repo stored in gerrit to https://github.com/eclipse-cdt/cdt
  - This may be as simple as stopping mirroring https://git.eclipse.org/r/admin/repos/cdt/org.eclipse.cdt,general?
- [x] Rename `master` branch to `main`
- [x] Delete/disable https://github.com/eclipse-cdt/cdt wiki
  - This was never used (bar a single/simple page)
  - Proposal/intent is that all documentation is part of the main repo, especially New and Noteworthy
  - See item farther down about future of the Wiki
- [x] Enable *Issues* and *Discussions* on https://github.com/eclipse-cdt/cdt 
- [x] Set description on https://github.com/eclipse-cdt/cdt to `Eclipse CDT™ C/C++ Development Tools`
- [x] Make https://git.eclipse.org/r/admin/repos/cdt/org.eclipse.cdt,general readonly
  - There is a lot of stuff on CDT's gerrit that is of differing value. My preferred option to leave CDT's gerrit in read-only state until gerrit is shutdown. That should give plenty of time for people to extract anything of value from gerrit and contribute it to CDT (as a GitHub PR)
- [x] Protect branches from force pushes, deletions and direct pushes
  - All pushes need to be via PR.
- [x] Migrate Process, Policy, etc from Wiki to README.md/CONTRIBUTING.md
- [x] Migrate FAQs and sample code to markdown
[...]

PR #97, opened 2022-10-10
DSF-GDB tests in GitHub Actions

jonahgraham:
- Fix the running so that the correct gdb is used.
- Only run DSF tests if dsf directories have been modified
- Default to using gdb,gdbserver on the PATH when running tests
- Remove double reporting of test results

Part of #32

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -15,12 +15,27 @@ jobs:
     - uses: actions/checkout@v3
       with:
         fetch-depth: 0
+    - uses: dorny/paths-filter@v2
+      id: filter
+      with:
+        filters: |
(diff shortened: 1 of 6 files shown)


jonahgraham:
This looks like it is working fine.

- #98 which modifies dsf related code takes much longer to run and more tests are run: https://github.com/eclipse-cdt/cdt/pull/98#issuecomment-1273368868
- #99 which does not modify dsf related code is much quicker and runs less tests: https://github.com/eclipse-cdt/cdt/pull/99#issuecomment-1273321064

jonahgraham:
PS. In case you are wondering, when the full build runs on Jenkins the DSF tests always run.

PR #103, opened 2022-10-16
Bug 580314 - Remove deprecated methods from core build

mbooth101:
Mark deprecated methods in CommandLauncher and CBuildConfig as "for removal"

Moved this PR over from gerrit: https://git.eclipse.org/r/c/cdt/org.eclipse.cdt/+/194515

### DIFF
--- a/NewAndNoteworthy/CHANGELOG-API.md
+++ b/NewAndNoteworthy/CHANGELOG-API.md
@@ -10,6 +10,8 @@ This section describes API removals that occurred in past releases, and upcoming
 
 - [org.eclipse.cdt.ui.newui.AbstractPage reduced visibility of many fields](#newUIAbstractPage)
 - [org.eclipse.cdt.dsf.gdb.breakpoints.Messages is no longer API](#org.eclipse.cdt.dsf.gdb.breakpoints.Messages)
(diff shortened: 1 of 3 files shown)


mbooth101:
Hi @jonahgraham I had originally targeted this change for an earlier release of CDT, but it looks like CDT will get a major version bump now -- should we just remove these deprecated methods instead?

jonahgraham:
My default answer is yes. If you think the api is widely used in the wild it may be nicer to just deprecate though.

jonahgraham:
> My default answer is yes. If you think the api is widely used in the wild it may be nicer to just deprecate though.

I have had a chance to look at which methods these are. Please just delete them and document the deletion (move the doc you already wrote to the correct section).

mbooth101:
Hmm, unusual amount of test failures. Will have to check if related to this change

jonahgraham:
Looking at the change, it can't possibly be related or my sanity would be seriously affected. It would be great to track down some of these intermittent test failures (using this PR if that helps), or feel free to merge this one as is. This is part of why I haven't turned off the Jenkins CI builds yet. Until we get stable builds again regularly I am happy to automatically have multiple runs.

mbooth101:
Understood, thanks.

PR #109, opened 2022-10-19
Add sanity checking and doc to do_format_code.sh

cmorty:
This should lower the hurdle to make the CI happy.


jonahgraham (approved):
Looks great. Thanks.

Once CI says something positive I will merge it.

PR #111, opened 2022-10-21
Do not overwrite custom build system ID with MBS ID

15knots:


15knots:
I'm unsure, whether this causes the BuildDescriptionModelTests.testDesRebuildState to fail.
Should I revert it temporarily and see if the test passes then?

jonahgraham (approved):
> I'm unsure, whether this causes the BuildDescriptionModelTests.testDesRebuildState to fail. Should I revert it temporarily and see if the test passes then?

That test is flaky, it passed on some of the runs.

Issue #81, opened 2022-10-03
Deadlock in indexer


jonahgraham:
Something in the recent changes to #79 has caused the indexer to deadlock more readily than before.

Running PDOMCPPBugsTest on the build machine locks up, and it does when running (but not debugging) a significant percentage of the time.

jonahgraham:
I have had two runs now with identical stack traces, so pretty sure this is what is going on.

Extracts of key threads:

```java
"main" #1 prio=6 os_prio=0 cpu=6380.63ms elapsed=721.72s tid=0x00007f29ec0267b0 nid=0x1eaa7c waiting for monitor entry  [0x00007f29f128e000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at java.lang.Object.wait(java.base@17.0.3/Native Method)
	- waiting on <no object reference available>
	at org.eclipse.cdt.internal.core.pdom.PDOMManager.joinIndexer(PDOMManager.java:1173)
	- locked <0x00000000e8e07aa0> (a [Z)
	at org.eclipse.cdt.core.testplugin.util.BaseTestCase5.waitForIndexer(BaseTestCase5.java:165)
	at org.eclipse.cdt.core.testplugin.util.BaseTestCase.waitForIndexer(BaseTestCase.java:234)
	at org.eclipse.cdt.internal.pdom.tests.PDOMCPPBugsTest.test191679(PDOMCPPBugsTest.java:227)


"Worker-5: Update Monitor" #56 prio=5 os_prio=0 cpu=19.67ms elapsed=712.00s tid=0x00007f2848000e70 nid=0x1eab17 waiting for monitor entry  [0x00007f2878487000]
   java.lang.Thread.State: BLOCKED (on object monitor)
	at org.eclipse.cdt.internal.core.pdom.PDOMManager.isIndexerIdle(PDOMManager.java:751)
	- waiting to lock <0x00000000e2035f38> (a java.util.ArrayDeque)
	at org.eclipse.cdt.internal.core.pdom.PDOMManager$7.done(PDOMManager.java:1153)
	- locked <0x00000000e8e07aa0> (a [Z)
	at org.eclipse.core.internal.jobs.JobListeners$$Lambda$127/0x0000000800da52f0.notify(Unknown Source)

[...]

PR #113, opened 2022-10-24
Fix build failure in standalone debugger application

mbooth101:
Does anybody actually use this?

We are clearly no longer routinely building it. If the project is not interested in maintaining it, maybe it can be deleted...

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,6 +33,7 @@ pipeline {
                       -Ddsf.gdb.tests.timeout.multiplier=50 \
                       -Dindexer.timeout=300 \
                       -P production \
+                      -P build-standalone-debugger-rcp \
                       -Dmaven.repo.local=/home/jenkins/.m2/repository \
                       --settings /home/jenkins/.m2/settings.xml \
                       "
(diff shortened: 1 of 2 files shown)


jonahgraham:
It is still (slightly) used, last I checked about this ([when we stopped shipping it](https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CDT-10.2.md#cdt-standalone-debugger-part-of-main-download)) is that RedHat were  making packages out of it. We have bug reports about it as recent as July - https://bugs.eclipse.org/bugs/show_bug.cgi?id=580365

That said I would be more than happy to remove it in CDT 11. 

Note that the .product file generates the config.ini that is used here: https://wiki.eclipse.org/CDT/StandaloneDebugger 

BTW Where did you see it fail? Local build? I think the Jenkinsfile should add `-P build-standalone-debugger-rcp` - at least that makes sure it builds with current cdt.target defined Eclipse Platform version.

mbooth101:
Yes, I was trying to build it locally, e.g:
```
$ JAVA_HOME=/usr/lib/jvm/java-17 mvn clean install -DskipTests -Pbuild-standalone-debugger-rcp --projects :org.eclipse.cdt.debug.application.product
```

jonahgraham (approved):
I approve this change - however note that we don't use 4.26 platform yet because of #81 - which means that even though it may build, it won't necessarily work in 2022-12.

PR #116, opened 2022-10-25
Bump org.eclipse.cdt.core major version

jonahgraham:
Due to the removal of previous API done in:

- PR #103
- commit 4de9516e97d7a347b44a665532284416049d2ba8


core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
```diff
@@ -518,15 +518,15 @@ public Process startBuildProcess(List<String> commands, IEnvironmentVariable[] e
...
+	 * @since 8.0
 	 */
 	protected int watchProcess(IConsole console, IProgressMonitor monitor) throws CoreException {
 		return launcher.waitAndRead(console.getInfoStream(), console.getErrorStream(), monitor);
 	}
 
 	/**
 	 * @return The exit code of the build process.
-	 * @since 7.5
+	 * @since 8.0
```
mbooth101, line 529: Is it right to change these since tags?

jonahgraham: AFAICT this new API hasn't actually been released yet, that is why it bumped.

mbooth101: Ah, fair enough :-)

Issue #117, opened 2022-10-25
Test stability umbrella bug


jonahgraham:
This is a list of known unstable tests since we moved to Java 17 + GitHub Actions.

If you have a failing test in your GitHub actions, please check it against this list. If the test looks unstable, add it to this list (or make a comment requesting a committer do that).

- [x] [MIBreakpointsTest.insertDprintfBreakpoint[gdbserver]](https://ci.eclipse.org/cdt/job/cdt/view/change-requests/job/PR-106-merge/lastCompletedBuild/testReport/org.eclipse.cdt.tests.dsf.gdb.tests/MIBreakpointsTest/insertDprintfBreakpoint_gdbserver_/) #115
- [x] [LanguageSettingsListenersTests.testNotification_globalProvider_SerializeEntries](https://ci.eclipse.org/cdt/job/cdt/job/main/60/testReport/junit/org.eclipse.cdt.core.language.settings.providers/LanguageSettingsListenersTests/testNotification_globalProvider_SerializeEntries/) #118
- [x] [ResourceLookupTests.testResourceDelta](https://ci.eclipse.org/cdt/job/cdt/job/main/60/testReport/junit/org.eclipse.cdt.core.internal.tests/ResourceLookupTests/testResourceDelta/) #120 
- [x] [GDBMultiNonStopRunControlTest.testSuspendProcessTwoThreadsRunning[gdbserver]](https://ci.eclipse.org/cdt/job/cdt/view/change-requests/job/PR-106-merge/lastCompletedBuild/testReport/org.eclipse.cdt.tests.dsf.gdb.tests.nonstop/GDBMultiNonStopRunControlTest/testSuspendProcessTwoThreadsRunning_gdbserver_/) #119 
- [x] [MIExpressionsNonStopTest.testLocalVariables[gdbserver]](https://ci.eclipse.org/cdt/job/cdt/view/change-requests/job/PR-106-merge/lastCompletedBuild/testReport/org.eclipse.cdt.tests.dsf.gdb.tests.nonstop/MIExpressionsNonStopTest/testLocalVariables_gdbserver_/) Haven't seen this fail in a while and have now lost the stack trace of the last fail
[...]

PR #127, opened 2022-10-29
Add ManagedBuildManager#createConfigurationForProject()

15knots:
This should allow ISV's to create MBS based project with a vendor-specific build-system ID without using internal API.

### DIFF
--- a/build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/ManagedBuildManager.java
+++ b/build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/ManagedBuildManager.java
@@ -153,14 +153,16 @@ import org.w3c.dom.NodeList;
 import org.w3c.dom.ProcessingInstruction;
 
 /**
- * This is the main entry point for getting at the build information
- * for the managed build system.
+ * This is the main entry point for getting at the build information for the
+ * managed build system.
  *
  * @noextend This class is not intended to be subclassed by clients.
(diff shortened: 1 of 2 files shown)


jonahgraham:
I added this unrelated test failure to #117 as:

- #128

jonahgraham (changes requested):
As this involves new public API that is useful for ISVs, please include a N&N entry in [CDT 11's N&N](https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CDT-11.0.md)

---

Can you please update this to not have all the unrelated comment changes please. This change is >1500 lines of code change as it is now, causing GitHub not to render diff by default and it is not easy to spot what code changes have been made.

Initially when we formatted all the code we left comments as do not format. It looks like you are using some custom formatter settings (as opposed to the project settings) which causes formatting of unrelated code.

I am open to changing the CDT defaults to include comment formatting, but I want to handle that in a separate commit/PR please.

15knots:
Sorry for the noise. The formatting rules of CDT now give me 20 unrelated changes.
Any of them is a empty line in comment-out code. These lines now are formatted as '	 * '.
I hope that will be OK.

jonahgraham:
20 lines is fine for me, we can see those changes easily enough. Not sure why the files are having any changes though, but as JDT versions change the formatting sometimes changes even though settings aren't changed. If we are seeing a lot of this I can deal with it down the line.

15knots:
Ah, again this weird reformatting.
Ok installing the IDE again using OOMPH didn't the trick.
Guess I should abandon this and restart with a fresh clone

Issue #128, opened 2022-10-29
Need to hold a write lock to clear result caches assertion error


jonahgraham:
This test, or one similar to it, has failed a couple of times recently. 

This is the detailed output:

```
Expected number (0) of Non-OK status objects in log differs from actual (1).
 Error while parsing /projC_testTripleLinear/h3.h. java.lang.reflect.InvocationTargetException
junit.framework.AssertionFailedError: 
Expected number (0) of Non-OK status objects in log differs from actual (1).
	Error while parsing /projC_testTripleLinear/h3.h. java.lang.reflect.InvocationTargetException

Caused by: java.lang.reflect.InvocationTargetException
Caused by: java.lang.AssertionError: Need to hold a write lock to clear result caches
```

The stack trace is incomplete on this, but the assertion is from:

https://github.com/eclipse-cdt/cdt/blob/10e95a182506740eabb9d4298ae2082ac9af15c5/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/index/WritableCIndex.java#L159-L163

jonahgraham:
testModelBuilderBug262785 (org.eclipse.cdt.core.model.tests.CModelBuilderBugsTest) failed with the same assertion error.

jonahgraham:
testTripleLinear (org.eclipse.cdt.internal.index.tests.IndexCompositeTests) failed  was the original failure

jonahgraham:
testModelBuilderBug274490 (org.eclipse.cdt.core.model.tests.CModelBuilderBugsTest) failed for the same reason.

jonahgraham:
And another: testModelBuilderBug222398 (org.eclipse.cdt.core.model.tests.CModelBuilderBugsTest)

jonahgraham:
<details>
<summary>Here is a trace of the test failing showing all the acquiring and releasing of the locks</summary>

```java
Expected number (0) of Non-OK status objects in log differs from actual (1).
 Error while parsing /testPreprocessorNodes/DeclaratorsTests.cpp. java.lang.reflect.InvocationTargetException
Stack Trace
java.lang.reflect.InvocationTargetException
 at org.eclipse.cdt.internal.core.pdom.AbstractIndexerTask.getMaskedException(AbstractIndexerTask.java:1206)
 at org.eclipse.cdt.internal.core.pdom.AbstractIndexerTask.swallowError(AbstractIndexerTask.java:1192)
[...]

PR #132, opened 2022-10-31
Cleaning project fails once the binary is expanded in project explorer on Windows

umairsair:
Steps:
======
1. Create a managed project and build it
2. Expand the built binary available in binary container in project explorer view
3. Now clean the project, clean will fail irrespective of number of tries you do

Reason:
=======
For finding the sources for binary, Elf instance is created and `Section.mapSectionData(..)` creates MappedByteBuffer of channel which locks the file on Windows until its garbage collected, see following
https://bugs.java.com/bugdatabase/view_bug.do?bug_id=4715154

Solution:
=========
Made ISymbolReader AutoCloseable and user is responsible to	properly close it. In case of dwarf reader, we remove all the references of ByteBuffer and call gc.

---

This solution extends work that was done in [Bug 553674](https://bugs.eclipse.org/bugs/show_bug.cgi?id=553674) to make more classes AutoCloseable.

There are a couple of things left to do:

- [x] Directly delete the deprecated method 
- [x] Document breaking API changes in https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CHANGELOG-API.md
- [x] Document API improvement (using autocloseable) in https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CDT-11.0.md#api-changes-current-and-planned

### DIFF
--- a/core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/elf/Elf.java
+++ b/core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/elf/Elf.java
@@ -59,7 +59,6 @@ public class Elf implements AutoCloseable {
 	private Symbol[] symbolsTable;
 	/** .dynSym section */
 	private Symbol[] dynamicSymbols;
-	private boolean areSectionsMapped; // Have sections been mapped? Used to clean up properly in Elf.Dispose.
 
 	protected String EMPTY_STRING = ""; //$NON-NLS-1$
 	private long elfOffset;
@@ -339,7 +338,6 @@ public class Elf implements AutoCloseable {
 		 */
 		public ByteBuffer mapSectionData() throws IOException {
 			makeSureNotCompressed();
-			areSectionsMapped = true;
 			return efile.getChannel().map(MapMode.READ_ONLY, sh_offset, sh_size).load().asReadOnlyBuffer();
 		}
 
@@ -987,9 +985,6 @@ public class Elf implements AutoCloseable {
 			if (efile != null) {
 				efile.close();
 				efile = null;
-				// ensure the mappings get cleaned up
-				if (areSectionsMapped)
-					System.gc();
 			}
 		} catch (IOException e) {
 		}
@@ -1304,6 +1299,12 @@ public class Elf implements AutoCloseable {
 		return reader;
 	}
 
+	/**
+	 * Creates a new symbol reader instance on each call. Caller is responsible for closing
+	 * the symbol reader
+	 *
+	 * @return symbol reader or {@code null} if couldn't create symbol reader
+	 */
 	public ISymbolReader getSymbolReader() {
 		ISymbolReader reader = null;
 		reader = createDwarfReader();
--- a/NewAndNoteworthy/CDT-11.0.md
+++ b/NewAndNoteworthy/CDT-11.0.md
@@ -55,6 +55,10 @@ It may have other uses to other API consumers as well and is therefore included 
 
 This should allow ISV's to create MBS based project with a vendor-specific build-system ID without using internal API.
 
+## Binary Parser code uses AutoCloseable
+
+The binary parser classes which open binary files now implement AutoCloseable so they can (and should) be used in a try-with-resources block.
+See https://github.com/eclipse-cdt/cdt/pull/132
 
 # Bugs Fixed in this Release
 
--- a/NewAndNoteworthy/CHANGELOG-API.md
+++ b/NewAndNoteworthy/CHANGELOG-API.md
@@ -14,6 +14,7 @@ This section describes API removals that occurred in past releases, and upcoming
 - [Removal of deprecated CBuildConfiguration.watchProcess() methods](#watchProcessCBuildConfig)
 - [Rework of API to determine GDB command line in org.eclipse.cdt.dsf.gdb](#gdbBackendDebuggerCommandLine)
 - [Removal of Qt plug-ins and features](#qt-plugins)
+- [Removal of constructor org.eclipse.cdt.utils.coff.CodeViewReader(RandomAccessFile, int, boolean)](#CodeViewReader-constructor-removal)
 
 ## API Changes in CDT 10.5.0
 
@@ -162,6 +163,14 @@ The following bundles and all their related API has been removed:
 - org.eclipse.cdt.qt.core
 
 See https://github.com/eclipse-cdt/cdt/issues/123
+
+### <span id="CodeViewReader-constructor-removal">Removal of constructor org.eclipse.cdt.utils.coff.CodeViewReader(RandomAccessFile, int, boolean)</span>
+
+Same instance of RandomAccessFile was shared between multiple objects which
+causes problems in closing it properly. A new constructor is introduced which
+accepts filename and opens a RandomAccessFile.
+
+See https://github.com/eclipse-cdt/cdt/pull/132
 
 ---
 
--- a/core/org.eclipse.cdt.core/model/org/eclipse/cdt/internal/core/model/Binary.java
+++ b/core/org.eclipse.cdt.core/model/org/eclipse/cdt/internal/core/model/Binary.java
@@ -306,11 +306,14 @@ public class Binary extends Openable implements IBinary {
 		// Try to get the list of source files used to build the binary from the
 		// symbol information.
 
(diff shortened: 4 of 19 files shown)


jonahgraham:
Please disregard the `testModelBuilderBug274490 (org.eclipse.cdt.core.model.tests.CModelBuilderBugsTest)` failure. I'll add it to #117, it looks like #128

core/org.eclipse.cdt.core/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.cdt.core; singleton:=true
-Bundle-Version: 8.0.0.qualifier
+Bundle-Version: 8.0.100.qualifier
```
jonahgraham, line 5: The version has already been bumped since the last release, so this version bump isn't needed.

core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/elf/Elf.java
```diff
@@ -340,7 +339,9 @@ public class Section {
 		public ByteBuffer mapSectionData() throws IOException {
 			makeSureNotCompressed();
 			areSectionsMapped = true;
-			return efile.getChannel().map(MapMode.READ_ONLY, sh_offset, sh_size).load().asReadOnlyBuffer();
+			ByteBuffer buff = ByteBuffer.allocateDirect((int) sh_size);
```
jonahgraham, line 342: If this is needed, I think that you shouldn't be changing this method, instead have the caller call loadSectionData, or possibly create a new method that returns a ByteBuffer.

Ideally the same fix should be applied to the other readers that have the same issue, e.g. Coff/Coff64

jonahgraham (commented):
I don't think overall this is the correct solution. Mapping was explicitly used with the comment "use mapping of sections for dwarf parser rather than allocating memory which avoids out of memory exceptions for large symbol files." in 7ef8d2899113533c8b9d1c8fd54b1177e1447961

Can you comment if that is no longer relevant?

As for the the garbage collection, I think that is why 7ef8d2899113533c8b9d1c8fd54b1177e1447961 calls System.gc in dispose method.

core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/elf/Elf.java
umairsair: I am trying to understand why do we need new method.. current and previous approach both uses `DirectByteBuffer` so ideally there should be no issue..

I'll checked other readers as well

umairsair:
> I don't think overall this is the correct solution. Mapping was explicitly used with the comment "use mapping of sections for dwarf parser rather than allocating memory which avoids out of memory exceptions for large symbol files." in [7ef8d28](https://github.com/eclipse-cdt/cdt/commit/7ef8d2899113533c8b9d1c8fd54b1177e1447961)
> 
> Can you comment if that is no longer relevant?

I think that both current and previous solution are same as both eventually gives the `DirectByteBuffer`.. maybe there is something deeper in the `DirectByteBuffer` implementation for both cases which I am missing? Apparently I see no issue..

> As for the the garbage collection, I think that is why [7ef8d28](https://github.com/eclipse-cdt/cdt/commit/7ef8d2899113533c8b9d1c8fd54b1177e1447961) calls System.gc in dispose method.

IMO this logic is not valid for all scenarios.. it is valid when its only called form `finalize`.. but its called from `close` as well as by the user which essentially doesn't mean that all the references of it are gone including the ByteBuffer instance references, so this gc doesn't work for scenarios where `dispose` is called from `close` or by user..

jonahgraham:
> > I don't think overall this is the correct solution. Mapping was explicitly used with the comment "use mapping of sections for dwarf parser rather than allocating memory which avoids out of memory exceptions for large symbol files." in [7ef8d28](https://github.com/eclipse-cdt/cdt/commit/7ef8d2899113533c8b9d1c8fd54b1177e1447961)
> > Can you comment if that is no longer relevant?
> 
> I think that both current and previous solution are same as both eventually gives the `DirectByteBuffer`.. maybe there is something deeper in the `DirectByteBuffer` implementation for both cases which I am missing? Apparently I see no issue..

Try a big file, you can map it without running out of memory, but you can't make a large ByteBuffer.

Here is a simple example:

- Create a large file (> Xmx size provided later)
- Use a program like this:

```java
package org.eclipse.cdt.utils.elf;

import java.io.IOException;
import java.nio.ByteBuffer;
import java.nio.channels.FileChannel.MapMode;

import org.eclipse.cdt.utils.ERandomAccessFile;

public class Main {
	public static void main(String[] args) throws IOException {
		System.out.println("Opening " + args[0]);
		ERandomAccessFile efile = new ERandomAccessFile(args[0], "r"); //$NON-NLS-1$
		long size = efile.getChannel().size();
		long sizeToRead = Math.min(Integer.MAX_VALUE / 2, size);
		ByteBuffer byteBuffer = efile.getChannel().map(MapMode.READ_ONLY, 0, sizeToRead).load().asReadOnlyBuffer();
		System.out.println("Type of byteBuffer " + byteBuffer.getClass());
		// This line will fail:
		ByteBuffer buff = ByteBuffer.allocateDirect((int) sizeToRead);
		System.out.println("Type of byteBuffer " + byteBuffer.getClass());
	}
}
```

Run it with an -Xmx smaller than `sizeToRead` and you will see the `allocateDirect` will fail with:

```
Exception in thread "main" java.lang.OutOfMemoryError: Cannot reserve 1073741823 bytes of direct buffer memory (allocated: 0, limit: 536870912)
```


> 
> > As for the the garbage collection, I think that is why [7ef8d28](https://github.com/eclipse-cdt/cdt/commit/7ef8d2899113533c8b9d1c8fd54b1177e1447961) calls System.gc in dispose method.
> 
> IMO this logic is not valid for all scenarios.. it is valid when its only called form `finalize`.. but its called from `close` as well as by the user which essentially doesn't mean that all the references of it are gone including the ByteBuffer instance references, so this gc doesn't work for scenarios where `dispose` is called from `close` or by user..

I agree - and this needs to be fixed, but not by allocating *everything* into memory. It may take some more judicious use of allocated memory so the handle can be freed and mapped buffers where that is needed. 

The other alternative is to fix the whole lifecycle so that the useful stuff is loaded out of the file, then the file is closed. The actually needed data is going to be much smaller than the whole section needing to be in memory.

Because this release is a new major release it is a good opportunity to make such drastic changes.

umairsair:
Thanks for the guidance @jonahgraham 

I have added a commit where I have made ISymbolReader AutoCloseable and user is responsible for closing it properly.

And if user calls `mapSectionData`, it'll be his responsibility to make sure that after using ByteBuffer he properly un-reference it so that its garbage collected.. let me know your feedback..

jonahgraham:
> @[umairsair](https://github.com/umairsair) umairsair requested a review from [jonahgraham](https://github.com/jonahgraham) [1 hour ago](https://github.com/eclipse-cdt/cdt/pull/132#event-7725882765)

@umairsair I'll try to review properly soon, on first pass this looks correct.

jonahgraham (commented):
This looks fine, thanks for the change.

I am going to merge https://github.com/eclipse-cdt/cdt/pull/135 first so we don't spend even more time reviewing code in deprecated classes.

(moved TODO list so it shows up in GitHub UI better)

jonahgraham:
(Does this show up properly if I add it here)


There are a couple of things left to do:

- [ ] Directly delete the deprecated method 
- [ ] Document breaking API changes in https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CHANGELOG-API.md
- [ ] Document API improvement (using autocloseable) in https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CDT-11.0.md#api-changes-current-and-planned

jonahgraham:
#135 was taking too long to resolve, so I have rebased this change, added the N&N entry and once I get a clean build I will merge it.

jonahgraham (approved):

jonahgraham:
Thank you @umairsair for the improvement.

PR #135, opened 2022-11-03
Remove deprecated binary parsers and supporting code

jonahgraham:
These binary parsers have been slated for deleting for a while and are replaced with 64-bit compatible
versions.

Some methods still refereneced the 32-bit variants and have been updated to the fully functioning
64-bit variant.

See also [Bug 562495](https://bugs.eclipse.org/bugs/show_bug.cgi?id=562495)

TODO before merging

- [x] Remove references in plugin.xml
- [x] Test what happens if a project has the old parser referenced and they upgrade to it gone, if there is user visible effect, document in N&N
- [x] Merge in #137


mbooth101:
You also have to update the plugin.xml too, right? https://github.com/eclipse-cdt/cdt/blob/main/core/org.eclipse.cdt.core/plugin.xml#L75

jonahgraham:
> You also have to update the plugin.xml too, right? https://github.com/eclipse-cdt/cdt/blob/main/core/org.eclipse.cdt.core/plugin.xml#L75

Yes.

It was rather frustrating that this wasn't caught automatically by deleting the class, it will be with:
- #136

jonahgraham:
That turned out to be more to unravel than I expected. This change now needs #137 merging too.

jonahgraham:
> Test what happens if a project has the old parser referenced and they upgrade to it gone, if there is user visible effect, document in N&N

Nothing bad happens now that #137 is merged. However if a project has unknown binary parsers in it, opening the Binary Parsers tab and saving the results will remove the unknown parsers from the .cproject file.

mbooth101:
It looks like some removed IDs are still used in this plugin.xml: https://github.com/eclipse-cdt/cdt/blob/b37e9c38120283ec1ec4f478e1516805f409f71b/build/org.eclipse.cdt.autotools.core/plugin.xml#L396

Should **org.eclipse.cdt.core.PE;org.eclipse.cdt.core.Cygwin_PE** be changed here to **org.eclipse.cdt.core.PE64;org.eclipse.cdt.core.Cygwin_PE64** ?

ruspl-afed:
> Nothing bad happens now that https://github.com/eclipse-cdt/cdt/pull/137 is merged. However if a project has unknown binary parsers in it, opening the Binary Parsers tab and saving the results will remove the unknown parsers from the .cproject file.

Not sure what will happen with a lot of existing .cproject files if user _will not_ open "Binary Parsers" tab @jonahgraham

jonahgraham:
Argh. How long is this piece of string....

jonahgraham:
Thanks for the reviews - I will address all of it before merging. The fact that projects in CDT 10.7 were still being created with the long deprecated parser ids lead me to want to use a different solution in the IDs. New PS coming soon.

jonahgraham:
Turns out there is no schema for `BinaryParserPage` extension point (and a bunch of other ones too). This means that the error checking I added in #136 didn't have any effect on numerous parts of the code.

jonahgraham:
> Not sure what will happen with a lot of existing .cproject files if user will not open "Binary Parsers" tab @jonahgraham

The unknown binary parser IDs are ignored. If no suitable binary parser exists in the project then binaries won't be identified at all. 

>  different solution

I have now pushed my alternate solution - this was what I originally considered, but I liked the idea of just removing the old stuff and progressing forward. 

What I have done is left the old IDs in place, but pointing at the new classes. The old IDs are private, so they don't appear in the GUI as an option for users to select anymore.

WDYT?

jonahgraham:
> So, it will continue to understand existing (old) identifiers, but will use *64 code. And the existing projects will not be affected at all, correct?

Correct. We'll just live with the old identifiers forever. In hindsight this changing which class is used for the identifier could have been done back in CDT 6.9 instead of creating new identifiers at all.

ruspl-afed (approved):
I support the solution that does not impact existing user data

jonahgraham:
The one failed test is what is tracked/fixed by #128

mbooth101 (approved):
LGTM

PR #137, opened 2022-11-04
Fix unhandled event loop exception when binary parser is not in plugin.xml

jonahgraham:
If a .cproject references a binary parser ID that is not in the plug-in XML, or in the XML, but marked as private, the UI cannot display the binary parsers and was raising an ArrayIndexOutOfBoundsException as below.

This fix rewrites the array handling using collections.

```java
!ENTRY org.eclipse.ui 4 0 2022-11-04 09:44:27.409
!MESSAGE Unhandled event loop exception
!STACK 0
java.lang.ArrayIndexOutOfBoundsException: Index 7 out of bounds for length 7
	at org.eclipse.cdt.ui.newui.BinaryParsTab.updateData(BinaryParsTab.java:253)
	at org.eclipse.cdt.ui.newui.AbstractCPropertyTab.setVisible(AbstractCPropertyTab.java:253)
	at org.eclipse.cdt.ui.newui.BinaryParsTab.setVisible(BinaryParsTab.java:221)
	at org.eclipse.cdt.ui.newui.AbstractCPropertyTab.handleTabEvent(AbstractCPropertyTab.java:630)
	at org.eclipse.cdt.ui.newui.AbstractPage.updateSelectedTab(AbstractPage.java:412)
	at org.eclipse.cdt.ui.newui.AbstractPage$4.widgetSelected(AbstractPage.java:382)
```

This is needed more-so now so that the changes proposed in #135 doesn't cause the above exception if a project references the old ID. 

Steps to reproduce the bug:

1. Create a project without #135 merged
2. In Project Build Settings, check some of the deprecated binary parsers:

image

3. Open the above project with #135 merged and open the binary parsers page
4. Observe the above mentioned exception in the log and the binary parser list is empty:

image

The same could be observed by opening a project created with some third-party plug-in that defines such a binary parser and then opening that project without the third-party plug-in installed.

### DIFF
--- a/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/BinaryParsTab.java
+++ b/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/BinaryParsTab.java
@@ -13,9 +13,16 @@
  *******************************************************************************/
 package org.eclipse.cdt.ui.newui;
 
+import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.Iterator;
+import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
+import java.util.Set;
 
 import org.eclipse.cdt.core.CCorePlugin;
 import org.eclipse.cdt.core.model.CoreModelUtil;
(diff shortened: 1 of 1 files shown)


core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/BinaryParsTab.java
```diff
@@ -224,42 +231,38 @@ public void setVisible(boolean _visible) {
...
+					}
+				}
 			}
-			if (ids == null)
-				ids = new String[0]; // no selection
 		} else { // project
 			ICConfigurationDescription[] cfgs = page.getCfgsEditable();
-			ids = CoreModelUtil.getBinaryParserIds(cfgs);
-		}
-		Object[] data = new Object[configMap.size()];
```
jonahgraham, line 240: The key error is that size of data was wrong, it assumed that it needed to be the size of all plugin.xml defined binary parsers. It didn't account for ids having other ids, so the i variable below would be over incremented.

All this is solved by removing hard coded array sizes.

ruspl-afed (approved):
LGTM

PR #120, opened 2022-10-26
More test stability fixes

jonahgraham:
More work on #117

PR #138, opened 2022-11-06
Enhance memory data initialization checks

jld01:


dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/service/GDBMemory.java
```diff
@@ -314,7 +316,13 @@ public void eventDispatched(IExitedDMEvent event) {
 	@Override
 	public int getAddressSize(IMemoryDMContext context) {
 		Integer addressSize = fAddressSizes.get(context);
-		return (addressSize != null) ? addressSize.intValue() : 8;
+		assert addressSize != null;
```
ruspl-afed, line 319: I'm not sure regarding `assert` here, since further code handles the case under assertion as expected situation

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/service/GDBMemory.java
```diff
@@ -323,7 +331,13 @@ public int getAddressSize(IMemoryDMContext context) {
 	@Override
 	public int getAddressableSize(IMemoryDMContext context) {
 		Integer addressableSize = fAddressableSizes.get(context);
-		return (addressableSize != null) ? addressableSize.intValue() : 1;
+		assert addressableSize != null;
```
ruspl-afed, line 334: I'm not sure regarding `assert` here, since further code handles the case under assertion as expected situation

jld01: I'm mirroring `isBigEndian()` in the same class here. All launch sequences _should_ initialize memory data but there may be some old 3rd party launch sequences that still do not call the required method. We continue to permit this in released code (assertions disabled) but the `assert` makes it clear to extenders that they must make changes.

ruspl-afed (approved):
LGTM

jonahgraham (approved):
LGTM - BTW the failed test (testClassInUnnamedNamespace) is one of those covered in #117 and will hopefully never be seen again once I merge #120 :-)

jld01:
Thank you for the reviews @ruspl-afed and @jonahgraham.

PR #140, opened 2022-11-06
Remove unmaintained LSP and DAP support from CDT

### LINKED ISSUE #139: Future of org.eclipse.cdt.debug.dap and org.eclipse.cdt.lsp
With CDT 11 approaching I am trying to thin out some of the parts of CDT that are not being maintained so that they aren't ongoing maintenance and API burdens (e.g recent removal of [Qt plug-ins](https://github.com/eclipse-cdt/cdt/pull/124) and soon to be merged removal of [old binary parsers](https://github.com/eclipse-cdt/cdt/pull/135)).

We have a lot of experimental code sitting around CDT, started projects or Proof of Concepts.

One of these is the CDT provided Language Sever + Debug Adapter support. These are critical technologies for the future of Language support in Eclipse, but what we have in CDT is not maintained. It sometimes interferes with normal CDT operations and I am not convinced it works or is well enough documented.

The last non-refactoring / non-releng changes on LSP part has been:

- [4a3e046dd](https://github.com/eclipse-cdt/cdt/commit/4a3e046ddaf9f819d423b7cfc1dfebb1dbc5c15f) which reduced the interference of CDT LSP in normal C/C++ Editor in Aug 2020
- [8c78a241d](https://github.com/eclipse-cdt/cdt/commit/8c78a241d51511d5f844dd4efa30f57171c580b0) which was to improve cquery support in 2018, this came in just after Manish's work during GSoC in summer 2018.

As for the DAP implementation, that was a PoC that no one other than me looked at and it is not maintained.

Therefore, for CDT 11 I propose removing this code from CDT. If anyone is looking at picking up this work in the future I would support this code being included again in the future. 

**To be clear - the future of C/C++ support in Eclipse requires a (probably) clangd backed language server. But this requires some investment that hasn't been made by anyone in CDT as of now.**

@ruspl-afed I have explicitly cc'ed you as you are the only active committer to have touched this code since 2018.

jonahgraham:
Please see the full discussion in #139.

I will wait a few days before merging this, but I expect it to be merged in time for 2022-12 M3 on 14 Nov.

Fixes #139


ruspl-afed:
@jonahgraham perhaps we don't need both LSP4J and LSP4E in cdt.target anymore

jonahgraham:
I had the same hope, but it's needed to resolve Linux tools dependency.

ruspl-afed:
> I had the same hope, but it's needed to resolve Linux tools dependency.

OMG... I definitely need to learn more about Linux tools functionality

akurtakov:
Dockerfile editor is implemented using lsp4e and https://github.com/rcjsuen/dockerfile-language-server-nodejs so if you have this editor in your target platform you need lsp4e.

mbooth101:
@jonahgraham It should be possible to include only the **org.eclipse.linuxtools.docker.core** and **org.eclipse.linuxtools.docker.ui** bundles in CDT target platform instead of the whole Linuxtools Docker feature. The Dockerfile editor (the part that requires LSP) is (should be) a separate bundle

ruspl-afed:
You are right @mbooth101 it is a separate [feature](https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/blob/master/containers/org.eclipse.linuxtools.docker.editor.ls-feature/feature.xml). 
My comment was not to eliminate all LSP4* components in CDT, but to remove them **only if** they are not needed anymore. From my PoV Docker editor is useful addition to C/C++ functionality and if this editor needs LSP4* - let it be so.

ruspl-afed (approved):
+1

jonahgraham:
> @jonahgraham It should be possible to include only the **org.eclipse.linuxtools.docker.core** and **org.eclipse.linuxtools.docker.ui** bundles in CDT target platform instead of the whole Linuxtools Docker feature. The Dockerfile editor (the part that requires LSP) is (should be) a separate bundle

Thanks - the change could be made, but would also require changing the feature definition in CDT to not reference it anymore:

https://github.com/eclipse-cdt/cdt/blob/b10979677e9d8c3b330abe904cef56ec4223b7ac/launch/org.eclipse.cdt.docker.launcher-feature/feature.xml#L32-L34

is the same feature as the target platform pulls in:

https://github.com/eclipse-cdt/cdt/blob/b10979677e9d8c3b330abe904cef56ec4223b7ac/releng/org.eclipse.cdt.target/cdt.target#L21-L24

For now I would prefer to keep our features referencing features from linux tools.

mbooth101:
Fair enough, just wanted to make it clear CDT does not strictly depend on the Dockerfile editor. For users of this part of CDT it probably makes sense to continue pulling in the editor as they may find it useful :-)

jonahgraham:
> [jonahgraham](https://github.com/jonahgraham) [force-pushed](https://github.com/eclipse-cdt/cdt/compare/4f2d536db5949db5534d035d3cc8298364d8622d..818b2fb6edb5dc3ffa129cd42865dd6641cfce44) the remove_lsp_dap branch from [4f2d536 ](https://github.com/eclipse-cdt/cdt/commit/4f2d536db5949db5534d035d3cc8298364d8622d)to [818b2fb ](https://github.com/eclipse-cdt/cdt/commit/818b2fb6edb5dc3ffa129cd42865dd6641cfce44)[3 minutes ago](https://github.com/eclipse-cdt/cdt/pull/140#event-7775414100)

Just rebased the change - for reasons I don't understand the GitHub actions checks disappeared.

PR #141, opened 2022-11-06
Fix for "Need to hold a write lock to clear result caches assertion error"

jonahgraham:
This is a PR to work on issue #128. Until #120 is merged it also includes all those changes.


jonahgraham:
Great :-( - with the extra logging the tests don't seem to fail.

jonahgraham:
The fix is in #146

PR #146, opened 2022-11-07
Stop trying to clear result caches when lock isn't held

### LINKED ISSUE #128: Need to hold a write lock to clear result caches assertion error

jonahgraham:
The indexer has a feature that allows readers of the index to read the index in the middle of write operations. This is done by using a YeildableIndexLock.

The YeildableIndexLock's yield method can be called to temporarily give up the write lock. However the assumption in the code was that it would always successfully
reaquire the lock after that.

However, if the indexing was cancelled the lock would fail to be reaquired. Therefore the code that thinks it owns the lock no longer owns it. In this case the code in PDOMWriter.storeSymbolsInIndex's finally block.

Therefore I have added an new exception type to explicitly identify this use case so the original code can differentiate between cases where an exception was thrown where the lock is still held, and cases where the lock is no longer held.

Note that instead of a new exception caught like this:

```java
} catch (FailedToReAcquireLockException e) {
    hasLock = false;
    e.reThrow();
```

I could have done this:

```java
} catch (InterruptedException | OperationCanceledException e) {
    hasLock = false;
    throw e;
```

But it is not obvious that nothing else other than the acquire can raise an OperationCanceledException because it is a RuntimeException. By having a new checked exception we can know for sure that in the finally block we have lost our lock.

There are no API implications of this change as all the classes and interfaces are internal to CDT.

Fixes #128

### DIFF
--- a/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/pdom/FailedToReAcquireLockException.java
+++ b/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/pdom/FailedToReAcquireLockException.java
@@ -1,0 +1,42 @@
+/*******************************************************************************
+ * Copyright (c) 2022 Kichwa Coders Canada, Inc. and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
+package org.eclipse.cdt.internal.core.pdom;
+
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.OperationCanceledException;
+
+/**
+ * This exception is raised when {@link YieldableIndexLock#yield()} fails to
+ * reacquire the lock after yielding, this may be due to an {@link InterruptedException}
+ * or an {@link OperationCanceledException} which will be the nested exception.
+ */
+public class FailedToReAcquireLockException extends Exception {
+
+	public FailedToReAcquireLockException(InterruptedException e) {
+		super(e);
+		Assert.isNotNull(e);
+	}
+
+	public FailedToReAcquireLockException(OperationCanceledException e) {
+		super(e);
+		Assert.isNotNull(e);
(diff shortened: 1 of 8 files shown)


core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/pdom/FailedToReAcquireLockException.java
```diff
@@ -0,0 +1,38 @@
...
+		super(e);
+	}
+
+	public void reThrow() throws InterruptedException {
+		if (getCause() instanceof InterruptedException ie) {
+			throw ie;
+		}
+		if (getCause() instanceof OperationCanceledException oce) {
+			throw oce;
+		}
```
jonahgraham, line 39: Handle case of neither one.

core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/pdom/YieldableIndexLock.java
```diff
@@ -58,14 +59,20 @@ public void release() {
 	/**
 	 * Yields the lock temporarily if it was held for YIELD_INTERVAL or more, and somebody is waiting
 	 * for a read lock.
-	 * @throws InterruptedException
+	 * @throws FailedToReAcquireLockException when lock is not reacquired.
```
jonahgraham, line 61: The rest of this javadoc is out of date

PR #150, opened 2022-11-08
Handle all errors and runtime exceptions that may be thrown

### LINKED ISSUE #128: Need to hold a write lock to clear result caches assertion error

jonahgraham:
We can fail to regain our lock in other cases than just operation cancelled exception, so capture all of those cases to throw an FailedToReAcquireLockException.

Also, fix another finally block that assumes it has the lock.

Fixes #128

---

I am going to run this change a bunch of times on GHA to see what happens. I may need to deal with the new *Exception hit in finally block* log message as the tests fail on unexpected log messages.

### DIFF
--- a/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/pdom/FailedToReAcquireLockException.java
+++ b/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/pdom/FailedToReAcquireLockException.java
@@ -16,27 +16,26 @@ import org.eclipse.core.runtime.OperationCanceledException;
 /**
  * This exception is raised when {@link YieldableIndexLock#yield()} fails to
  * reacquire the lock after yielding, this may be due to an {@link InterruptedException}
- * or an {@link OperationCanceledException} which will be the nested exception.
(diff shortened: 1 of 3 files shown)


jonahgraham:
Three successful runs - calling this done and I will monitor builds and PRs.

PR #156, opened 2022-11-10
Workaround EASE 0.9.0 headless startup issue

jld01:
We register org.eclipse.cdt.codan.internal.ui.cxx.Startup with EASE to avoid triggering bug 580960.


codan/org.eclipse.cdt.codan.ui.cxx/plugin.xml
```diff
@@ -8,6 +8,14 @@
             class="org.eclipse.cdt.codan.internal.ui.cxx.Startup"></startup>
    </extension>
 
+   <!-- Suppress startup in org.eclipse.ease.runScript application (bug 580960) -->
+   <extension
```
ruspl-afed, line 12: [Sorry, I missed the direction on this discussion during the last CDT call]

I still think EASE need to stop invoking IStartup code. We should not encourage misuse of the Platform.

codan/org.eclipse.cdt.codan.ui.cxx/.settings/org.eclipse.pde.prefs
```diff
@@ -27,7 +27,7 @@ compilers.p.unknown-class=0
 compilers.p.unknown-element=1
 compilers.p.unknown-identifier=0
 compilers.p.unknown-resource=0
-compilers.p.unresolved-ex-points=0
+compilers.p.unresolved-ex-points=1
```
jonahgraham, line 30: This will need something like:

```sh
        if echo $i | grep 'org.eclipse.cdt.codan.ui.cxx' > /dev/null; then
            # Special case, see comment in org.eclipse.cdt.codan.ui.cxx/plugin.xml
            sed -i \
                '-es@compilers.p.unresolved-ex-points=0@compilers.p.unresolved-ex-points=1@' \
                $d/.settings/org.eclipse.pde.prefs
        fi
```

added to `releng/scripts/do_project_settings.sh` in the PDE section

codan/org.eclipse.cdt.codan.ui.cxx/plugin.xml
jonahgraham: 💯 what EASE is doing is bad idea. They should ditch the whole IStartup and, in order of my preference do one of these:

- Take the IStartup they want access to and use a different non-UI startup mechanism, this probably requires contributions to whatever upstream project they have dependency on.
- Create a new EASE Startup extension point that will be run that plug-ins can opt-in to
- Provide a way in the script to explicitly run the IStartup of a bundle they care about, e.g in the script.

```python
runEclipseUIStart("name.of.bundle")
```
- Make it a "whitelist" so that plug-ins opt-in to the behaviour instead of opting out.

But, until this is resolved we need to have a non-broken CDT + EASE integration.

jonahgraham (changes requested):
My approval is only pending on code cleanliness. @jld01 please let me know if you want me to just apply my suggested change to do_project_settings.

codan/org.eclipse.cdt.codan.ui.cxx/.settings/org.eclipse.pde.prefs
jld01: Thanks for the code snippet, @jonahgraham. I will integrate and update the PR.

jonahgraham (approved):
LGTM (assuming builds all pass)

PR #157, opened 2022-11-10
Indexing source files when multiple toolchains are used.

ewaterlander:
This change solves the indexing of C/C++ files when multiple toolchains are used in a single Makefile. This is for the use case in which one (Linux) gcc compiler plus one or more custom embedded C compilers (all producing ELF format binaries) are used.

To get proper indexing we need to know for each resource which toolchain was used. The sub build configuration (via extension point org.eclipse.cdt.core.buildConfigProvider) extends
StandardBuildConfiguration.java, and overrides method IToolChain (List<String> commandgetToolChain). tcMap is filled with a map of toolchains per resource. The primary toolchain keeps pointing to the gcc toolchain.

### DIFF
--- a/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
+++ b/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
@@ -106,6 +106,8 @@ public abstract class CBuildConfiguration extends PlatformObject implements ICBu
 
 	private static final List<String> DEFAULT_COMMAND = new ArrayList<>(0);
 
+	private final Map<IResource, IToolChain> tcMap = new HashMap<IResource, IToolChain>();
+
 	private final String name;
 	private final IBuildConfiguration config;
 	private final IToolChain toolChain;
@@ -308,6 +310,31 @@ public abstract class CBuildConfiguration extends PlatformObject implements ICBu
 	@Override
 	public IToolChain getToolChain() throws CoreException {
 		return toolChain;
+	}
+
+	/**
+	 * Get the toolchain based on the compiler found in the command line.
+	 * Sub-classes can override this method to set their own method for
+	 * getting the toolchain.
+	 *
+	 */
+	protected IToolChain getToolChain(List<String> command) throws CoreException {
+		return getToolChain();
+	}
+
+	/**
+	 * Get the toolchain that was used to compile this resource.
+	 *
+	 * @param resource
+	 * @return
+	 * @throws CoreException
+	 */
+	private IToolChain getToolChain(IResource resource) throws CoreException {
+		IToolChain tc = tcMap.get(resource);
+		if (tc == null)
+			return getToolChain();
+		else
+			return tc;
 	}
 
 	@Override
@@ -665,7 +692,8 @@ public abstract class CBuildConfiguration extends PlatformObject implements ICBu
 			if (celement instanceof ITranslationUnit) {
 				try {
 					ITranslationUnit tu = (ITranslationUnit) celement;
-					info = getToolChain().getDefaultScannerInfo(getBuildConfiguration(), getBaseScannerInfo(resource),
+					IToolChain tc = getToolChain(resource);
+					info = tc.getDefaultScannerInfo(getBuildConfiguration(), getBaseScannerInfo(resource),
 							tu.getLanguage(), getBuildDirectoryURI());
 					synchronized (scannerInfoLock) {
 						scannerInfoCache.addScannerInfo(DEFAULT_COMMAND, info, resource);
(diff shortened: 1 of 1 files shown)


ewaterlander:
About the code cleanliness check: I ran the standard Source > Format menu on the code, using the Eclipse built-in formatter profile.

cmorty:
@ewaterlander : You need to run `./releng/scripts/check_code_cleanliness.sh` locally. Otherwise fixing them is a PITA!

jonahgraham:
The JDT's Save Actions should have generated the correct format for the code. However over time as JDT bugs are fixed we need to update the JDT on the build machine to make sure they match as there are slight formatting changes. 

Also, some refactorings skip the save actions, if that refactoring changes the length of an identifier it leaves badly formatted code behind.

I am more than happy to accept improvements in this area. One idea that I had was to simply commit the change back to the PR rather than fail the code cleanliness check. This is easier now that we are on GitHub.

jonahgraham:
> About the code cleanliness check: I ran the standard Source > Format menu on the code, using the Eclipse built-in formatter profile.

PS - the CDT project does not use the built-in formatter profile, there is a CDT profile. But on closer inspection I see that there is also trailing whitespace on some lines. The save actions are supposed to clean that up, but sometimes they don't as there seems to be a conflict between JDT removing trailing spaces and the formatter adding trailing spaces in javadoc comments.

jonahgraham:
Anyway, don't worry about the formatting, I can fix that up before merge this time once I complete the rest of the review.

jonahgraham (commented):
I need more time to mull this over. It is much better than the #130 as you have solved the problem of leaving things in a bad state, so I am inclined to merge it. Leave it with me for a day or two.

jonahgraham:
I cannot push my formatting updated because (I think) you have disabled ["Allow edits and access to secrets by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) - which is fine as there are security considerations around it - but it means you'll have to do the formatting fix before I can merge.

The checkbox is on the right side column:
image

ewaterlander:
> @ewaterlander : You need to run `./releng/scripts/check_code_cleanliness.sh` locally. Otherwise fixing them is a PITA!
Thanks!

ewaterlander:
> I cannot push my formatting updated because (I think) you have disabled ["Allow edits and access to secrets by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) - which is fine as there are security considerations around it - but it means you'll have to do the formatting fix before I can merge.
> 
> The checkbox is on the right side column: image

Indeed, I did not check that box. I have fixed the formatting now.

cmorty:
OT:

> One idea that I had was to simply commit the change back to the PR rather than fail the code cleanliness check. This is easier now that we are on GitHub.

This is ugly stuff. Especially as you probably want to run it using `git filter-branch ` to actually fix the right patch.

ewaterlander:
> > I cannot push my formatting updated because (I think) you have disabled ["Allow edits and access to secrets by maintainers"](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork) - which is fine as there are security considerations around it - but it means you'll have to do the formatting fix before I can merge.
> > The checkbox is on the right side column: image
> 
> Indeed, I did not check that box. I have fixed the formatting now.

I also allow maintainers edits now.

jonahgraham (dismissed):
> I need more time to mull this over. It is much better than the #130 as you have solved the problem of leaving things in a bad state, so I am inclined to merge it. Leave it with me for a day or two.

Approved. I am still a little unsure about this, but it does address your use case and it doesn't have the problems I was concerned about in #130. So once build gives it +1 I can merge this.

ewaterlander:
Do not merge it yet.
I see a hang if the example code generation is on when the project is created.

ewaterlander:
> Do not merge it yet. I see a hang if the example code generation is on when the project is created.

The hang was not caused by my patch. I had it too in the unmodified version. After rebuilding my stuff I did not see it anymore.

ewaterlander:
> > Do not merge it yet. I see a hang if the example code generation is on when the project is created.
> 
> The hang was not caused by my patch. I had it too in the unmodified version. After rebuilding my stuff I did not see it anymore.

This change can be merged.

jonahgraham:
>  I had it too in the unmodified version

If you are able to, please provide a bug report on this. Ideally including the output of jstack which will hopefully show any deadlocks

ewaterlander:
> > I had it too in the unmodified version
> 
> If you are able to, please provide a bug report on this. Ideally including the output of jstack which will hopefully show any deadlocks

OK, I will. I do not have experience with jstack. Do you have some hints?

jonahgraham:
jstack should be included with your JDK, it is a command line program that you pass the PID to to get the stack traces of all running threads, and it is able to detect some kinds of deadlocks.

PR #158, opened 2022-11-11
Add all flags to scanner discovery

jonahgraham:
Fixes Bug [571722](https://bugs.eclipse.org/bugs/show_bug.cgi?id=571722). 
    
This change adds the ALL_FLAGS that does not limit tool options to
those declared as IOption::isForScannerDiscovery when launching the
compiler to discover compiler built-ins.

This is needed as many other flags, either entered manually in "Other
flags" or some of the existing flags with checkboxes such as "-ansi",
"-fPIC", and "-fstack-protector-all" which all affect scanner discovery
as they can all change what macros are built-in to the compiler.

The current solution has as a drawback that some settings, like -I and -D
then appear twice. For example in the "Includes" node in the "Project
Explorer"

My only reservation about this change is if there is an option
that can be specified successfully at build time, but when used
at scanner discovery time causes the compiler to fail, or return
incorrect results. Therefore I have added a new field,
excludeFromScannerDiscovery to tool options (buildDefinitions
extension point) that allows tool integrators to always exclude
a command line option from ALL_FLAGS. I have also added
a new "Other flags (excluded from discovery)" to the
"Miscellaneous" tab to allow compiler options to be entered
by the user.

TODO:

- [x] N&N entry
- [x] Provide PR to embed-cdt: https://github.com/eclipse-embed-cdt/eclipse-plugins/pull/547
- [x] This PR includes some code cleanups in the parent commits, submit them separately  https://github.com/eclipse-cdt/cdt/pull/160


jonahgraham:
@jld01 @T-Svensson I would appreciate your thoughts on this while I work to make sure it doesn't regress anything - I wrote the N&N entries too:

- https://github.com/jonahgraham/cdt/blob/add_ALL_FLAGS_to_scanner_discovery/NewAndNoteworthy/CDT-11.0.md#scanner-discovery-considering-all-flags
- https://github.com/jonahgraham/cdt/blob/add_ALL_FLAGS_to_scanner_discovery/NewAndNoteworthy/CDT-11.0.md#exclude-from-scanner-discovery-flag-for-options

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/internal/ui/newui/Messages.properties
```diff
@@ -101,7 +101,8 @@ AbstractPage_6=Configuration:\u0020
 AbstractPage_7=Exclude resource from build
 AbstractPage_8=Cannot apply
 AbstractPage_9=Internal error
-AbstractPage_rebuildIndex_question=Changes made will not be reflected in the index until it is rebuilt. Do you wish to rebuild it now?
+AbstractPage_Rebuild_Index=Rebuild Index
+AbstractPage_rebuildIndex_question=Some build settings changes may affect the index. These changes won't take effect until the index until it is rebuilt. Do you wish to rebuild it now?
```
Torbjorn-Svensson, line 105: Some changes to build settings may affect the index. These changes won't take effect until the index is rebuilt. Do you wish to rebuild it now?

jonahgraham:
Thanks @Torbjorn-Svensson for the review. I agree with all your changes.

At the high level, is this change a good idea, or cause you any concern?

I am pretty confident about most of it, the only thing that concerns me is if I should change the default for new projects to include `${ALL_FLAGS}` or leave it for affected users to change it.

Torbjorn-Svensson:
> At the high level, is this change a good idea, or cause you any concern?
>
> I am pretty confident about most of it, the only thing that concerns me is if I should change the default for new projects to include `${ALL_FLAGS}` or leave it for affected users to change it.

My first reaction was "why is this needed...", but after reading the entire PR and the changes in it, I suppose it makes senses. 
If we are going to do the change, it should be done now as CDT11 contains breaking changes in other areas.
Maybe we should leave the default pattern to be the old one, but then I suppose we will never be able to offer the new pattern as the default value and we would sort of stay with the current limitations (the ones that this PR is suppose to resolve).
Do you think there is a good way to select the pattern on a high level without having to write it manually in the plugin.xml? Some property that will dynamically select the the new pattern if the toolchain description states that it's compatible but without knowing the actual pattern, or something like that? Or we just let our users do this one time switch when they upgrade to CDT11? I don't see this as a major issue anyway.

jonahgraham:
I don't have confidence to change the default, so this latest version adds the new `${ALL_FLAGS}` but make it not the default.

> Do you think there is a good way to select the pattern on a high level without having to write it manually in the plugin.xml?

Short answer is that I didn't come up with a good way. So this feature is there for those that know about it (who read the N&N/docs).

Separately, it may be worth going to set `useByScannerDiscovery` to true for many more options so that people are less likely to fall for the problem in the first place. I am going to do that for embed-cdt instead of the previous PR (https://github.com/eclipse-embed-cdt/eclipse-plugins/pull/547) which also has the advantage of working with older CDTs as embed-cdt has CDT 10 as minimum version.

Torbjorn-Svensson:
I just had another thought about this.
Can't we simply flip the default value for useByScannerDiscovery from false to true.
Sure, there are plugins out there that want to be compatible with both CDT10 and CDT11 and this way, they just have to declare what settings that should *not* be part of the scanner logic. I mean, is there really a need for a new pattern?

jonahgraham:
I think that has the same risks of breaking existing projects as making ALL_FLAGS the default. In practice it will mean that all existing options will be effectively useByScannerDiscovery == true. But unlike the ALL_FLAGS being default, there will be no way for users to get it to work if there is a conflict between the option and scanner discovery (except to move those options into the new option field I added, which isn't even possible for some enum type options)

Your proposal *should* have been the original design!

Torbjorn-Svensson:
I guess you are right, but the ALL_FLAGS workaround means that after you do that, it's strictly CDT11 as the symbol is unavailable in CDT10. Isn't that worse of a break?

I'm not saying that we should do A or B, I'm just venting our options.

jonahgraham:
> Isn't that worse of a break?

I don't think that is a break at all. The change is backwards compatible, but not forwards compatible.

Torbjorn-Svensson:
> > Isn't that worse of a break?
> 
> I don't think that is a break at all. The change is backwards compatible, but not forwards compatible.

I was thinking more about if someone adopts the ALL_FLAGS instead of FLAGS, then that project will not work with CDT10, while the solution to invert the default value for useByScannerDiscovery would work for both, but "forces" the toolchain integration provider to have the correct value on the options.

jonahgraham:
I fear you may be right, this change is backwards compatible, but not forwards compatible. Basically what level of support should we provide users who open a project created with CDT 11 in CDT < 11?

jonahgraham:
BTW there is too separate parts of the conversation, changing useByScannerDiscovery default to true is not backwards compatible, so I don't think that can be changed at all.

So now my concern is by introducing ALL_FLAGS we have something that is not forwards compatible from pre-CDT 11 to CDT 11.

I'm going to look at how we handled that kind of not forward compatible change in the past and we'll see what we do from there.

jonahgraham:
On review I don't think we have handled these types of forward compatibility problems properly in the past, so I don't think we need to worry about it now.

For example if in version N a new compiler option is added to the UI, a user checks it, and then opens that project up in version N-1, the option will be silently ignored, which, depending on the option, will lead to compiler error, or differently compiled code.

It would probably be nice to have a warning for projects, similar to what happens when you try to open a workspace in an older version of eclipse, but I think that is outside the scope of this issue.

image

PR #159, opened 2022-11-11
Make running code cleanliness less painfull

cmorty:
* Only check for modifications in tracked files
* Tell the user to stash stuff, before running

### DIFF
--- a/releng/scripts/check_code_cleanliness.sh
+++ b/releng/scripts/check_code_cleanliness.sh
@@ -12,6 +12,10 @@
 
 set -e
 
+if test ! -z "$(git status -s -uno)"; then
+    echo "You have changes. Please stash them before continuing."
+fi
+
 ##
 # Check the features are all branded and all content has proper licenses
 ##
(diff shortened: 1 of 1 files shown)


jonahgraham:
I had no idea people ran the scripts before commit and push :-)

releng/scripts/check_code_cleanliness.sh
```diff
@@ -12,6 +12,10 @@
 
 set -e
 
+if test ! -z "$(git status -s -uno)"; then
+    echo "You have changes. Please stash them before continuing."
+fi
```
jonahgraham, line 17: Right after submitting I wondered if this should actually `exit 1`?

cmorty: OOps. That somehow got lost. Sorry :disappointed:

cmorty:
@jonahgraham Moving all the patches to a local system by hand is a PITA. It's easier to just run them locally.

releng/scripts/check_code_cleanliness.sh
cmorty: Actually not somehow. I removed it (and didn't comment it out :-( ) to be able to run the script on the other branch without committing it.

jonahgraham:
> @jonahgraham Moving all the patches to a local system by hand is a PITA. It's easier to just run them locally.

OK - I am more than happy to receive any script updates that make them more useful to you!

PR #162, opened 2022-11-11
Correct typographical error in remote console translation

mbooth101:
Openning -> Opening


jonahgraham (approved):
The key name has the same typo - but if it is non-trivial to fix that too it can stay.

mbooth101:
> The key name has the same typo - but if it is non-trivial to fix that too it can stay.

Let me check, I honestly just grep'd for the typo and edited it in vim.

PR #164, opened 2022-11-14
Update to latest Orbit which means new mockito version

jonahgraham:


jonahgraham:
@betamaxbandit the CDT Oomph install should work again, although you may want to wait a few minutes as I think there is sync runs every 15 minutes.

betamaxbandit:
> @betamaxbandit the CDT Oomph install should work again, although you may want to wait a few minutes as I think there is sync runs every 15 minutes.

Thanks, it's working now.

PR #168, opened 2022-11-15
Fix CHANGELOG-API anchor

jld01:


jonahgraham (approved):
Thanks for the fix!

jonahgraham:
My endless whack-a-mole has a new entry #169 - obviously the test failure has nothing to do with your change, but does prevent a proper green tick :-(

PR #173, opened 2022-11-21
Resolve deadlock in Jobmanager related code

### LINKED ISSUE #81: Deadlock in indexer

jonahgraham:
Because of [changes](https://www.eclipse.org/eclipse/news/4.26/platform_isv.php#JobManager_Implementation) in Eclipse Platform where the jobmanager's behaviour changes (within the API), the consumers of the jobmanager can deadlock due to incorrect assumptions.

In particular, where we call job.schedule(), the callback can happen in different threads to the IJobChangeListener's. As CDT was holding a lock while calling schedule that is also required in those listeners, we need to no longer lock when calling schedule.

As the code already dealt with the case of when there was a delay between the job.schedule() and where & when it was run, we can move the schedule call out of the synchronized block.

Fixes #81

### DIFF
--- a/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/pdom/PDOMManager.java
+++ b/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/pdom/PDOMManager.java
@@ -709,8 +709,8 @@ public class PDOMManager implements IWritableIndexManager, IListener {
 				}
 			}
 			fTaskQueue.addLast(subjob);
-			fIndexerJob.schedule();
 		}
+		fIndexerJob.schedule();
 	}
 
 	IPDOMIndexerTask getNextTask() {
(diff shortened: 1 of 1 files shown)


ruspl-afed:
This should eliminate the deadlock described, but for some reason it was synchronized earlier. Is this reason still valid?

jonahgraham:
> Is this reason still valid?

There is no available information about why it was in that synchronized block. That change was introduced in 2006 (74e45cb1a32cbb627fed829627700e2e916ef294 and 6fad7334f71228cd2f8033f0d7d951947c188858) when the whole PDOM was substantially under development.

My best guess is that it was included in the block because it didn't hurt to have it in. A schedule call is fundamentally an asynchronous call, so making it in the synchronized block was probably never important. 

FWIW the restructuring I did now looks simple-ish as the schedule call was the list call in the synchronized block. However over time that became the last item, in earlier versions of the code the schedule call was somewhat more entangled in the synchronized block's logic.

---

This is part of why I pushed back in https://github.com/eclipse-platform/eclipse.platform/issues/193#issuecomment-1268409175 and https://github.com/eclipse-platform/eclipse.platform/issues/193#issuecomment-1284711057 (last paragraph)

Fundamentally code that has (seemingly) worked for 10-20 years is going to be broken by the change in platform. It has been accepted that that change is valid, but I pity any adopters who have to identify and fix why there code is now deadlocking.

ruspl-afed:
>  It has been accepted that that change is valid, but I pity any adopters who have to identify and fix why there code is now deadlocking.

Yes, at this age (of Platform) it is better to avoid such a change. Even if the behavior is not completely reasonable it has already been workarounded and tested. In any case, the time to discuss this change has passed, and it has already been made.

PR #177, opened 2022-11-22
Use mi-async instead of target-async on GDB >= 7.12

### LINKED ISSUE #176: target-async is deprecated
When launching a debug session for a target, the following is printed in the GDB traces:

```
237,992 13-gdb-set target-async on
238,005 ~"Warning: 'set target-async', an alias for the command 'set mi-async', is deprecated.\n"
238,006 ~"Use 'set mi-async'.\n\n"
238,006 13^done
```

I suppose we should switch to the mi-async variant instead, but I don't see any easy way to accomplish that without code duplication and/or some kind of API break in the FinalLaunchSequence* classes.

Looking at the GDB sources, it appears it was introduced [329ea57934a9d4b250a0b417af1ec47bc2d0ceb6](https://sourceware.org/git/?p=binutils-gdb.git;a=commit;h=329ea57934a9d4b250a0b417af1ec47bc2d0ceb6) and was first included in the GDB 7.8 release.

jonahgraham:
target-async is the old name for mi-async. mi-async came into existence in GDB 7.8, but in later releases it becomes a warning to use the old name.

Because we already have the version infra for 7.12 we change to using mi-async from 7.12

A similar change was made for [cdt-gdb-adapter a while ago](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/main/src/GDBBackend.ts#L114-L116)

Fixes #176

### DIFF
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/mi/service/command/CommandFactory.java
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/mi/service/command/CommandFactory.java
@@ -135,6 +135,7 @@ import org.eclipse.cdt.dsf.mi.service.command.commands.MIGDBSetDisconnectedTraci
 import org.eclipse.cdt.dsf.mi.service.command.commands.MIGDBSetEnv;
 import org.eclipse.cdt.dsf.mi.service.command.commands.MIGDBSetHostCharset;
 import org.eclipse.cdt.dsf.mi.service.command.commands.MIGDBSetLanguage;
+import org.eclipse.cdt.dsf.mi.service.command.commands.MIGDBSetMIAsync;
 import org.eclipse.cdt.dsf.mi.service.command.commands.MIGDBSetNewConsole;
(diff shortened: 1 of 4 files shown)


dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/mi/service/command/CommandFactory.java
```diff
@@ -895,6 +896,11 @@ public ICommand<MIInfo> createMIGDBSetTargetAsync(ICommandControlDMContext ctx,
 		return new MIGDBSetTargetAsync(ctx, isSet);
```
Torbjorn-Svensson, line 896: Should we mark createMIGDBSetTargetAsync() as deprecated?

jonahgraham: No, only because it wouldn't match how we have dealt with all the support for older GDB. If we start down that road a huge amount of DSF-GDB would get marked as deprecated, and I don't think there is appetite for that.

PR #178, opened 2022-11-24
C and C++ should not be labelled as GNU dialects

TheShermanTanker:
C and C++ are currently labelled as "GNU C" and "GNU C++" in the explorer when choosing what options to set for any specified language (For example, in the Preprocessor Include Paths, Macros etc screen). Not only is this confusing since it implies that Eclipse by default uses the GNU dialects of the languages (which is not true), this is also flat out wrong in other cases, such as when Workspaces configured with the Microsoft Visual C++ compiler (one that couldn't possibly be more incompatible with the actual GNU dialect) are _also_ labelled as "GNU C". This was likely due to Eclipse only really supporting gcc when it started out, but this is no longer the case with the IDE becoming more and more flexible over the years, and it's probably time to ditch the GNU labels in the hardcoded language names.

### DIFF
--- a/NewAndNoteworthy/CDT-11.0.md
+++ b/NewAndNoteworthy/CDT-11.0.md
@@ -34,6 +34,10 @@ Please see the corresponding issue for more details.
   - `org.eclipse.cdt.lsp.cquery`
   - `org.eclipse.cdt.lsp.ui`
 
+## GNU removed from display name language for C/C++
+
+C and C++ have been renamed from "GNU C" and "GNU C++" to just "C" and "C++". If you have exported settings with CDT prior to CDT 10.5 they may not import successfully with CDT 11. See #178
+
 # Build
 
 ## Scanner Discovery able to consider all flags
--- a/NewAndNoteworthy/CDT-11.0.md
+++ b/NewAndNoteworthy/CDT-11.0.md
@@ -36,7 +36,8 @@ Please see the corresponding issue for more details.
 
 ## GNU removed from display name language for C/C++
 
-C and C++ have been renamed from "GNU C" and "GNU C++" to just "C" and "C++". If you have exported settings with CDT prior to CDT 10.5 they may not import successfully with CDT 11. See #178
+C and C++ have been renamed from "GNU C" and "GNU C++" to just "C" and "C++". If you have exported settings with CDT prior to CDT 10.5 they may not import
+successfully with CDT 11. See #178
 
 # Build
 
--- a/core/org.eclipse.cdt.core/plugin.properties
+++ b/core/org.eclipse.cdt.core/plugin.properties
@@ -110,8 +110,8 @@ PDOMProviderName=PDOM Provider
 fastIndexer.name=Fast Indexer
 
 # Built-in languages
-language.name.gcc= GNU C
-language.name.gpp= GNU C++
+language.name.gcc= C
+language.name.gpp= C++
 language.name.asm= Assembly
 
 CConfigurationDataProvider.name = CConfigurationData provider
--- a/core/org.eclipse.cdt.ui.tests/ui/org/eclipse/cdt/ui/tests/wizards/settingswizards/SettingsImportExportTest.java
+++ b/core/org.eclipse.cdt.ui.tests/ui/org/eclipse/cdt/ui/tests/wizards/settingswizards/SettingsImportExportTest.java
@@ -246,7 +246,7 @@ public class SettingsImportExportTest extends BaseUITestCase {
 	// {badXML6}
 	// <cdtprojectproperties>
 	// <section name="org.eclipse.cdt.internal.ui.wizards.settingswizards.Macros">
-	// <language name="GNU C++">
+	// <language name="C++">
 	// <macro>
 	// <name>aaaa</name>
 	// </macro>
@@ -257,7 +257,7 @@ public class SettingsImportExportTest extends BaseUITestCase {
 	// {badXML7}
 	// <cdtprojectproperties>
 	// <section name="org.eclipse.cdt.internal.ui.wizards.settingswizards.Macros">
-	// <language name="GNU C++">
+	// <language name="C++">
 	// <macro>
 	// <name>aaaa</name><value></value><value></value>
 	// </macro>
@@ -268,7 +268,7 @@ public class SettingsImportExportTest extends BaseUITestCase {
 	// {badXML8}
 	// <cdtprojectproperties>
 	// <section name="org.eclipse.cdt.internal.ui.wizards.settingswizards.IncludePaths">
-	// <language name="GNU C++">
+	// <language name="C++">
 	// <includepath>C:\WINDOWS</includepath><invalid></invalid>
 	// </language>
 	// </section>


jonahgraham:
Thanks @TheShermanTanker for the Pull Request. 

I am not sure of all places that is used, so I will need to review it a little first.

In the meantime, can you sign the ECA please: https://github.com/eclipse-cdt/cdt/blob/main/CONTRIBUTING.md#eclipse-contributor-agreement

TheShermanTanker:
> Thanks @TheShermanTanker for the Pull Request.
> 
> I am not sure of all places that is used, so I will need to review it a little first.
> 
> In the meantime, can you sign the ECA please: https://github.com/eclipse-cdt/cdt/blob/main/CONTRIBUTING.md#eclipse-contributor-agreement

Thanks, I did sign it earlier actually, but the bot doesn't seem to be able to resolve the ECA. I'm guessing there's a bit of a delay involved?

jonahgraham:
> ## Test Results
>      594 files     594 suites   15m 50s [stopwatch](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) 10 127 tests 8 588 [heavy_check_mark](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) 22 [zzz](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) 1 517 [x](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) 10 143 runs  8 599 [heavy_check_mark](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) 22 [zzz](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols) 1 522 [x](https://github.com/EnricoMi/publish-unit-test-result-action/blob/v1.20/README.md#the-symbols)
> 
> For more details on these failures, see [this check](https://github.com/eclipse-cdt/cdt/runs/9695663692).
> 
> Results for commit [1d8cdc4](https://github.com/eclipse-cdt/cdt/commit/1d8cdc4e256cd8341a236a7121d98b285f1cd089).

Hmm - is there really something that depends on that string in the code/tests? I have retriggered tests to make sure we aren't getting a false failure.

jonahgraham:
> Thanks, I did sign it earlier actually, but the bot doesn't seem to be able to resolve the ECA. I'm guessing there's a bit of a delay involved?

Seems that way - revalidating worked.

jonahgraham:
> This was likely due to Eclipse only really supporting gcc when it started out

FWIW non-GCC has always been an important part of the CDT ecosystem, from the very early days. Has no one ever noticed/commented or bothered to fix this before? Strange.

jonahgraham:
There is indeed tests that depend on the name as is, e.g.:

https://github.com/eclipse-cdt/cdt/blob/1d73f112dc309bea749519cf8854fbc32be2ebc8/core/org.eclipse.cdt.ui.tests/ui/org/eclipse/cdt/ui/tests/wizards/settingswizards/SettingsImportExportTest.java#L249

The test output is this:

```
 <cdtprojectproperties>
 <section name="org.eclipse.cdt.internal.ui.wizards.settingswizards.IncludePaths">
 <language name="GNU C++">
 <includepath>C:\WINDOWS</includepath><invalid></invalid>
 </language>
 </section>
 </cdtprojectproperties>
 <cdtprojectproperties>
 <section name="org.eclipse.cdt.internal.ui.wizards.settingswizards.IncludePaths">
 <language name="GNU C++">
 <includepath>C:\WINDOWS</includepath><invalid></invalid>
 </language>
 </section>
 </cdtprojectproperties>

	at org.eclipse.cdt.ui.tests.wizards.settingswizards.SettingsImportExportTest.vaidateCorrectErrorHandling(SettingsImportExportTest.java:191)
	at org.eclipse.cdt.ui.tests.wizards.settingswizards.SettingsImportExportTest.testNotValid(SettingsImportExportTest.java:284)
```

which doesn't really help much.

The test is really old, and I suspect not relevant as it refers to the old format for CDT projects.

The other tests all fail because the above test doesn't clean up after itself when it fails and leaves the workspace in a bad state.

TheShermanTanker:
The test might be old, but I don't feel confident enough to do anything with it. I've instead just refactored it to expect the new language name instead to minimize the amount of changes I'm making

jonahgraham:
As it turns out CDT 11 is a good time to make this change because in [Bug 577074](https://bugs.eclipse.org/577074) commit https://github.com/eclipse-cdt/cdt/commit/fa8a6279d26c357a502f067b94f57a3614954a34 I changed the import and export wizard to use the language ID instead of the name of the language. This change will mean people who have exported with CDT < 10.5 won't be able to import anymore. Therefore an entry in the N&N is needed to guide them.

It would be great if you updated the [New and Noteworthy](https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CDT-11.0.md), but if you aren't able to by Monday I will do it before I submit the final build for CDT 11.

That said, I haven't actually completed my review to make sure there are no other side-effects of this name change. (The joys of working with old code bases, even simple thing like changing a display string is a non-trivial effort.)

TheShermanTanker:
> As it turns out CDT 11 is a good time to make this change because in [Bug 577074](https://bugs.eclipse.org/577074) commit [fa8a627](https://github.com/eclipse-cdt/cdt/commit/fa8a6279d26c357a502f067b94f57a3614954a34) I changed the import and export wizard to use the language ID instead of the name of the language. This change will mean people who have exported with CDT < 10.5 won't be able to import anymore. Therefore an entry in the N&N is needed to guide them.
> 
> It would be great if you updated the [New and Noteworthy](https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CDT-11.0.md), but if you aren't able to by Monday I will do it before I submit the final build for CDT 11.
> 
> That said, I haven't actually completed my review to make sure there are no other side-effects of this name change. (The joys of working with old code bases, even simple thing like changing a display string is a non-trivial effort.)

Something like this?
`C and C++ have been internally renamed from "GNU C" and "GNU C++" to just "C" and "C++". You may need to take note of this change when using the import/export wizard`
Which section of "Noteworthy" would you like this to go under?

jonahgraham:
> Something like this?
> C and C++ have been internally renamed from "GNU C" and "GNU C++" to just "C" and "C++". You may need to take note of this change when using the import/export wizard

That looks good, some small mods, plus lets be explicit about versions.

C and C++ have been renamed from "GNU C" and "GNU C++" to just "C" and "C++". If you have exported settings with CDT prior to CDT 10.5 they may not import successfully with CDT 11. See https://github.com/eclipse-cdt/cdt/pull/178

> Which section of "Noteworthy" would you like this to go under?

It can go in the "release notes" section with heading "GNU removed from display name language for C/C++"

jonahgraham:
I have taken some time to look at this in the code. It is not so straightforward to rename this. This is because the name *really* is the name for `GNU C` and `GNU C++`. It is just typical that only `GNU C` and `GNU C++` is installed, especially if you get plain CDT. Those names assign languages to the specific language implementations.

> this is also flat out wrong in other cases, such as when Workspaces configured with the Microsoft Visual C++ compiler (one that couldn't possibly be more incompatible with the actual GNU dialect) are also labelled as "GNU C". 

However CDT *is* treating those C and C++ files as GNU C and GNU C++ files. i.e. standard CDT won't handle MSVC extensions (generally) because no one has contributed a subclass of AbstractCLikeLanguage for MSVC. Therefore CDT will use the GNU version.

Vendors (like ARM, Renesas, probably IAR too) do contribute their own language definitions with their own names. When you create one of those vendor C/C++ projects you get a project where the language mappings are from org.eclipse.cdt.core.cSource -> vendor.c.language id/name.

So, I don't think this change can be accepted as is.

Which toolset are you using? If you are using the MSVC integration that comes with CDT, then it needs that additional work to mark the language mappings as MSVC C and C++.

Am I understanding the problem correctly? Please feel free to share your thoughts on this.

BTW are you an end user or a vendor integrating CDT?

TheShermanTanker:
> I have taken some time to look at this in the code. It is not so straightforward to rename this. This is because the name _really_ is the name for `GNU C` and `GNU C++`. It is just typical that only `GNU C` and `GNU C++` is installed, especially if you get plain CDT. Those names assign languages to the specific language implementations.
> 
> Vendors (like ARM, Renesas, probably IAR too) do contribute their own language definitions with their own names. When you create one of those vendor C/C++ projects you get a project where the language mappings are from org.eclipse.cdt.core.cSource -> vendor.c.language id/name.
> 
> So, I don't think this change can be accepted as is.
> 
> Which toolset are you using? If you are using the MSVC integration that comes with CDT, then it needs that additional work to mark the language mappings as MSVC C and C++.
> 
> Am I understanding the problem correctly? Please feel free to share your thoughts on this.

Ah, seems like it's been hardcoded from the start. That's unfortunate.

I typically use the default CDT setup, which was where I first noticed that while switching to the Visual C++ integration (as a test for something unrelated) that the language name would erroneously remain as "GNU C/C++", and assumed it was a visual error that could be rather easily fixed with a quick config switch. Seems like this isn't the case (I didn't realize that the language name was actually used as the marker for choosing the language implementation), so I apologize for the noise. That aside, seems a little strange to me that the setting for the language name is `language.name.gcc` in what I assume is the core part of CDT however.

> > this is also flat out wrong in other cases, such as when Workspaces configured with the Microsoft Visual C++ compiler (one that couldn't possibly be more incompatible with the actual GNU dialect) are also labelled as "GNU C".
> 
> However CDT _is_ treating those C and C++ files as GNU C and GNU C++ files. i.e. standard CDT won't handle MSVC extensions (generally) because no one has contributed a subclass of AbstractCLikeLanguage for MSVC. Therefore CDT will use the GNU version.

Speaking of Visual C++, CDT has handled quite a number of its extensions quite well actually, so I'm surprised to hear that support for it is spotty at best

Independently of this issue though, it might be a good idea to also move assigning of language backends to also use the language id instead of the display name as has already been done with https://github.com/eclipse-cdt/cdt/commit/fa8a6279d26c357a502f067b94f57a3614954a34, and I'd be happy to help do that, as well as fixing the display name for Visual C++ as well. Out of curiosity, where does this actually happen? All I can find is ATTR_NAME in LanguageSettingsExtensionManager and a few related classes, but nothing really concrete about how the language provider is selected.

> BTW are you an end user or a vendor integrating CDT?

I'm a user of CDT, currently working on [introducing basic support for CDT and its counterpart Eclipse JDT into OpenJDK](https://github.com/openjdk/jdk/pull/10853)

TheShermanTanker:
> The complication is that how llvm/visual C++ has been supported is by making them look like GNU tool chains, rather than by making them first class citizens (like ARM, Renesas, IAR etc do).

I think I know what my next commit will be ;)

> AFAIK that is how it already happens with IDs. The .cproject will store a language mapping that maps from a content type (e.g. C Source file) to a language implementation (e.g. GNU C language). For example, if I can map C source files (.c extension) to be analysed as a C++ file by changing the language mappings. This creates the following snippet of XML in the .cproject file: (This can be useful with projects like GDB, which is now written in C++ but kept file names as .c)

That raises an interesting question: Can one (in theory) assign the language to a specific implementation by tampering with the xml in .cproject irrespective of what toolchain is being used? (say, something ridiculous like marking a workspace compiled by gcc as "MSVC C")

jonahgraham:
> That raises an interesting question: 

Yes. But you don't need to hack the .cproject, just change the language mappings in the project properties.

TheShermanTanker:
> > That raises an interesting question:
> 
> Yes. But you don't need to hack the .cproject, just change the language mappings in the project properties.

Thanks. Can I also define custom languages this way, by any chance? (Eg copying the language definition block in plugin.xml and putting it into the .cproject or language.settings.xml)

jonahgraham:
> Can I also define custom languages this way, by any chance? 

I don't think so, I think only the ID can be in the .cproject and that ID [needs to be defined](https://help.eclipse.org/latest/topic/org.eclipse.cdt.doc.isv/reference/extension-points/org_eclipse_cdt_core_language.html) in the plugin.xml.

PR #179, opened 2022-11-28
Bug: Docker: Save correct docker image name

cmorty:
When the docker image was not locally available the name of the image was not correctly saved.

@jonahgraham @jjohnstn : I fixed this one before, but somehow it got lost during rebasing. Sorry for not testing it earlier.


jonahgraham:
@jjohnstn this looks like it should go in 11.0.0 which I am building today. If you have no objections I will merge this before I do the build.

@cmorty Thanks for testing before it is too late! It is easy to respin through EOD Wednesday. After that it needs to be a critical bug to respin.

jjohnstn:
@jonahgraham LGTM, go ahead.

PR #181, opened 2022-11-28
Bug: Docker: Do not pull all tags

cmorty:
If the latest tag did not exist locally and no tag was given for an image, all tags were pulled.

...things you find when preparing a talk on how to use Eclipse CDT & Docker... :)
It would be nice to have this in 2022-12, but I think it's ok when it does not make it.

Related to https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/issues/140

### DIFF
--- a/launch/org.eclipse.cdt.docker.launcher.tests/src/org/eclipse/cdt/docker/launcher/tests/DockerServerInteractions.java
+++ b/launch/org.eclipse.cdt.docker.launcher.tests/src/org/eclipse/cdt/docker/launcher/tests/DockerServerInteractions.java
@@ -13,6 +13,7 @@ package org.eclipse.cdt.docker.launcher.tests;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.notNullValue;
 import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNull;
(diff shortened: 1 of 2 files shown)


jjohnstn (approved):
Looks ok.  I think a test using a repo name that has a port specified would be advisable in the future but I release that time is tight.

jonahgraham:
> Looks ok. I think a test using a repo name that has a port specified would be advisable in the future but I release that time is tight.

Thanks @jjohnstn 

@cmorty I'm going to hold off doing the RC2 build until tomorrow, so let me know if you want me to merge this as is, or take into account Jeff's comment.

jjohnstn:
@jonahgraham I think you should merge it.  The test is not important as the logic takes the port number into account.

cmorty:
@jjohnstn Are you happy with the current solution? Should I write a test? Add documentation on why the check should be sufficient? Or is https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/issues/143 the way to go?

I thought about adding validation to the input and forcing a tag. But then again, "latest" is the default. Also, when validating the input https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/issues/143 would be nice. :)

jjohnstn:
@cmorty I think adding validation would be a good idea.  We can make an API available in 2023-03 M1.

Issue #167, opened 2022-11-15
Core build launch configurations are not persistent.


ewaterlander:
I noticed the following with core build cmake and make projects.

Create a Core Build Makefile project. Change Build Output Location to "Build in Project directory".
When I close the project and re-open it, the build settings are back to default. The Build Output Location goes back to "Build configuration in specific directory". Or in case of a Cmake project, the generator goes back from Makefile to Ninja.

When I close the project, exit Eclipse, open Eclipse, open the project, the launch configuration disappeared.

The core build launch configurations are only visible in in the launch bar, not in the Run/Debug Configurations dialogs.
The core build launch configurations are of launch-type "Auto C/C++ Local Application" (org.eclipse.cdt.debug.core.localCoreBuildLaunchConfigType). They are present in the .metadata folder, but they contain no build settings. I think that's why the build information is lost when you close the project.
The only way to restore a disappeared launch configuration is to delete the project and recreate a new project op top of the old location.

I would expect that the core build launch configurations would keep their settings and would not disappear.

OS: Linux
CDT: 10.5

ewaterlander:
CDT 10.7.1 still has the problem.

ewaterlander:
Isn't a launch configuration a strange place to store build configuration settings (besides the fact that they are only in memory, not in the launch file)?
In a managed build project the build configuration settings are stored in the .cproject file.

mbooth101:
Core build was an initiative to simplify build mechanism especially in the case where the build system took care of most things internally (e.g. Makefile, CMake, Meson). These build systems are things that CDT must "launch", I don't think it's very strange to store configuration for things we must launch in launch configurations.

Documentation is a bit sparse, but here's one of Doug's old slide decks on the off-chance it is helpful: https://wiki.eclipse.org/images/e/e2/CDTCoreBuild2017.pdf

*However* you certainly found a bug here I think because when I change the build output location, the compiled binary artifact is not detected (there is no binaries container in the project explorer) and cannot be launched (I believe the dynamically created launch configurations rely on being able to detect the built binary artifact in order to avoid any additional user configuration).

ewaterlander:
I think this bug is related: https://bugs.eclipse.org/bugs/show_bug.cgi?id=552070

ewaterlander:
The fix for bug 552070 fixed the problem of the settings getting back to default after close and reopen of the project.
See PR https://github.com/eclipse-cdt/cdt/pull/192

[...]

PR #192, opened 2022-12-05
Bug 552070 - Build not configured correctly error while building a pr…

ewaterlander:
…oject

https://bugs.eclipse.org/bugs/show_bug.cgi?id=552070

After closing and opening a CMake or Makefile Core Buil project, the project could not be built for Debug anymore. Error: "Build not configured correctly".

The CBuildConfigurationManager calls
provider.getCBuildConfiguration(buildConfig, configName), which starts a recreation of the ICBuild and IBuild configurations. For the Debug configurations this led to a call to
CBuildConfiguration(IBuildConfiguration config, String name), which always fails with a CoreException. The CBuildConfigurationManager silently catched the exception and put the IBuildConfiguration for debug in the noConfigs list.

With this change the CBuildConfigurationProviders will for both Run and Debug find a toolchain and create build configurations for the correct launch mode (run or debug).

### DIFF
--- a/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
+++ b/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
@@ -59,6 +59,7 @@ import org.eclipse.cdt.core.parser.IScannerInfoChangeListener;
 import org.eclipse.cdt.core.resources.IConsole;
 import org.eclipse.cdt.internal.core.BuildRunnerHelper;
 import org.eclipse.cdt.internal.core.ConsoleOutputSniffer;
+import org.eclipse.cdt.internal.core.build.CBuildConfigurationManager;
 import org.eclipse.cdt.internal.core.build.Messages;
 import org.eclipse.cdt.internal.core.model.BinaryRunner;
 import org.eclipse.cdt.internal.core.model.CModelManager;
@@ -94,6 +95,19 @@ import com.google.gson.GsonBuilder;
 /**
  * Root class for CDT build configurations. Provides access to the build
  * settings for subclasses.
+ *
+ * Each Eclipse project has one or more build configurations ({@link IBuildConfiguration}).
+ * A CDT Core Build project pairs each build configuration with a Core Build configuration
+ * ({@link ICBuildConfiguration}). A Core Build configuration has variable config pointing to
+ * the IBuildConfiguration. The link from IBuildConfiguration to ICBuildConfiguration
+ * goes via getAdapter(ICBuildConfiguration.class) which gets the ICBuildConfiguration
+ * from Map configs in the {@link CBuildConfigurationManager}.
+ *
+ * In a new project the initial Core Build configurations creation is triggered by the
+ * {@link CoreBuildLaunchBarTracker}. The initial configuration for Debug will only be
+ * created if the user selects the Debug launch mode. Restoration of Core Build configurations,
+ * after an Eclipse restart or close and open of the project, uses the settings which are
+ * persistently stored in the backing store. @see org.osgi.service.prefs.Preferences
  *
  * @since 6.0
  */
--- a/core/org.eclipse.cdt.core/src/org/eclipse/cdt/internal/core/build/CBuildConfigurationManager.java
+++ b/core/org.eclipse.cdt.core/src/org/eclipse/cdt/internal/core/build/CBuildConfigurationManager.java
@@ -367,7 +367,12 @@ public class CBuildConfigurationManager
 					Preferences projectNode = parentNode.node(project.getName());
 					if (projectNode != null) {
 						try {
-							projectNode.removeNode();
+							if (event.getType() == IResourceChangeEvent.PRE_DELETE) {
+								// We need to keep the settings when the project is closed. They are used by
+								// CBuildConfiguration.CBuildConfiguration(IBuildConfiguration config, String name)
+								// to restore Debug core build configurations when the project is reopened.
+								projectNode.removeNode();
+							}
 							parentNode.flush();
 						} catch (BackingStoreException e) {
 							CCorePlugin.log(e);


jonahgraham:
@kolipakakondal Can you have a look at this and see if it addresses the initial issue you raised in [Bug 552070]
(https://bugs.eclipse.org/bugs/show_bug.cgi?id=552070)?

@ewaterlander Thanks for the PR. I'll try to have a look soon, hopefully @kolipakakondal can help me do the review to confirm it resolves the original problem. 

PS core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java has some FIXME comments, can you expand on what they mean please?

ewaterlander:
> PS core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java has some FIXME comments, can you expand on what they mean please?

The fix is actually to not use the contructor CBuildConfiguration(IBuildConfiguration config, String name) anymore in the CBuildConfigurationProviders. This constructor is the place where is starts to go wrong. Because I'm not using the constructor anymore I did not fix it, but I added FIXMEs.  First of all the Preferences settings is empty, and also the get statements, using keys TOOLCHAIN_TYPE and TOOLCHAIN_ID, make no sense to me. What I see in the debugger is that this method always throws a CoreException.

I don't know if somebody sees any use for the constructor, but I think this constructor should go completely. It is fixed to the "run" launch mode, but the "run" build configs are not created with this. The providers use it to recreate "debug" build configs with it!

Because of the exception the  CBuildConfigurationManager puts the Debug build configuration in the noConfigs list. Once it is there there is no escape, but to exit and restart eclipse. Then the CBuildConfigurationProviders will use the other methods to create the build configurations.

ewaterlander:
The second FIXME is about that a second identical call to the toolchain manager is done to get the toolchain. This is not checking for other versions as the comment suggests.

kolipakakondal:
Ok @jonahgraham, I will take a look at this.

kolipakakondal:
Hi @ewaterlander @jonahgraham I got a chance to check this today.

**Make project:**

- Create > Choose Run > Build > Close the project > Re-open it > Build again > Works fine
- Create > Choose Debug > Build > Close the project > Re-open it > Build again > Works fine

**CMake project:**

- Create > Choose Run > Build > Close the project > Re-open it > Build again > Works fine
- Create > Choose Debug > Build > Close the project > Re-open it > Build again > Error out `Build not configured correctly` Error

Please find the recording here http://g.recordit.co/SFiInZN1vs.gif

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeBuildConfigurationProvider.java
```diff
@@ -42,48 +42,45 @@ public String getId() {
...
+				ICMakeToolChainFile file = manager.getToolChainFileFor(toolChain);
+				if (file == null) {
+					Collection<ICMakeToolChainFile> files = manager.getToolChainFilesMatching(properties);
+					if (!files.isEmpty()) {
+						file = files.iterator().next();
+					}
+				}
+				if (file != null) {
+					return new CMakeBuildConfiguration(config, name, toolChain, file, "debug"); //$NON-NLS-1$
+				}
```
kolipakakondal, line 79: This is enforcing the user to have a mandatory ICMakeToolChainFile for building in "Debug" launch mode and I believe that's not mandatory.

If we takeout `if (file != null) ` condition then we could avoid the Build not configured error in the case of the Build+Debug launch mode scenario for CMake projects.

ewaterlander: You are right. It is OK if file == null.
I update the PR.
Thanks.

jonahgraham:
Thanks @kolipakakondal for having a look.

@ewaterlander / @kolipakakondal when the two of you are satisfied with the changes I will be happy to review and merge it.

kolipakakondal:
Hi @ewaterlander @jonahgraham LGTM.

ewaterlander:
Is there anything that blocks the merge of this PR?

core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
```diff
@@ -128,11 +128,12 @@ protected CBuildConfiguration(IBuildConfiguration config, String name) throws Co
 		String typeId = settings.get(TOOLCHAIN_TYPE, ""); //$NON-NLS-1$
 		String id = settings.get(TOOLCHAIN_ID, ""); //$NON-NLS-1$
 		IToolChainManager toolChainManager = CCorePlugin.getService(IToolChainManager.class);
+		// FIXME: typeId and id are always empty string, and therefore tc is always null.
```
jonahgraham, line 131: I still don't understand this comment. I can see typeId and id being non-empty string without your change applied if I do this:

1. Create CMake project
2. Build project
3. Restart Eclipse

At this point the typeId/id will be non-empty.

To me this indicates that the bug fix has missed a use case that may or may not be broken with this PR.

>  It is fixed to the "run" launch mode, but the "run" build configs are not created with this.

I don't think this statement is true either. Using the above flow launchMode == debug

core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
```diff
@@ -128,11 +128,12 @@ protected CBuildConfiguration(IBuildConfiguration config, String name) throws Co
...
 		IToolChain tc = toolChainManager.getToolChain(typeId, id);
 
 		if (tc == null) {
 			// check for other versions
-			tc = toolChainManager.getToolChain(typeId, id);
+			tc = toolChainManager.getToolChain(typeId, id); // FIXME: same call as above.
```
jonahgraham, line 136: Yes, lets just fix it.

jonahgraham (commented):
> Is there anything that blocks the merge of this PR?

My vacation schedule coupled with the fact no one "owns" this code.

Your change seems ok, but incomplete. Does the AutotoolsBuildConfigurationProvider and MesonBuildConfigurationProvider not suffer from the same problem? The pattern of these methods seem the same in all 4 classes.

When I tried to do a quick review on this last week I got hung up by the FIXME comments. I have now loaded up this change in the IDE and my findings don't match the comments

core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
ewaterlander: No problem.
I'm mainly focused on Makefile flow. I did not look at Autotools and Meson. I never used it. I can have look.
I may have overlooked a use case. I will look again.

ewaterlander: I quickly tried Autotools and Meson.
For Meson the problem is similar as in Makefile and CMake.
For Autotools you can get the problem even without closing the project. First compile for Run end then for Debug gives the problem. If you compile first for Debug and then for Run it compiles OK. After closing and opening the project you get the same problem as with the others.

ewaterlander:
I will be offline for more than a week.

core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
ewaterlander: > I still don't understand this comment. I can see typeId and id being non-empty string without your change applied if I do this:
> 
> 1. Create CMake project
> 2. Build project
> 3. Restart Eclipse
> 
> At this point the typeId/id will be non-empty.

You are right. When you restart Eclipse it works as expected. The problem occurs when you close the project and reopen the project. You don't need to exit Eclipse. When you reopen the project the settings have no properties. This is the root of the problem. Then typeId and id will be empty string. tc will be null, and an exception is thrown. And then the CBuildConfigurationManager puts the configuration in the noConfigs list. After that it will not be possible to build the project for debug.

I don't know (yet) why the settings properties are lost when the project is closed. My fix is to avoid using this constructor, because it also not used for the Run launch mode. Fixing the loss of the settings could be a better fix. But why was it not used for Run mode?

ewaterlander: I have changed the patch to keep the build configuration's settings at project closure. This fixes the problem for all Core Build project types. I added some comment to CBuildConfiguration to make it a bit easier to understand.

ewaterlander:
Hi,
Could my latest change be reviewed?

jonahgraham (approved):

jonahgraham:
Thank you for this change - the current iteration was fairly straightforward to review and it looks like a reasonable solution to the problem.

PS. Sorry for the delay in reviewing, bad timing of Christmas vacations + critical work delayed my ability to come back to this sooner.

ewaterlander:
> Thank you for this change - the current iteration was fairly straightforward to review and it looks like a reasonable solution to the problem.
> 
> PS. Sorry for the delay in reviewing, bad timing of Christmas vacations + critical work delayed my ability to come back to this sooner.

No problem! Thanks for accepting my change.

I'm going to see if the Run configurations can be restored the same way. Currently after Eclipse restart or project close/open Run configurations are created newly, not restoring existing preferences. I see no reason why Run and Debug configs should be restored differently. I also suspect there is a link with  https://github.com/eclipse-cdt/cdt/issues/167

PR #199, opened 2022-12-07
Add Cygwin.java and MSYS2.java

xu-chiheng:
To replace the original Cygwin.java, to do path translations without invoking external program "cygpath", to improve performance and reliability.


core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/MSYS2.java
```diff
@@ -0,0 +1,164 @@
...
+			Map<String, String> environment = System.getenv();
+			rootDir = environment.get("MSYS2_DIR"); //$NON-NLS-1$
+			if (rootDir != null) {
+				if (dirHasMsys2Dll(rootDir)) {
+					present = true;
+					rootDir = new Path(rootDir).toPortableString();
+					return;
+				}
+			}
+			for (char drive = 'C'; drive < 'H'; drive++) {
```
jonahgraham, line 39: What makes `H` special here? Why not check all drives?

core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/MSYS2.java
```diff
@@ -0,0 +1,164 @@
...
+
+	// initialize static data fields
+	private static void initialize() {
+		if (Platform.getOS().equals(Platform.OS_WIN32)) {
+			Map<String, String> environment = System.getenv();
+			rootDir = environment.get("MSYS2_DIR"); //$NON-NLS-1$
+			if (rootDir != null) {
+				if (dirHasMsys2Dll(rootDir)) {
+					present = true;
+					rootDir = new Path(rootDir).toPortableString();
```
jonahgraham, line 35: toPortableString is only for use when you are later calling fromPortableString. See [javadoc](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/core/runtime/IPath.html#toPortableString()) for more info, key part quoted here for convenience:

> Returns a platform-neutral string representation of this path. The format is not specified, except that the resulting string can be passed back to the Path#fromPortableString(String) constructor to produce the exact same path on any platform.

I think you want use [toString](https://help.eclipse.org/latest/topic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/core/runtime/IPath.html#toString()) which is explicitly documented as using `/` which is my assumption as to why you are using toPortableString:

> Returns a string representation of this path, including its device id. The same separator, "/", is used on all platforms.

core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/MSYS2.java
```diff
@@ -0,0 +1,164 @@
...
+		IPath dirLocation = new Path(dirString);
+		File dir = dirLocation.toFile();
+		if (dir.isAbsolute() && dir.exists() && dir.isDirectory()) {
+			File file = dirLocation.append("/usr/bin/msys-2.0.dll").toFile(); //$NON-NLS-1$
+			if (file.exists() && file.isFile() && file.canRead()) {
+				return true;
+			}
+		}
+		return false;
+	}
```
jonahgraham, line 63:
```suggestion
				IPath path = new Path(drive, "msys64"); //$NON-NLS-1$
				if (dirHasMsys2Dll(dirString)) {
					present = true;
					rootDir = path.toString();
					return;
				}
			}
		}
	}

	private static boolean dirHasMsys2Dll(IPath path) {
		File dir = path.toFile();
		if (dir.isAbsolute() && dir.exists() && dir.isDirectory()) {
			File file = path.append("/usr/bin/msys-2.0.dll").toFile(); //$NON-NLS-1$
			if (file.exists() && file.isFile() && file.canRead()) {
				return true;
			}
		}
		return false;
	}
```

core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/MSYS2.java
```diff
@@ -0,0 +1,164 @@
...
+
+	public boolean isPresent() {
+		return present;
+	}
+
+	public String getRootDir() {
+		return rootDir;
+	}
+
+	// Convert Unix path to Windows path
```
jonahgraham, line 73: New API should have Javadocs.

core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/MSYS2.java
```diff
@@ -0,0 +1,164 @@
...
+			return unixPath;
+		}
+		if (unixPath == null || unixPath.trim().length() == 0) {
+			return unixPath;
+		}
+
+		String windowsPath;
+		IPath path = Path.fromOSString(unixPath);
+		if (path.getDevice() != null) {
+			// already a windows absolute path
```
jonahgraham, line 85: A device doesn't mean an absolute path on Windows. There is Path.isAbsolute() that will confirm that. Also, not sure if/how this code handles UNC, but perhaps that is out of scope for now.

e.g. `c:file.c` will be identified as absolute according to this code, but it refers to the `file.c` in the cwd of the C: drive.

core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/Cygwin.java
```diff
@@ -0,0 +1,175 @@
...
+
+import java.io.File;
+import java.util.Map;
+
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.Platform;
+
+// A collection of Cygwin-related utilities.
+public class Cygwin {
```
jonahgraham, line 20: This codes doesn't really belong here, there is already [Cygwin.java](https://github.com/eclipse-cdt/cdt/blob/main/core/org.eclipse.cdt.core/utils/org/eclipse/cdt/internal/core/Cygwin.java) that it should live in.

AFAICT this code is making a bunch of assumptions that should be deferred to cygwin to make, using tools like cygpath as done by org.eclipse.cdt.internal.core.Cygwin.runCygpath(String[])

For example, this code is making assumptions about cygwin mounts and where to find cygwin itself.

core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/Cygwin.java
```diff
@@ -0,0 +1,175 @@
...
+							return windowsPath;
+						}
+					}
+				}
+				if (segments[0].equals("usr") && (segments[1].equals("bin") || segments[1].equals("lib"))) { //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$
+					// /usr/lib --> /lib
+					// /usr/bin --> /bin
+					// /usr/include unchanged
+					String[] newSegments = new String[segments.length - 1];
+					System.arraycopy(segments, 1, newSegments, 0, segments.length - 1);
```
jonahgraham, line 116: (non-blocker) Please try to avoid System.arraycopy in new code. It is much harder to maintain than using List<> and toArray.

jonahgraham (changes requested):
Some of the review comments apply in multiple places, but I have avoided unnecessarily duplicating comments that say the same thing.

Please provide a summary of what this code does in the commit message or PR. "Add abc.java" is not a sufficient commit message.

PR #209, opened 2022-12-09
Replace toOSString with toString

xu-chiheng:


jonahgraham:
I noticed some false test reports in this PR, it was down to new version of Ubuntu/GDB. See #210

PR #216, opened 2022-12-09
Replace toOSString with toString(failed)

xu-chiheng:


jonahgraham:
Can you please stop submitting these PRs until you find a committer to review this change. I don't think I was explicit enough before, but I will not be reviewing a broad change that changes toOSString -> anything.

I will only consider reviewing specific changes.

PS If you want to run tests, do so as much as you like, enable github actions on your fork and then you can run all the tests you want.

PR #252, opened 2023-01-19
Handle case where AST name has no location

### LINKED ISSUE #251: Computing hover expression has encountered a problem
**Describe the bug**

Received a pop-up error when hovering mouse over some code while debugging:

image

**To Reproduce**
1. Create a project
2. Make the contents of the source file:

```c
struct s {
	int f;
};

#define macro(a) a

int main(void) {
	struct s x;
	macro(x.f); // hover over the f here
	return 0;
}
```
3. Add a macro for `x=z`:

image

4. Build and start a debug session
5. Hover over the `f` in the code as indicated in the code comment (don't hover over the f in the comment)

jonahgraham:
In cases where an IASTName has no image location we were getting NPEs in this code. See javadoc for getImageLocation for cases when image location can be null. See #251's description for a full case of when this can happen.

All other calls to IASTName.getImageLocation in CDT were also checked and this was the only place in the code where the return value was not checked for null.

Fixes #251

### DIFF
--- a/debug/org.eclipse.cdt.debug.ui/META-INF/MANIFEST.MF
+++ b/debug/org.eclipse.cdt.debug.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.cdt.debug.ui; singleton:=true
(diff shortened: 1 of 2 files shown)


ddscharfe (approved):
LGTM.

Maybe consider adding the changes proposed here: [ #251](https://github.com/eclipse-cdt/cdt/issues/251#issuecomment-1398467288)

Issue #254, opened 2023-01-19
Constructors of inner classes which are declared in the implementation are not resolved properly when using 'new' expressions


ddscharfe:
**Bug Description**
Consider the following simple example (struct is used for simplicity, the problem also applies to classes):

MyClass.h
```C++
#ifndef MYCLASS_H_
#define MYCLASS_H_

struct MyClass {
	MyClass();
	struct MyInnerClass;
};

#endif /* MYCLASS_H_ */
```

MyClass.cpp
```C++
#include "MyClass.h"

struct MyClass::MyInnerClass {
	MyInnerClass(bool a, bool b) {
	}
};

MyClass::MyClass() {
	new MyInnerClass(true, true);
}
```

This is correct C++, but the constructor which is declared in the .cpp file is not resolved properly:
![image](https://user-images.githubusercontent.com/34916836/213426984-f5d0a0e3-1b03-4630-a059-88740ac965ba.png)
This happens only if `MyInnerClass` is instantiated using the `new` operator, e.g. `MyInnerClass inner(true, true)` will work fine.

**Explaination**
I debugged this issue and will try to provide a fix soon. My current findings are best illustrated by the following stack trace:
![image](https://user-images.githubusercontent.com/34916836/213429528-191a5dd6-5cf6-4c85-a56e-3342e2baefc3.png)
`CPPSemantics::resolveAmbiguities` tries to find the binding for the constructor, by analyzing all constructors. Constructors which are not declared yet, are filtered out (`CPPSemantics::2228`). 
The constructors are indexed, hence they are `IPDOMBinding`s. `CPPSemantics::declaredBefore` treats `ICPPInternalBinding`s and `IPDOMBinding`s differently. In the latter case the implementation will use the `IIndexFileSet` of the `IASTTranslationUnit` and check if it contains the currently analyzed constructor.
When the constructor is declared in the header file, it will be found in the index file set, **but in our case it won't**. As a result, the constructor candidate is filtered out and not passed to the following logic.

**Fun Fact**
When there is another constructor defined in the header file, which has a different signature, but the same number of arguments, the bug disappears:
```C++
#ifndef MYCLASS_H_
#define MYCLASS_H_

struct MyClass {
	MyClass();
	struct MyInnerClass {
		MyInnerClass(int a, int b) {
		}
	};
};
#endif /* MYCLASS_H_ */
```
[...]

PR #255, opened 2023-01-20
Issue #254: Constructors of inner classes not resolved properly

ddscharfe:
- add test case to reproduce #254
- add special case to detect constructors which were declared in the index file set of the tu

### DIFF
--- a/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/cpp/semantics/CPPSemantics.java
+++ b/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/cpp/semantics/CPPSemantics.java
@@ -2110,7 +2110,14 @@ public class CPPSemantics {
 				}
 				IASTTranslationUnit tu = node.getTranslationUnit();
 				IIndexFileSet indexFileSet = tu.getIndexFileSet();
-				return (indexFileSet != null && indexFileSet.containsDeclaration(indexBinding));
+				if (indexFileSet != null && indexFileSet.containsDeclaration(indexBinding)) {
+					return true;
+				} else if (indexBinding instanceof ICPPConstructor) {
+					IIndexFileSet astFileSet = tu.getASTFileSet();
+					return astFileSet != null && astFileSet.containsDeclaration(indexBinding);
+				} else {
+					return false;
+				}
 			}
 		}
 		return pointOfDecl < pointOfRef;
--- a/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/internal/index/tests/IndexCPPBindingResolutionBugsTest.java
+++ b/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/internal/index/tests/IndexCPPBindingResolutionBugsTest.java
@@ -1397,4 +1397,21 @@ public abstract class IndexCPPBindingResolutionBugsTest extends IndexBindingReso
 		ast = workingCopy.getAST(strategy.getIndex(), ITranslationUnit.AST_SKIP_INDEXED_HEADERS);
 		checkBindings(ast);
 	}
+
+	// struct MyClass {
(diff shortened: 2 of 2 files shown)


i-garrison:
> * add special case to detect constructors which were declared in the index file set of the tu

I found now that if you move the definition of `struct MyClass::MyInnerClass` below the use in `MyClass::MyClass()` constructor, CDT is able to resolve reference to `MyInnerClass` even though it is now an incomplete type.

Hope this is intentional (and probably fine as you will get compiler error after performing a build.)
Another fun fact is that before this change CDT was able to resolve constructor call in both cases:
```c++
#include "pr-255-header-defining-MyClass"
// inner class is forward-declared at this point
MyClass::MyClass() {
	MyInnerClass(true, true); // << resolved with and without this change
	new MyInnerClass(true, true); // << now resolved with this change too
}
// inner class is defined later
struct MyClass::MyInnerClass {
	MyInnerClass(bool a, bool b) {
	}
};
```

ddscharfe:
> I found now that if you move the definition of `struct MyClass::MyInnerClass` below the use in `MyClass::MyClass()` constructor, CDT is able to resolve reference to `MyInnerClass` even though it is now an incomplete type.
> 
> Hope this is intentional (and probably fine as you will get compiler error after performing a build.)

Well... It's not intentional tbh, but I couldn't find a better way to solve the problem. Not sure if it's okay to fix this one bug by introducing an imho slightly less bad one.
There are some failing checks anyways, so I converted the PR to a draft just for now.

i-garrison:
> Not sure if it's okay to fix this one bug by introducing an imho slightly less bad one.

There is no _new_ bug introduced by this PR. Name resolution behavior with class definition after use existed before this PR and is not affected by these changes at all as far as I can see - so if that needs fixing it would be separate issue anyway.

> There are some failing checks

I cannot reproduce these errors locally, maybe they are transient?

ddscharfe:
> There is no _new_ bug introduced by this PR. Name resolution behavior with class definition after use existed before this PR and is not affected by these changes at all as far as I can see - so if that needs fixing it would be separate issue anyway.

Okay, I got you wrong then, thanks for checking! I will set this PR to "Ready for review" again.
 
> > There are some failing checks
> 
> I cannot reproduce these errors locally, maybe they are transient?

Me neither, maybe @jonahgraham can restart them.

core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/cpp/semantics/CPPSemantics.java
```diff
@@ -2110,7 +2110,14 @@ public static boolean declaredBefore(Object obj, IASTNode node, boolean indexBas
 				}
 				IASTTranslationUnit tu = node.getTranslationUnit();
 				IIndexFileSet indexFileSet = tu.getIndexFileSet();
-				return (indexFileSet != null && indexFileSet.containsDeclaration(indexBinding));
+				if (indexFileSet != null && indexFileSet.containsDeclaration(indexBinding)) {
+					return true;
+				} else if (indexBinding instanceof ICPPConstructor) {
+					IIndexFileSet astFileSet = tu.getASTFileSet();
+					return astFileSet != null && astFileSet.containsDeclaration(indexBinding);
```
i-garrison, line 2117: I confirm this change resolves the issue.

i-garrison (approved):

ddscharfe:
@i-garrison Thanks for reviewing

jonahgraham (approved):

jonahgraham:
> > I cannot reproduce these errors locally, maybe they are transient?
> 
> Me neither, maybe @jonahgraham can restart them.

One of the items is #259, and as you both verified it works locally, I won't bother retriggering. Instead I filled #264.

jonahgraham:
Thanks for the fix @ddscharfe. Thanks for doing a review @i-garrison.

PR #260, opened 2023-01-26
Fix copyright header typo

i-garrison:
... ` and others` clause is already there on the following line.

### DIFF
--- a/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/cpp/semantics/EvalPackAccess.java
+++ b/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/cpp/semantics/EvalPackAccess.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2016,2022 Institute for Software, HSR Hochschule fuer Technik and others
+* Copyright (c) 2016,2022 Institute for Software, HSR Hochschule fuer Technik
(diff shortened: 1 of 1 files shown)


core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/cpp/semantics/EvalPackAccess.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
-* Copyright (c) 2016,2022 Institute for Software, HSR Hochschule fuer Technik and others
+* Copyright (c) 2016,2022 Institute for Software, HSR Hochschule fuer Technik
```
jonahgraham, line 2: Thanks for fixing this up, I guess the next line should actually be deleted?

i-garrison: No, I believe this should be kept as, two lines actually match attributions in header of `EvalCompositeAccess.java` except for new years of contribution.

jonahgraham: Yes, I see and understand now. Thanks.

jonahgraham (approved):

Issue #262, opened 2023-01-27
ConcurrentModificationException in LaunchBarManager.storeActiveDescriptor()


trancexpress:
**Describe the bug**
During our ARTs, we encountered the following CME:

```
java.util.ConcurrentModificationException
	at java.base/java.util.LinkedHashMap$LinkedHashIterator.nextNode(LinkedHashMap.java:756)
	at java.base/java.util.LinkedHashMap$LinkedKeyIterator.next(LinkedHashMap.java:778)
	at org.eclipse.launchbar.core.internal.LaunchBarManager.storeActiveDescriptor(LaunchBarManager.java:511)
	at org.eclipse.launchbar.core.internal.LaunchBarManager.setActiveLaunchDescriptor(LaunchBarManager.java:487)
	at org.eclipse.launchbar.core.internal.LaunchBarManager.addDescriptor(LaunchBarManager.java:309)
	at org.eclipse.launchbar.core.internal.LaunchBarManager.launchObjectAdded(LaunchBarManager.java:358)
	at org.eclipse.launchbar.core.internal.LaunchBarManager.launchConfigurationAdded(LaunchBarManager.java:899)
	at org.eclipse.debug.internal.core.LaunchManager$ConfigurationNotifier.run(LaunchManager.java:219)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
	at org.eclipse.debug.internal.core.LaunchManager$ConfigurationNotifier.notify(LaunchManager.java:209)
	at org.eclipse.debug.internal.core.LaunchManager.launchConfigurationAdded(LaunchManager.java:2120)
	at org.eclipse.debug.internal.core.LaunchConfigurationWorkingCopy.writeNewFile(LaunchConfigurationWorkingCopy.java:375)
	at org.eclipse.debug.internal.core.LaunchConfigurationWorkingCopy.doSave0(LaunchConfigurationWorkingCopy.java:259)
	at org.eclipse.debug.internal.core.LaunchConfigurationWorkingCopy.doSave(LaunchConfigurationWorkingCopy.java:229)
	at org.eclipse.debug.internal.core.LaunchConfigurationWorkingCopy.doSave(LaunchConfigurationWorkingCopy.java:187)
        ...
        at org.eclipse.equinox.launcher.Main.main(Main.java:1440)
```

The above is triggered from one of our plug-ins, while creating a launch configuration (and saving it).

If I debug the respective test, I see also a job is trying to change the container in question:

[...]

Issue #259, opened 2023-01-26
testDCOM_BDFG (org.eclipse.cdt.ui.tests.text.doctools.DocCommentHighlightingTest) flaky


jonahgraham:
The flaky test:

`testDCOM_BDFG (org.eclipse.cdt.ui.tests.text.doctools.DocCommentHighlightingTest)`

with this output;

```
expected:<[offset: 149, length: 17, offset: 184, length: 18, offset: 223, length: 16, offset: 241, length: 17, offset: 258, length: 16, offset: 345, length: 16]> but was:<[]>
junit.framework.AssertionFailedError: expected:<[offset: 149, length: 17, offset: 184, length: 18, offset: 223, length: 16, offset: 241, length: 17, offset: 258, length: 16, offset: 345, length: 16]> but was:<[]>
	at org.eclipse.cdt.ui.tests.text.doctools.DocCommentHighlightingTest.testDCOM_BDFG(DocCommentHighlightingTest.java:211)
```

i-garrison:
This looks like a race in the test.
Adding a couple of sleeps and stepping through in debugger I seldom reproduce the issue with empty result from `findRangesColored()` and also managed to get different failure in nearby test `testDCOM_B` 
```
junit.framework.AssertionFailedError:
    expected:<[offset: 149, length: 17, offset: 345, length: 16]>
     but was:<[offset: 129, length: 18, offset: 328, length: 17]>
```

jonahgraham:
Thanks @i-garrison for the investigation.

This is now failing basically all the time on GH Actions. I am now looking at this, either full fix, or apply ["flaky"](https://github.com/eclipse-cdt/cdt/blob/main/TESTING.md#marking-tests-as-slow-or-flaky) to it.

jonahgraham:
The key thing is that makes `testDCOM_BDFG` different from the rest of the tests in `DocCommentHighlightingTest` is that it is run first. Therefore my guess is that some pre-existing state is not being cleared properly.

PR #263, opened 2023-01-27
Prevent CME in LaunchBarManager

trancexpress:
This change synchronizes access to `LaunchBarManager.descriptors`, to avoid a `ConcurrentModificationException` when adding descriptors for 2 launches at the same time.

Fixes: #262

### DIFF
--- a/launchbar/org.eclipse.launchbar.core/src/org/eclipse/launchbar/core/internal/LaunchBarManager.java
+++ b/launchbar/org.eclipse.launchbar.core/src/org/eclipse/launchbar/core/internal/LaunchBarManager.java
@@ -140,11 +140,13 @@ public class LaunchBarManager implements ILaunchBarManager, ILaunchTargetListene
 				ILaunchDescriptor last = null;
 				for (String id : split) {
 					Pair<String, String> key = toId(id);
-					ILaunchDescriptor desc = descriptors.get(key);
-					if (desc != null) {
-						descriptors.remove(key);
-						descriptors.put(key, desc);
-						last = desc;
+					synchronized (descriptors) {
+						ILaunchDescriptor desc = descriptors.get(key);
+						if (desc != null) {
+							descriptors.remove(key);
+							descriptors.put(key, desc);
+							last = desc;
+						}
 					}
 				}
 				// Set the active desc, with MRU, it should be the last one
@@ -304,7 +306,10 @@ public class LaunchBarManager implements ILaunchBarManager, ILaunchTargetListene
 	}
 
 	private void addDescriptor(Object launchObject, ILaunchDescriptor descriptor) throws CoreException {
-		descriptors.put(getDescriptorId(descriptor), descriptor);
+		Pair<String, String> descriptorId = getDescriptorId(descriptor);
+		synchronized (descriptors) {
+			descriptors.put(descriptorId, descriptor);
+		}
 		objectDescriptorMap.put(launchObject, descriptor);
 		setActiveLaunchDescriptor(descriptor);
 	}
@@ -371,7 +376,10 @@ public class LaunchBarManager implements ILaunchBarManager, ILaunchTargetListene
 		Activator.trace("launch object removed " + launchObject); //$NON-NLS-1$
 		ILaunchDescriptor descriptor = objectDescriptorMap.remove(launchObject);
 		if (descriptor != null) {
-			descriptors.remove(getDescriptorId(descriptor));
+			Pair<String, String> descriptorId = getDescriptorId(descriptor);
+			synchronized (descriptors) {
+				descriptors.remove(descriptorId);
+			}
 			if (descriptor.equals(activeLaunchDesc)) {
 				setActiveLaunchDescriptor(getLastUsedDescriptor());
 			}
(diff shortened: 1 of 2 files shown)


jonahgraham:
Failed test is unrelated, see #259

launchbar/org.eclipse.launchbar.core/src/org/eclipse/launchbar/core/internal/LaunchBarManager.java
```diff
@@ -475,8 +489,14 @@ public void setActiveLaunchDescriptor(ILaunchDescriptor descriptor) throws CoreE
...
 		}
-		if (descriptor != null && !descriptors.containsValue(descriptor)) {
-			throw new IllegalStateException(Messages.LaunchBarManager_1);
+		if (descriptor != null) {
+			boolean isContained;
+			synchronized (descriptors) {
+				isContained = descriptors.containsValue(descriptor);
+			}
+			if (!isContained) {
+				throw new IllegalStateException(Messages.LaunchBarManager_1);
```
jonahgraham, line 498: I am pretty sure you don't need to save off the `isContained` here, I think throwing from within the synchronized block is fine:

```java
syncrhonized(descriptors) {
    if (descriptor != null && !descriptors.containsValue(descriptor)) {
        throw new IllegalStateException(Messages.LaunchBarManager_1);
    }
}
```

If you need to optimize to not get the lock if `descriptor == null` then a partial split is fine too:

```java
if (descriptor != null) {
    syncrhonized(descriptors) {
        if (!descriptors.containsValue(descriptor)) {
            throw new IllegalStateException(Messages.LaunchBarManager_1);
        }
    }
}
```

launchbar/org.eclipse.launchbar.core/src/org/eclipse/launchbar/core/internal/LaunchBarManager.java
```diff
@@ -140,11 +140,13 @@ void init() throws CoreException {
 				ILaunchDescriptor last = null;
 				for (String id : split) {
 					Pair<String, String> key = toId(id);
-					ILaunchDescriptor desc = descriptors.get(key);
-					if (desc != null) {
-						descriptors.remove(key);
```
jonahgraham, line 145: I think the remove/put call is redundant?

jonahgraham (commented):
Can we simply change the maps to `ConcurrentHashMap` or wrap it in `Collections.synchronizedMap` (for the linked one)? That would mean that future editors of the code won't need to worry about getting it right as it would "just work" (moreso for concurrent hash map, less so for wrapper synchronized map)? That would address your other issues. I don't think there is a performance issue in the code of having concurrent hash maps.

I am happy with this change as is, but I'll hold off merging until you have a chance to consider my question.

jonahgraham (approved):
I'm going to approve these changes as they are now. Fixing up the code more widely can be a discussion for another day, but this change is a move in the right direction.

trancexpress:
> Can we simply change the maps to `ConcurrentHashMap` or wrap it in `Collections.synchronizedMap` (for the linked one)? That would mean that future editors of the code won't need to worry about getting it right as it would "just work" (moreso for concurrent hash map, less so for wrapper synchronized map)? That would address your other issues. I don't think there is a performance issue in the code of having concurrent hash maps.
> 
> I am happy with this change as is, but I'll hold off merging until you have a chance to consider my question.

Sorry I must have missed the e-mail notification for this comment.

I actually thought about making the container a synchronized one, but there are operations in `LaunchBarManager` which are odd to me. From what I can tell, in some places the code takes care to ensure the order in the `LinkedHashMap` is the correct one (i.e. the order the code needs) and this is done by first removing an element and then putting the new one... such sequences have to either still be in a synchronized block, or they have to be carefully replaced by single call APIs from the map. Since I'm not sure if the "replacement" methods of the map keep the order in the `LinkedHashMap` as expected (and even if they do, whether this is behavior as specified by documentation, or just an accident of implementation).

What I can do of course is keep the operations as they are, but wrap only the more complicated sequences with synchronized blocks. That will already reduce the amount of added code for this change.

My bigger concern was that synchronizing only on the one map here might not be solving race condition problems, just hiding the CMEs. As I've commented at the start of the PR. But since the change got merged I'll look into cleaning the code a bit. I guess better to not have the CMEs, even if not all problems are solved.

launchbar/org.eclipse.launchbar.core/src/org/eclipse/launchbar/core/internal/LaunchBarManager.java
trancexpress: I believe this is to keep the order in the `LinkedHashMap` as expected, since there is other code that depends on the order of adding elements to the map. It might be possible to replace the 2 calls with a single call to another API, though I'm not 100% the order will be the right one. It will have to be double-checked.

jonahgraham: Good point, I had overlooked that use of remove/put. I wouldn't spend more time on it than that.

jonahgraham:
Your explanation is good, thanks for taking the time to document it out. 

> I guess better to not have the CMEs, even if not all problems are solved.

I agree.

Issue #265, opened 2023-01-30
Type of arithmetic conversion is wrong for combinations of signed/unsigned types with different rank but same size


PhilippWendler:
There is a bug in the logic that CDT uses for computing the result type of a "usual arithmetic conversion" according to § 6.3.1.8 of the C standard, which is used for many operators.

The code for this is here (introduced by 16944a0de71433953e8078b854426ae327560f7d): https://github.com/eclipse-cdt/cdt/blob/b7fac2cfafa56a79ee1cea6ca8e6deb5416efbd6/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/ArithmeticConversion.java#L199-L210

The respective snippet of the C standard says (taken from C11 § 6.3.1.8 (1)):
> Otherwise, if the operand that has unsigned integer type has rank greater or equal to the rank of the type of the other operand, then the operand with signed integer type is converted to the type of the operand with unsigned integer type.
>
> Otherwise, if the type of the operand with signed integer type can represent all of the values of the type of the operand with unsigned integer type, then the operand with unsigned integer type is converted to the type of the operand with signed integer type.
>
> Otherwise, both operands are converted to the unsigned integer type corresponding to the type of the operand with signed integer type.

The first case is implemented correctly.
However, the second case uses the logic `signedRank.ordinal() > unsignedRank.ordinal()` to check "if the type of the operand with signed integer type can represent all of the values of the type of the operand with unsigned integer type". This is wrong for combinations of types that have different ranks but the same size, because then the signed type can *not* represent all values of the unsigned type.

This leads for example to wrong results from `getExpressionType()` on operators using arithmetic conversion.
Concrete cases where this produces a wrong result are `signed long` and `unsigned int` on 32-bit x86 (returned type is `signed long` but should be `unsigned long`) or `signed long long` and `unsigned long` on 64-bit x86.

As a fix, line 205 should compare the sizes of the two types. I think the assumption that a signed type with larger size than an unsigned type can represent all of the unsigned type's values holds.

Note:
- The last `return` statement in the above code is actually dead code right now (both if branches have exactly opposite conditions), also indicating that there is a bug here.
- The existence of this problem was already [acknowledged by the original submitter of the fix](https://bugs.eclipse.org/bugs/show_bug.cgi?id=231859#c7).

jonahgraham:
Thanks @PhilippWendler for the bug report. Are you considering providing a PR to test and fix this?

[...]

PR #269, opened 2023-01-30
Fix arithmetic conversion from larger unsigned operand type

i-garrison:
Fixes: 16944a0de7 ("Arithmetic conversions for unary and binary expressions, bug 231859.")
Closes: https://github.com/eclipse-cdt/cdt/issues/265

### DIFF
--- a/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/ArithmeticConversion.java
+++ b/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/ArithmeticConversion.java
@@ -425,6 +425,10 @@ public abstract class ArithmeticConversion {
 
 	/**
 	 * Checks whether a target integral type can represent all values of a source integral type.
+	 * <br><br>
+	 * If CPPSemantics current lookup point is set, size and alignment information is derived
+	 * from predefined type size macros available through current AST. If there is no current
+	 * lookup point, {@link #getApproximateSize(IBasicType)} is used to guess size of integral types.
 	 *
 	 * @param target the target integral type
 	 * @param source the source integral type
--- a/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/ArithmeticConversion.java
+++ b/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/ArithmeticConversion.java
@@ -201,8 +201,8 @@ public abstract class ArithmeticConversion {
 			return unsignedType;
 		}
 
-		// The signed has the higher rank.
-		if (signedRank.ordinal() > unsignedRank.ordinal()) {
+		// The signed has the higher rank, check if signed type can represent all unsigned values
+		if (fitsIntoType(signedType, unsignedType)) {
 			return signedType;
 		}
 
--- a/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/core/parser/tests/ast2/AST2Tests.java
+++ b/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/core/parser/tests/ast2/AST2Tests.java
@@ -6947,6 +6947,45 @@ public class AST2Tests extends AST2TestBase {
 		}
 	}
 
+	//  unsigned long int unsignedLongInt = 1;
+	//  signed long long int signedLongLongInt = 1;
+	//  unsigned long long var = 1;
+	//  void test() {
+	//      // with 64bit long and long long types expression type should be an unsigned long long int
+	//      var = signedLongLongInt + unsignedLongInt;
+	//  }
+	public void testTypeConversion_signedHigherRankSameSizeInts() throws Exception {
+		for (ParserLanguage lang : ParserLanguage.values()) {
+			IASTTranslationUnit ast = parseAndCheckBindings(getAboveComment(), lang, ScannerKind.STD);
+			IASTFunctionDefinition func = null;
+
+			for (IASTDeclaration d : ast.getDeclarations()) {
+				if (d instanceof IASTFunctionDefinition) {
+					func = (IASTFunctionDefinition) d;
+					break;
+				}
+			}
+			assertNotNull(func);
+
+			IASTStatement[] bodyStmts = ((IASTCompoundStatement) func.getBody()).getStatements();
+
+			// STD scanner configuration defines 64bit long and long long types (aka LP64),
+			// result should be an unsigned long long int
+			{
+				IASTBinaryExpression expr = (IASTBinaryExpression) ((IASTBinaryExpression) ((IASTExpressionStatement) bodyStmts[0])
+						.getExpression()).getOperand2();
+
+				IBasicType type = (IBasicType) expr.getExpressionType();
+				assertNotNull(type);
+				assertEquals(Kind.eInt, type.getKind());
+				assertTrue(type.isUnsigned());
+				assertFalse(type.isShort());
+				assertFalse(type.isLong());
+				assertTrue(type.isLongLong());
+			}
+		}
+	}
+
 	// char c;
 	// void func() {
 	//    c;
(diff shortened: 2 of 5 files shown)


core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/ArithmeticConversion.java
```diff
@@ -201,8 +201,8 @@ private final IType convert(IType type1, IType type2) {
 			return unsignedType;
 		}
 
-		// The signed has the higher rank.
-		if (signedRank.ordinal() > unsignedRank.ordinal()) {
+		// The signed has the higher rank, check if signed type can represent all unsigned values
+		if (fitsIntoType(signedType, unsignedType)) {
```
jonahgraham, line 205: Is there a test we can add to the testsuite to verify this?

i-garrison: There is an option to use one of `testTypePromotion_*` tests as starting point to probe the type of binary expression.
I commented on #265 that the test I'd like to have instead did not quite worked because binary expression evaluation did not performed as expected (e.g. 64bit case (1LL - 2UL) has ULL type but still evaluates less than zero). Debugging that confirms the type of binary expression value is now correct so `testTypePromotion_*`-style probe of expression type should succeed.
I also asked on the ticket if there is a real life test case for this which I believe would be preferred in test suite.

PhilippWendler:
Will this work in all cases? I don't know much about CDT internals, so parts of the following might be wrong, but I am understanding the solution as follows:
1. `fitsIntoType()` calls `SizeofCalculator.getSizeAndAlignment()`.
2. `getSizeAndAlignment()` calls `CPPSemantics.getCurrentLookupPoint()` in order to get the translation unit. When I am parsing C code, this returns `null`, and also the name indicates it should work only for C++. Is my understanding here correct? Or should it also work for C?
3. If it fails to get the translation unit, `getSizeAndAlignment()` uses the default `SizeofCalculator`, which just returns `null` for types like `int`, `long`, etc.
4. When `getSizeAndAlignment()` fails, `fitsIntoType()` uses `getApproximateSize()`. This returns 4 for `int`, 6 for `long`, and 8 for `long long`. The 6 is a hack for making `long` larger than `int` and smaller than `long long`.
5. This means that `fitsIntoType('signed long', 'unsigned int')` will return true.

However, a `signed long` is only large enough to represent all `unsigned int` values in a LP64 environment (e.g., 64-bit Linux). For ILP32 (32-bit x86, for example), this does not hold and thus the result would still be wrong.

AFAIU the solution here really needs to take the current parser configuration and type sizes into account, and so far this seems to work only for C++.

jonahgraham:
> AFAIU the solution here really needs to take the current parser configuration and type sizes into account

AFAIK one of the ways to do that is with [`SizeofCalculator`](https://github.com/eclipse-cdt/cdt/blob/main/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/SizeofCalculator.java)

PhilippWendler:
The current PR uses `SizeofCalculator`, but I think this is not good enough, at least I fail to see how it takes the parser configuration into account.

i-garrison:
> how it takes the parser configuration into account.

If translation unit is available, the `SizeofCalculator` derives everything from translation unit: https://github.com/eclipse-cdt/cdt/blob/main/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/SizeofCalculator.java#L110

i-garrison:
@PhilippWendler it would be nice if you could outline how do you use CDT parser. I can think of a way to provide it with access to necessary`__SIZEOF_*` macros without but details would depend on how people are using this outside of parsing the translation unit. Some unit tests are probably doing this already, by specifying custom scanner with extra macros pre-populated to tune lexer and parser behavior.

I agree that `getApproximateSize()` hack would sometimes yield surprising results, but would prefer to defer adding the whole lot of configuration UI for that to someone who can think this all out.

i-garrison:
I now added a test case which makes sure integer size macros are derived from STD scanner in the test.

I'm not sure whether this pattern of setting `CPPSemantics` lookup point before calling `getExpressionType` is a good idea in general, but here it helps testing changed code path. It might be a better idea to ask affected expression classes to push lookup point in their respective implementations before accessing `SizeofCalculator` or even better to explicitly fetch `SizeofCalculator` from translation unit without altering lookup point.

PhilippWendler:
Thanks a lot!

> @PhilippWendler it would be nice if you could outline how do you use CDT parser. I can think of a way to provide it with access to necessary`__SIZEOF_*` macros without translation unit in hand but details would depend on how people are using this outside of parsing the translation unit.

We are using it to parse a translation unit via `GCCLanguage.getDefault().getASTTranslationUnit(...)` and afterwards analyze the returned AST, I don't think we do anything special. And if we need to adjust and define some macros, this would be totally fine.

My question from https://github.com/eclipse-cdt/cdt/pull/269#issuecomment-1410069059 is: Is there really a translation unit at hand when parsing C code? The `SizeofCalculator` uses a class named `CPPSemantics` to access the translation unit, which makes me assume that it is only relevant for parsing C++ code. If this is nevertheless also guaranteed to be usable for C code, then it is fine for me.

The fact that the test seems to have the need to manually ensure that a translation unit is present also makes me assume that this would not work when parsing C code?

i-garrison:
> My question from [#269 (comment)](https://github.com/eclipse-cdt/cdt/pull/269#issuecomment-1410069059) is: Is there really a translation unit at hand when parsing C code? The `SizeofCalculator` uses a class named `CPPSemantics` to access the translation unit, which makes me assume that it is only relevant for parsing C++ code. If this is nevertheless also guaranteed to be usable for C code, then it is fine for me.

Yes, the translation unit is created by `parse()` to become the root of AST tree. Here is pure `C` use of `SizeofCalculator` by `CVisitor` to resolve integer type from `__attribute__((mode(DI)))` : 
https://github.com/eclipse-cdt/cdt/blob/main/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/c/CVisitor.java#L1311
Note in this case `CVisitor` gets `SizeofCalculator` using available tree node so no C++ lookup point hack is required.

> The fact that the test seems to have the need to manually ensure that a translation unit is present also makes me assume that this would not work when parsing C code?

Well I'm still exploring CDT internals )) As far as I can see the hack to pass translation unit via C++ lookup point mechanism in this test is only required because you cannot get the translation unit from `IType`, and current implementation has this problem because it gets the `SizeofCalculator` after it only has `IType` objects and no tree nodes. Once this is fixed we can drop setting lookup point from test code. Fixing this properly may need some time, help would be appreciated.

i-garrison:
> Does `SizeofVisitor` actually have access to it via `CPPSemantics` while parsing C code, or only while parsing C++ code?

Yes it does, example is in `CVisitor`



> The question is, would this be a problem for other uses? I imagine that inside CDT there is also code that analyzes the AST after it was parsed...

This is not an issue if `SizeofCalculator` can get hold of predefined macros which would provide it with type size info. This can be done after AST is parsed, like I did in this test.

At the moment I tend to believe that everything using `SizeofCalculator` needs to explicitly provide it with size info macros (via translation unit object) like `CVisitor` does. I'll try to experiment with that, and for now let this change be in draft state.

i-garrison:
Looking into code history, the change discussed in https://bugs.eclipse.org/bugs/show_bug.cgi?id=513105 introduced `CPPSemantics.pushLookupPoint()` mechanism to provide everything with access to point of instantiation and current translation unit. `SizeofCalculator` access to translation unit was changed there too.

Turned out that there are about five places which may eventually use `SizeofCalculator` in CDT which are not covered by `CPPSemantics.pushLookupPoint()` change: two are in C parser which I now fixed in this PR. This allowed to drop the hack from test so I believe it should work for external callers too.

Remaining 3 are less trivial but on the surface are used with C++ parser and should be covered as well. I did not looked into that, but added a comment at `ArithmeticConversions.fitsIntoType()` about CPPSemantics current lookup point - hope that will help with future bugs which may need the same change.

@PhilippWendler please have a look if this addresses some of your concerns about this fix.

core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/ArithmeticConversion.java
i-garrison: Added test case, both C and C++ now pass.

PhilippWendler:
Would certainly work for us, thanks a lot! Whether this covers everything in CDT I cannot claim to know.

jonahgraham (approved):

jonahgraham:
Thanks @i-garrison for the code change and @PhilippWendler for review and discussion.

It is great to see the community working well together.

PR #278, opened 2023-02-08
Revert "Add c++17 fold expression"

jonahgraham:
Reverts eclipse-cdt/cdt#234


jonahgraham:
Fixup, instead of revert, is in #279

PR #279, opened 2023-02-08
Fix rebase error for pr #234 - C++ Fold Expressions

jonahgraham:
PR #234 had a compilation error once it was rebased which wasn't identified until after it was rebased and merged to main causing build to fail.

This PR is to fix that rebase error.


jonahgraham:
I am merging this now to solve the build error on `main` branch. Please have a once over this change to make sure it is what you expect.

PR #283, opened 2023-02-11
c++17 fold expression - simplify parser and restrict to cast expressions

i-garrison:


jonahgraham (approved):
This looks fine to me, but I will give @ddscharfe a chance to weigh in as he's had some good insights on the earlier reviews on this topic.

ddscharfe:
- The parser accepts this, which I think should be an error: 
```C++
template<typename ... T>
void sum(T ... vals) {
	(vals + ... + vals);
	(vals + vals + vals);
}
```
- There are two branches of your validity checks (lines 1349 and 1353) which are not yet covered by the tests, maybe you could add tests for this:
image

i-garrison:
> * There are two branches of your validity checks (lines 1349 and 1353) which are not yet covered by the tests

Thanks! I now added invalid fold op test which should cover line 1349.

I cannot confirm coverage finding for line 1353 as `testFoldExpressionErrors()` hits that line at least for test case line 156 `(vals + ... + 1 * 2);` - tried adding an executable statement to that clause body and breakpoint is being hit for me (multiple times for that test case.)

i-garrison:
> * The parser accepts this, which I think should be an error:
> 
> ```c++
> 	(vals + ... + vals);
> ```
Correct, this is not a valid fold-expression. While this is special case and you can tell that it is wrong by just matching names on the left and right sides, I decided to not do that and instead just do that at evaluation time where check for unexpanded parameter pack is implemented. The check that any of names in fold expression are actually parameter packs is not done too for same reason.

> ```c++
> 	(vals + vals + vals);
> ```

This is actually not a fold expression because there is no `...` anywhere. As far as  I know CDT parser cannot decide if name is a [valid] unexpanded parameter pack at parse time, so it is not possible to tell that about this otherwise valid bracketed expression at parse time.

Probably both of these cases can be implemented in codan scanner.

ddscharfe:
> I cannot confirm coverage finding for line 1353 as testFoldExpressionErrors() hits that line at least for test case line 156...

The code coverage checks all branches, so in this case it would only become green if all four possible branches of this expression are covered by tests. I think what's missing is a test case where `rightChain != null` and `expr == null`.

> ...
> Probably both of these cases can be implemented in codan scanner.

Thanks for the explanation, that makes sense and I agree that the codan scanner would be the right place for these further checks.

i-garrison:
> The code coverage checks all branches, so in this case it would only become green if all four possible branches of this expression are covered by tests. I think what's missing is a test case where `rightChain != null` and `expr == null`.

Thanks again, I overlooked that if `rightChain != null` then `expr != null` is always true (because having a chain in BinaryOperator requires having non-null `expr` too.) - dropped the excessive check.

ddscharfe (approved):
LGTM, thanks @i-garrison for providing this PR.

jonahgraham:
Thank you @i-garrison for the code and @ddscharfe for taking time to provide such a thorough review.

PR #284, opened 2023-02-12
c++ apply integer conversions add tests

i-garrison:
This change fixes a few places where integer conversion was missing, and adds a few tests using corrected evaluation results.

I do use these primarily for constexpr evaluations in CDT UI so there are probably more cases where integer conversions are missing (e.g. I did not looked at C code at all.)


jonahgraham (approved):
Thanks for this move in the right direction.

> [...] where integer conversions are missing (e.g. I did not looked at C code at all.)

It is ok by me that these other cases aren't done. I am glad you are able to move your use case forward.

PS I am quite curious about your use case as this code hasn't gotten any love for a while it is interesting to me to know who and why someone is investing time in it.

i-garrison:
Thanks @jonahgraham Well there is nothing major behind this at the moment - I'm just using CDT to explore various code bases (both C and C++) for ages now, some of those projects recently moved to C++20, cdt.cloud is not ready yet and I cannot find anything about ongoing LSP integration efforts - looks like that stalled. So I decided to add a few missing features to CDT which would in short term unblock browsing c++20 code to some extent, and part of that is correct constexpr evaluations.

LSP integration looks more future-proof but I do not see anyone doing that for CDT at the moment.

Issue #194, opened 2022-12-05
org.eclipse.cdt.ui.tests.text.contentassist2.PlainC_CompletionTests


jonahgraham:
Even with the fixes in #117 I am still seeing tests in org.eclipse.cdt.ui.tests.text.contentassist2.PlainC_CompletionTests fail too often.

PR #285, opened 2023-02-14
Fix codan return value check with typedef return and template function

i-garrison:


i-garrison:
I cannot reproduce this UI `testCasts2` test error locally, could it be flaky too?

jonahgraham:
> I cannot reproduce this UI `testCasts2` test error locally, could it be flaky too?

- Yes covered in #194 but I hadn't seen it fail for a while recently.

jonahgraham (approved):

PR #286, opened 2023-02-15
Option to override the matching of macro defines for the Core Build GCC

ewaterlander:
toolchain. This may be needed for custom compilers.


jonahgraham:
@ewaterlander the idea of this looks fine to me, but as this adds new API I think it is important to have the API be a bit cleaner and avoid passing mutable objects into the method.

Rather than try to explain what I think is needed I tried to draft something up. Please look at this change and let me know what you think. Does this work for you?

If you needed the mutable `symbols` passed to the subclass, then my change breaks it. But if you need that, can you please explain why?

Issue #287, opened 2023-02-15
Null check missing for input argument (source) when updating Disassembly view


LizzMre:
**Repo:** https://github.com/eclipse-cdt/cdt/tree/main/dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly

**Bug Description**

***Scenario***
Let’s consider that we have a Client-Server architecture where the server is sending to the client information on which lines to append to the Disassembly view from a specific file. 
There might be a possibility for the server to send some incorrect data, for example: sending additional lines that cannot be found in the associated file. 
In this case, the GUI will display a NullPointerException when trying to fetch the lines that don’t exist.
Even though the server might send incorrect data, it would be nice to provide the user a message with a suggestive description of the issue and prevent the GUI from crashing. 

***Code flow***
1. Information was received and it will reach the following line:
[SourceColorerJob.runInUIThread() -> line 63](https://github.com/eclipse-cdt/cdt/blob/58ab28dfd120c5c6fd320c51949d5cf0d245ff5b/dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly/SourceColorerJob.java#L63)
<img width="469" alt="image" src="https://user-images.githubusercontent.com/103601453/219044844-2a72cd4b-86da-4ca1-be73-78668cacc41d.png">

2.  Next, the following method will be called:
[DisassemblyPart.updateInvalidSource() -> line 1706](https://github.com/eclipse-cdt/cdt/blob/58ab28dfd120c5c6fd320c51949d5cf0d245ff5b/dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly/DisassemblyPart.java#L1706)
![image](https://user-images.githubusercontent.com/103601453/219044498-6106899f-199e-4d97-8103-2e9bbc3dbf7f.png)

3. From here, for each line (that must be fetched from the file) it will reach the following method: [DisassemblyPart.insertSource() -> line 2756](https://github.com/eclipse-cdt/cdt/blob/58ab28dfd120c5c6fd320c51949d5cf0d245ff5b/dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly/DisassemblyPart.java#L2756) that has multiple blocks of code similar to the sample below:

***Sample from this method***
[...]

PR #295, opened 2023-02-22
Add null check for Disassembly view source

LizzMre:
Disassembly view is expected to be populated with lines that are fetched from a given source file. There might be the case where instructions on what lines to append are wrong. This results in a null response that will propagate through the code leading to a NPE.

The current commit is proofing the code from NPE by:
- removing the source position of the lines that were not found within the given file
- null checking the source before becoming a key element in the code flow
- adding logging if expected lines are not found in the given file

Resolves: #287

### DIFF
--- a/dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly/model/SourceFileInfo.java
+++ b/dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly/model/SourceFileInfo.java
@@ -25,6 +25,7 @@ import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IStorage;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Status;
 import org.eclipse.core.runtime.content.IContentType;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.jface.text.BadLocationException;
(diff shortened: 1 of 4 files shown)


dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly/model/SourceFileInfo.java
```diff
@@ -147,6 +147,10 @@ public String getLines(int first, int last) {
 			}
 			return fSource.get(startOffset, endOffset - startOffset);
 		} catch (BadLocationException e) {
+			// Log error to indicate what is tha cause of the issue
+			String warningMessage = "Line(s) " + Integer.toString(first) + "-" + Integer.toString(last) //$NON-NLS-1$
+					+ " cannot be found in file: " + fFileKey + ".\nSkipping lines!"; //$NON-NLS-1$
+			DsfUIPlugin.logErrorMessage(warningMessage); //$NON-NLS-1$
```
jonahgraham, line 153: This logErrorMessage has two minor problems:

1. It discards the exception. I think you should log the exception too, it provides the stack trace/context to diagnose the problem
2. What is logged in this case is `Internal message logged from Debug UI` as type `INTERAL_ERROR`, which seems a little to harsh here.

```suggestion
			DsfUIPlugin.log(Status.warning(warningMessage, e));
```

jonahgraham (changes requested):
This looks good to me. May need some cosmetic and version changes to get it through the build, but the code change itself is done apart from the minor logging issue.

jonahgraham:
GitHub doesn't allow me to add a comment to a non-modified file. The build failed with this error:

```
The following bundles are missing a service segment version bump:
  - org.eclipse.cdt.dsf.ui
Please bump service segment by 100 if on master branch
```

which means this line https://github.com/eclipse-cdt/cdt/blob/ab721eed87089d61e4ce4724646a1b1d3a79a484/dsf/org.eclipse.cdt.dsf.ui/META-INF/MANIFEST.MF#L6

needs to be changed to

```
Bundle-Version: 2.7.100.qualifier
```

dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly/model/SourceFileInfo.java
LizzMre: I updated this and the Bundle-Version. Thank you!

jonahgraham (approved):

jonahgraham:
Thanks @LizzMre - this change looks good now. I am waiting on CI to confirm no issues and then it can be merged.

LizzMre:
@jonahgraham Thank you very much for your help!

PR #311, opened 2023-03-07
Improved getting resources from the gcc commandline.

ewaterlander:
The old method assumed all resources were at the end of the line, optionally followed by option -o.
Now the complete line is scanned for valid resource file extensions.

### DIFF
--- a/build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/GCCToolChain.java
+++ b/build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/GCCToolChain.java
@@ -20,10 +20,13 @@ import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
+import java.util.TreeSet;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -46,6 +49,8 @@ import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.PlatformObject;
+import org.eclipse.core.runtime.content.IContentType;
+import org.eclipse.core.runtime.content.IContentTypeManager;
 
 /**
  * The GCC toolchain. This is the base class for all GCC toolchains. It
@@ -68,6 +73,7 @@ public class GCCToolChain extends PlatformObject implements IToolChain {
 	private String cCommand;
 	private String cppCommand;
 	private String[] commands;
+	private Set<String> resourcesFileExtensions = getResourcesFileExtensions();
 
 	@Deprecated
 	public GCCToolChain(IToolChainProvider provider, String id, String version) {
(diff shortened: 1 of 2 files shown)


build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/GCCToolChain.java
```diff
@@ -640,42 +645,102 @@ public String[] getCompileCommands(ILanguage language) {
...
+		}
+		return ext;
+	}
+
 	@Override
 	public IResource[] getResourcesFromCommand(List<String> cmd, URI buildDirectoryURI) {
 		// Start at the back looking for arguments
 		List<IResource> resources = new ArrayList<>();
 		IWorkspaceRoot root = ResourcesPlugin.getWorkspace().getRoot();
+		String extensionPattern = getPatternFileExtensions();
```
jonahgraham, line 707: Below I recommend getting rid of the regular expression entirely, but if it is needed: Because this code can be processing large amounts of output we should be pre-compiling, or compiling as early as possible, the patterns.

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/GCCToolChain.java
```diff
@@ -640,42 +645,102 @@ public String[] getCompileCommands(ILanguage language) {
...
-				URI uri;
-				if (srcPath.isAbsolute()) {
-					uri = srcPath.toUri();
-				} else {
-					String mingwPath = fixMingwPath(arg);
-					if (mingwPath != arg) {
-						uri = Paths.get(mingwPath).toUri();
+
+			String ext = getFileExtension(arg);
+			if (Pattern.matches(extensionPattern, ext)) {
```
jonahgraham, line 722: I don't think a regular expression is the suitable data structure here. How about something like this:

```java

	private Set<String> getPatternFileExtensions() {
		IContentTypeManager manager = Platform.getContentTypeManager();

		Set<String> fileExts = new TreeSet<String>(String.CASE_INSENSITIVE_ORDER);

		IContentType contentTypeCpp = manager.getContentType(CCorePlugin.CONTENT_TYPE_CXXSOURCE);
		fileExts.addAll(Arrays.asList(contentTypeCpp.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));

		IContentType contentTypeC = manager.getContentType(CCorePlugin.CONTENT_TYPE_CSOURCE);
		fileExts.addAll(Arrays.asList(contentTypeC.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));

		return fileExts;
	}

```

and then here do:

```java
    if (getPatternFileExtensions().contains(ext)) {
```

(although it may be best to cache the getPatternFileExtensions result somewhere as the number of arguments and number of command lines to be processed could be quite large)

jonahgraham (changes requested):
Your change looks good. It really makes  me wonder why originally it was coded to only consider the last arguments to the command line though. I hope there is nothing we are overlooking on this and really wish that there were some preexisting tests for this. It seems unlikely there would be an errant `.c` or similar  in the command line so I am ok with this change.

The code you copied needed to use regular expressions because it was integrated into a longer regex. As you only care about extensions here I recommend a simpler bit of code for your consideration.

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/GCCToolChain.java
ewaterlander: Hi Jonah,
Thanks for your comment. Regular expression is indeed overkill here. I will update the change.
kind regards,
Erwin

jonahgraham (approved):
LGTM. I will be merging soon

jonahgraham:

jonahgraham:
Thank you @ewaterlander for the fix.

ewaterlander:
> Thank you @ewaterlander for the fix.

You're welcome.

PR #312, opened 2023-03-08
Defer resolving function set if target is deferred class or constructor

i-garrison:


jonahgraham:

jonahgraham (approved):

PR #313, opened 2023-03-08
Allow const-qualified arg for __atomic builtins which accept that

i-garrison:


jonahgraham:
Please ignore the failing ECA - there is a network problem at the foundation, see https://www.eclipsestatus.io/ for more details.

PR #315, opened 2023-03-11
Fix assertion evaluating binary shifts of comparison op results

i-garrison:
Currently CDT evaluates result type of comparison ops to the converted type of their operands. When floating point values are compared and then the result is binary shifted this trips an assertion trying to promote floating point value.

Fix this by limiting operand types of binary shifts to integral or unscoped enumerations. Additionally fix return type of comparison op to be boolean value.

Issue found looking at Qt code, similar to this https://github.com/qt/qt/blob/4.8/src/gui/painting/qpainterpath.cpp#L1824-L1834


jonahgraham:

jonahgraham (approved):

PR #317, opened 2023-03-12
Restore counting iterations in computeContainsDeclaration cycle detector

i-garrison:
A follow-up change to [bug 509898](https://bugs.eclipse.org/bugs/show_bug.cgi?id=509898) missed actually incrementing the counter, restore it now.

Fixes: 5462bac381 ("Bug 509898 - IndexFileSet.containsDeclaration is slow and is causing UI freezes")


jonahgraham (approved):
Good catch on this one. Did you encounter a case where this is making a difference, or did you hit this by code inspection?

jonahgraham:

PR #318, opened 2023-03-13
Return IntegralValue.UNKNOWN if recursed creating CompositeValue

i-garrison:
Currently CDT would return empty `CompositeValue` if recursion creating for class type is detected. If this happens inside `CompositeValue.computeForFunctionCall()` and class type has fields, this leads to an attempt to assign a value to non-existent index into empty values array.

Fix this by returning usual `IntegralValue.UNKNOWN` and additionally checking whether created value is actually a `CompositeValue` instance.

### DIFF
--- a/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/cpp/semantics/EvalConstructor.java
+++ b/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/cpp/semantics/EvalConstructor.java
@@ -160,10 +160,14 @@ public final class EvalConstructor extends CPPDependentEvaluation {
 			return this;
 		}
 		final ICPPClassType classType = (ICPPClassType) unwrappedType;
-		final CompositeValue compositeValue = CompositeValue.create(classType);
+		final IValue classObject = CompositeValue.create(classType);
 		ICPPEvaluation[] argList = evaluateArguments(fArguments, callSiteRecord, context);
-		EvalFixed constructedObject = new EvalFixed(fType, ValueCategory.PRVALUE, compositeValue);
+		EvalFixed constructedObject = new EvalFixed(fType, ValueCategory.PRVALUE, classObject);
 		CPPVariable binding = new CPPVariable(TEMP_NAME);
+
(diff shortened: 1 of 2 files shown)


i-garrison:
Hi @jonahgraham no worries, keen to see the release!

core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/cpp/semantics/EvalConstructor.java
```diff
@@ -160,11 +160,15 @@ public ICPPEvaluation computeForFunctionCall(ActivationRecord callSiteRecord, Co
...
 		}
 		final ICPPClassType classType = (ICPPClassType) unwrappedType;
-		final CompositeValue compositeValue = CompositeValue.create(classType);
+		final IValue classObject = CompositeValue.create(classType);
 		ICPPEvaluation[] argList = evaluateArguments(fArguments, callSiteRecord, context);
-		EvalFixed constructedObject = new EvalFixed(fType, ValueCategory.PRVALUE, compositeValue);
+		EvalFixed constructedObject = new EvalFixed(fType, ValueCategory.PRVALUE, classObject);
 		CPPVariable binding = new CPPVariable(TEMP_NAME);
 
+		if (!(classObject instanceof CompositeValue compositeValue)) {
```
jonahgraham, line 168: This variable is unused

```suggestion
		if (!(classObject instanceof CompositeValue)) {
```

jonahgraham:

jonahgraham (approved):

core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/dom/parser/cpp/semantics/EvalConstructor.java
jonahgraham: Sorry, I didn't know that the code could do what it did. That variable was indeed used, fix coming in a moment.

jonahgraham:
Thank you @i-garrison for this set of PRs that will give all users a better experience. I think I have now merged all your outstanding changes.

PR #328, opened 2023-03-17
Binary not found in Core Build Makefile project

ewaterlander:
Fixed the problem for Core Build Makefile projects that the output binary could not be found during launching when the user set the Build Output Location to "Build in project directory". The project was only looking for binaries in the "build" folder.

### DIFF
--- a/build/org.eclipse.cdt.make.core/src/org/eclipse/cdt/make/core/MakefileProjectGenerator.java
+++ b/build/org.eclipse.cdt.make.core/src/org/eclipse/cdt/make/core/MakefileProjectGenerator.java
@@ -84,7 +84,7 @@ public class MakefileProjectGenerator extends FMProjectGenerator {
 			}
 		}
 
-		entries.add(CoreModel.newOutputEntry(buildFolder.getFullPath()));
+		entries.add(CoreModel.newOutputEntry(project.getFullPath()));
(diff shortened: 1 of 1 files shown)


jonahgraham (approved):
LGTM. I think this is a workaround rather than the full solution as presumably the "correct" solution would be the output entry would update based on the setting "Build in project directory" or not. As launching seems to work as expected this is ready to go.

jonahgraham:
This is ready to go, but I don't understand why it consistently fails the same test so I have restarted it again. I think it is just a flaky test, but I want to make sure there is no unintended consequence here.

jonahgraham:
Thanks @ewaterlander

ewaterlander:
> LGTM. I think this is a workaround rather than the full solution as presumably the "correct" solution would be the output entry would update based on the setting "Build in project directory" or not. As launching seems to work as expected this is ready to go.

Currently a Core Build Makefile project will just launch the first binary found in the binaries container. The real output location also depends on the launch mode set in the LaunchBar. If you set the location to "configuration specific directory" and switch from Run to Debug without cleaning you will not see debug info, because the first binary under "build" is found in the build subdirectory "default".

The project doesn't control the content of the Makefile. In theory the binary could be written anywhere, even with location set to "configuration specific directory". The user needs to be aware which binary is built. It is currently not foolproof.

PR #331, opened 2023-03-19
Handle null "DisplayName" in registry

### LINKED ISSUE #330: mingw detector causes NPE
### Discussed in https://github.com/eclipse-cdt/cdt/discussions/329

mingw detector code generates an NPE. This appears to be a regression introduced by #243

`displayName` in this bit of code can be `null` leading to an NPE because `MSYS2_64BIT_NAMES` doesn't allow `null` values so you can't even to `contains(null)` on it.

https://github.com/eclipse-cdt/cdt/blob/2b297c4355a07ec26b68c88cc1d8d0bbcab87186/core/org.eclipse.cdt.core/utils/org/eclipse/cdt/internal/core/MinGW.java#L97-L99

The same issue is here:

https://github.com/eclipse-cdt/cdt/blob/a949830dd36a84297cac39452f725dad9ef9875c/build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/internal/Msys2ToolChainProvider.java#L50

I think the fix is simply adding a null check (or using a set that allows `null`).

cc @jld01

jonahgraham:
Fixes #330


jonahgraham:
@jld01 please let me know what you think, and if there is anywhere else this may need to be fixed too. Should we spin up a 11.1.1 release to include this?

jld01 (approved):
Looks good to me, @jonahgraham.

jonahgraham:
Should we spin up a 11.1.1 release to include this?

jld01:
> Should we spin up a 11.1.1 release to include this?

The workaround is undesirable and error-prone, so a bugfix release seems highly desirable to me.

jonahgraham:
+1

PR #336, opened 2023-03-22
Bug 528145 - Breakpoints are not working with remote attach launch

umairsair:
Looking at the logs, it seems that the regression is caused at 8bec791
where support for multi-process was added. We removed breakpoints
tracking support from final launch sequence and moved it to debug new
process and attach to process logic but none of these are run for remote
attach launch, hence breakpoint tracking is not started for remote
attach launch.

To fix the problem, IGDBProcesses.attachDebuggerToProcess(..) is updated
to handle remote attach launch as well instead of final launch sequence
handling it.

This commit is created after reverting 7bddb5f and 96839a0 which is the
older fix done to fix this issue and the other commit was to fix the
regression caused by the old fix.

The problem with older fix was that for non-stop mode, attach to process
was not working for remote launches when there is already a process
being debugged. Note that to use this feature, gdbserver should be
started with --multi option.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -37,6 +37,7 @@ jobs:
       run: |
         export DISPLAY=:99
         sudo Xvfb -ac :99 -screen 0 1280x1024x24 > /dev/null 2>&1 &
+        echo 0| sudo tee /proc/sys/kernel/yama/ptrace_scope
         mvn \
           clean verify -B -V \
           -Dmaven.test.failure.ignore=true \
--- a/NewAndNoteworthy/CHANGELOG-API.md
+++ b/NewAndNoteworthy/CHANGELOG-API.md
@@ -599,7 +599,7 @@ spelled BuiltinDetectionArgsGeneric instead.
 
 ## API Removals after June 2025
 
-### <span id="FinalLaunchSequence">FinalLaunchSequence.stepRemoteConnection() and FinalLaunchSequence.stepAttachRemoteToDebugger() will be removed</span>
+### FinalLaunchSequence.stepRemoteConnection() and FinalLaunchSequence.stepAttachRemoteToDebugger() will be removed
 
 These APIs will be removed and remote connection for attach launch will be moved in the implementation of `IGDBProcesses.attachDebuggerToProcess()`.
 
--- a/NewAndNoteworthy/CHANGELOG-API.md
+++ b/NewAndNoteworthy/CHANGELOG-API.md
@@ -596,3 +596,11 @@ The class BuiltinDetctionArgsGeneric will be removed. Use the correctly
 spelled BuiltinDetectionArgsGeneric instead.
 
 - org.eclipse.cdt.jsoncdb.core.participant.Arglets.BuiltinDetctionArgsGeneric
+
+## API Removals after June 2025
+
+### <span id="FinalLaunchSequence">FinalLaunchSequence.stepRemoteConnection() and FinalLaunchSequence.stepAttachRemoteToDebugger() will be removed</span>
+
+These APIs will be removed and remote connection for attach launch will be moved in the implementation of `IGDBProcesses.attachDebuggerToProcess()`.
+
+See https://github.com/eclipse-cdt/cdt/pull/336
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.cdt.dsf.gdb;singleton:=true
-Bundle-Version: 7.0.0.qualifier
+Bundle-Version: 7.1.0.qualifier
 Bundle-Activator: org.eclipse.cdt.dsf.gdb.internal.GdbPlugin
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime,
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
@@ -19,7 +19,6 @@
  *     Anton Gorenkov - A preference to use RTTI for variable types determination (Bug 377536)
  *     Xavier Raynaud (Kalray) - Avoid duplicating fields in sub-classes (add protected accessors)
  *     Marc Khouzam (Ericsson) - Output the version of GDB at startup (Bug 455408)
- *     Jonathan Tousignant (NordiaSoft) - Remote session breakpoint (Bug 528145)
  *******************************************************************************/
 package org.eclipse.cdt.dsf.gdb.launching;
 
@@ -39,7 +38,6 @@ import org.eclipse.cdt.dsf.concurrent.RequestMonitor;
 import org.eclipse.cdt.dsf.concurrent.RequestMonitorWithProgress;
 import org.eclipse.cdt.dsf.datamodel.DataModelInitializedEvent;
 import org.eclipse.cdt.dsf.datamodel.IDMContext;
-import org.eclipse.cdt.dsf.debug.service.IProcesses.IProcessDMContext;
 import org.eclipse.cdt.dsf.debug.service.ISourceLookup.ISourceLookupDMContext;
 import org.eclipse.cdt.dsf.debug.service.command.ICommandControlService.ICommandControlDMContext;
 import org.eclipse.cdt.dsf.gdb.IGDBLaunchConfigurationConstants;
@@ -52,7 +50,6 @@ import org.eclipse.cdt.dsf.gdb.service.SessionType;
 import org.eclipse.cdt.dsf.gdb.service.command.IGDBControl;
 import org.eclipse.cdt.dsf.mi.service.CSourceLookup;
 import org.eclipse.cdt.dsf.mi.service.IMIProcesses;
-import org.eclipse.cdt.dsf.mi.service.MIProcesses;
 import org.eclipse.cdt.dsf.mi.service.command.CommandFactory;
 import org.eclipse.cdt.dsf.mi.service.command.output.MIGDBVersionInfo;
 import org.eclipse.cdt.dsf.mi.service.command.output.MIInfo;
@@ -142,8 +139,6 @@ public class FinalLaunchSequence extends ReflectionSequence {
 					"stepNewProcess", //$NON-NLS-1$
 					// For local attach launch only
 					"stepAttachToProcess", //$NON-NLS-1$
-					// For remote attach launch only
-					"stepAttachRemoteToDebugger", //$NON-NLS-1$
 					// Global
 					"stepDataModelInitializationComplete", //$NON-NLS-1$
 					"stepCleanup", //$NON-NLS-1$
@@ -647,23 +642,6 @@ public class FinalLaunchSequence extends ReflectionSequence {
 					requestMonitor.done();
 				}
 			}
-		} else {
-			requestMonitor.done();
-		}
-	}
-
-	/**
-	 * If we are dealing with an remote attach session, perform the attach to debugger.
-	 * Bug 528145
-	 * @since 6.6
-	 */
-	@Execute
-	public void stepAttachRemoteToDebugger(final RequestMonitor requestMonitor) {
-		if (fGDBBackend.getIsAttachSession() && fGDBBackend.getSessionType() == SessionType.REMOTE) {
-			IProcessDMContext processContext = fProcService.createProcessContext(fCommandControl.getContext(),
-					MIProcesses.UNKNOWN_PROCESS_ID);
-			fProcService.attachDebuggerToProcess(processContext,
-					new DataRequestMonitor<IDMContext>(getExecutor(), requestMonitor));
 		} else {
 			requestMonitor.done();
 		}
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
@@ -660,18 +660,13 @@ public class FinalLaunchSequence extends ReflectionSequence {
 	@Execute
 	public void stepAttachRemoteToDebugger(final RequestMonitor requestMonitor) {
 		if (fGDBBackend.getIsAttachSession() && fGDBBackend.getSessionType() == SessionType.REMOTE) {
-			DataRequestMonitor<Boolean> rm = new DataRequestMonitor<>(getExecutor(), null);
-			fProcService.canDetachDebuggerFromProcess(null, rm);
-
-			if (rm.getData()) {
-				IProcessDMContext processContext = fProcService.createProcessContext(fCommandControl.getContext(),
-						MIProcesses.UNKNOWN_PROCESS_ID);
-				fProcService.attachDebuggerToProcess(processContext,
-						new DataRequestMonitor<IDMContext>(getExecutor(), requestMonitor));
-				return;
-			}
+			IProcessDMContext processContext = fProcService.createProcessContext(fCommandControl.getContext(),
+					MIProcesses.UNKNOWN_PROCESS_ID);
+			fProcService.attachDebuggerToProcess(processContext,
+					new DataRequestMonitor<IDMContext>(getExecutor(), requestMonitor));
+		} else {
+			requestMonitor.done();
 		}
-		requestMonitor.done();
 	}
 
 	/**
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
@@ -45,11 +45,12 @@ import org.eclipse.cdt.dsf.gdb.IGdbDebugPreferenceConstants;
 import org.eclipse.cdt.dsf.gdb.actions.IConnect;
 import org.eclipse.cdt.dsf.gdb.internal.GdbPlugin;
 import org.eclipse.cdt.dsf.gdb.service.IGDBBackend;
+import org.eclipse.cdt.dsf.gdb.service.IGDBProcesses;
 import org.eclipse.cdt.dsf.gdb.service.IGDBSourceLookup;
 import org.eclipse.cdt.dsf.gdb.service.SessionType;
 import org.eclipse.cdt.dsf.gdb.service.command.IGDBControl;
 import org.eclipse.cdt.dsf.mi.service.CSourceLookup;
-import org.eclipse.cdt.dsf.mi.service.IMIProcesses;
+import org.eclipse.cdt.dsf.mi.service.MIProcesses;
 import org.eclipse.cdt.dsf.mi.service.command.CommandFactory;
 import org.eclipse.cdt.dsf.mi.service.command.output.MIGDBVersionInfo;
 import org.eclipse.cdt.dsf.mi.service.command.output.MIInfo;
@@ -74,7 +75,7 @@ public class FinalLaunchSequence extends ReflectionSequence {
 
 	private IGDBControl fCommandControl;
 	private IGDBBackend fGDBBackend;
-	private IMIProcesses fProcService;
+	private IGDBProcesses fProcService;
 	private CommandFactory fCommandFactory;
 
 	private DsfServicesTracker fTracker;
@@ -133,11 +134,11 @@ public class FinalLaunchSequence extends ReflectionSequence {
 					//
 					// "stepSetSourceLookupPath",   //$NON-NLS-1$
 
-					// For remote-attach launch only
+					// For remote-attach launch only (deprecated)
 					"stepRemoteConnection", //$NON-NLS-1$
 					// For all launches except attach ones
 					"stepNewProcess", //$NON-NLS-1$
-					// For local attach launch only
+					// For all attach launch only
 					"stepAttachToProcess", //$NON-NLS-1$
 					// Global
 					"stepDataModelInitializationComplete", //$NON-NLS-1$
@@ -173,7 +174,7 @@ public class FinalLaunchSequence extends ReflectionSequence {
 
 		fCommandFactory = fCommandControl.getCommandFactory();
 
-		fProcService = fTracker.getService(IMIProcesses.class);
+		fProcService = fTracker.getService(IGDBProcesses.class);
 		if (fProcService == null) {
 			requestMonitor.setStatus(
 					new Status(IStatus.ERROR, GdbPlugin.PLUGIN_ID, -1, "Cannot obtain process service", null)); //$NON-NLS-1$
@@ -547,13 +548,17 @@ public class FinalLaunchSequence extends ReflectionSequence {
 		rm.done();
 	}
 
+	@Deprecated(forRemoval = true)
 	private static final String INVALID = "invalid"; //$NON-NLS-1$
 
 	/**
 	 * If we are dealing with a remote-attach debugging session, connect to the target.
 	 * @since 4.0
+	 *
+	 * When removing, revive/uncomment code in implementations of IGDBProcesses.attachDebuggerToProcess()
 	 */
 	@Execute
+	@Deprecated(forRemoval = true)
 	public void stepRemoteConnection(final RequestMonitor rm) {
 		if (fGDBBackend.getSessionType() == SessionType.REMOTE && fGDBBackend.getIsAttachSession()) {
 			boolean isTcpConnection = CDebugUtils.getAttribute(fAttributes,
@@ -588,19 +593,9 @@ public class FinalLaunchSequence extends ReflectionSequence {
 	@Execute
 	public void stepNewProcess(final RequestMonitor rm) {
 		if (!fGDBBackend.getIsAttachSession()) {
-			boolean noBinarySpecified = CDebugUtils.getAttribute(fAttributes,
-					IGDBLaunchConfigurationConstants.ATTR_DEBUGGER_USE_SOLIB_SYMBOLS_FOR_APP,
-					IGDBLaunchConfigurationConstants.DEBUGGER_USE_SOLIB_SYMBOLS_FOR_APP_DEFAULT);
-
-			String binary = null;
-			final IPath execPath = fGDBBackend.getProgramPath();
-			if (!noBinarySpecified && execPath != null && !execPath.isEmpty()) {
-				binary = execPath.toString();
-			}
-
 			// Even if binary is null, we must call this to do all the other steps
 			// necessary to create a process.  It is possible that the binary is not needed
-			fProcService.debugNewProcess(fCommandControl.getContext(), binary, fAttributes,
+			fProcService.debugNewProcess(fCommandControl.getContext(), getBinary(), fAttributes,
 					new DataRequestMonitor<IDMContext>(getExecutor(), rm) {
 						@Override
 						protected void handleCancel() {
@@ -618,14 +613,28 @@ public class FinalLaunchSequence extends ReflectionSequence {
 	}
 
 	/**
-	 * If we are dealing with an local attach session, perform the attach.
-	 * For a remote attach session, we don't attach during the launch; instead
-	 * we wait for the user to manually do the attach.
+	 * @since 7.1
+	 */
+	protected String getBinary() {
+		boolean noBinarySpecified = CDebugUtils.getAttribute(fAttributes,
+				IGDBLaunchConfigurationConstants.ATTR_DEBUGGER_USE_SOLIB_SYMBOLS_FOR_APP,
+				IGDBLaunchConfigurationConstants.DEBUGGER_USE_SOLIB_SYMBOLS_FOR_APP_DEFAULT);
+
+		String binary = null;
+		final IPath execPath = fGDBBackend.getProgramPath();
+		if (!noBinarySpecified && execPath != null && !execPath.isEmpty()) {
+			binary = execPath.toString();
+		}
+		return binary;
+	}
+
+	/**
+	 * If we are dealing with an attach session, perform the attach.
 	 * @since 4.0
 	 */
 	@Execute
 	public void stepAttachToProcess(final RequestMonitor requestMonitor) {
-		if (fGDBBackend.getIsAttachSession() && fGDBBackend.getSessionType() != SessionType.REMOTE) {
+		if (fGDBBackend.getIsAttachSession()) {
 			// Is the process id already stored in the launch?
 			int pid = CDebugUtils.getAttribute(fAttributes, ICDTLaunchConfigurationConstants.ATTR_ATTACH_PROCESS_ID,
 					-1);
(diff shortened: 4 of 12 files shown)


jonahgraham:
I have added me and a couple of others to review this change. If neither of the other two can look soon I will schedule a deeper dive on this next week.

jld01 (commented):
These changes look reasonable to me. It appears that the `GDBJtagDSFFinalLaunchSequence` will not be affected but perhaps @umairsair could confirm that he has considered this case.

umairsair:
> These changes look reasonable to me. It appears that the `GDBJtagDSFFinalLaunchSequence` will not be affected but perhaps @umairsair could confirm that he has considered this case.

Yes, it does not impact.. I tried "GDB Hardware Debugging" launch and it seems to work fine.. it is not attach launch, so it is not impacted..

jonahgraham:
> @jonahgraham a kind reminder..

Thanks - sorry it hasn't made it to the top of my list.

@jld01 if you feel your review is sufficient, please go ahead and merge it.

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
```diff
@@ -552,39 +546,6 @@ protected void handleError() {
...
 	}
 
-	private static final String INVALID = "invalid"; //$NON-NLS-1$
-
-	/**
-	 * If we are dealing with a remote-attach debugging session, connect to the target.
-	 * @since 4.0
-	 */
-	@Execute
-	public void stepRemoteConnection(final RequestMonitor rm) {
```
jonahgraham, line 562: This is now API and can't be removed this simply.

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
```diff
@@ -652,28 +621,6 @@ public void stepAttachToProcess(final RequestMonitor requestMonitor) {
 		}
 	}
 
-	/**
-	 * If we are dealing with an remote attach session, perform the attach to debugger.
-	 * Bug 528145
-	 * @since 6.6
-	 */
-	@Execute
-	public void stepAttachRemoteToDebugger(final RequestMonitor requestMonitor) {
```
jonahgraham, line 661: This is now API and can't be removed this simply.

jonahgraham (commented):
I don't think I understand the scope of the problem. Please can you provide what the steps are to reproduce this problem?

I started `gdbserver --multi :1234` and made a `C/C++ Attach to Application` with Debugger as `gdbserver` in Debugger tab, and set to non-stop. I am able to `Debug New Executable` multiple times.

I am using gdb/gdbserver 10.2 on linux.

jonahgraham:
> I am able to Debug New Executable multiple times.

To clarify, I was able to debug new exec multiple times without this patch.

umairsair:
> > I am able to Debug New Executable multiple times.
> 
> To clarify, I was able to debug new exec multiple times without this patch.

The problem is with connecting to existing process.
1. Start a process with debug new exec.
2. Now select connect and ask debugger to connect to any process, it'll connect.

Description in first comment also mentions the same..

> The problem with older fix was that for non-stop mode, attach to process
was not working for remote launches when there is already a process
being debugged.

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
umairsair: yup! what should be the way forward?

umairsair:
corrected last comment..

jonahgraham:
Thanks for step-by-step. I can reproduce it and see it is fixed. Now back to the review.

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
jonahgraham: Now that I have reproduced I will start trying to figure out what workarounds we need to do about this.

API compatibility errors is just about binary compatibility, sometimes we make semantic changes that doesn't break API, but is essentially breaking change in behaviour.

Anyway, I'll try to provide a concrete suggestion.

jonahgraham (commented):
Approved - subject to fixing up the API problems, which I think can be done like this:

```diff
$ git diff
diff --git a/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java b/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
index 67ee548755..19b7786043 100644
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
@@ -134,6 +134,7 @@ public class FinalLaunchSequence extends ReflectionSequence {
                                        //
                                        // "stepSetSourceLookupPath",   //$NON-NLS-1$
 
+                                       "stepRemoteConnection", //$NON-NLS-1$
                                        // For all launches except attach ones
                                        "stepNewProcess", //$NON-NLS-1$
                                        // For all attach launch only
@@ -546,6 +547,17 @@ public class FinalLaunchSequence extends ReflectionSequence {
                rm.done();
        }
 
+       /**
+        * This method used to be used to handle remote connections, that code has been folded
+        * into stepAttachToProcess which uses the equivalent implementation in
+        * fProcService.attachDebuggerToProcess
+        */
+       @Execute
+       @Deprecated(forRemoval = true)
+       public void stepRemoteConnection(final RequestMonitor rm) {
+               rm.done();
+       }
+
        /**
         * Start a new process if we are not dealing with an attach session
         * i.e., a local session, a remote session or a post-mortem (core) session.
```

and the same for the other removed method. I don't know how important it is for it to be in the array still.

This needs entry in N&N. Also as I recommend we deprecate for removal, in the CHANGELOG-API.md too.

Let me know if you need help or guidance on writing these items.

jonahgraham:
The final thing is it would be nice to have a test that covered this use case. It is somewhat out of the ordinary and I think it has a high chance of regressing again. I don't know what it would take to test, but I would consider looking at https://github.com/eclipse-cdt/cdt/blob/5007c8b3371a77ccd389660c24893d8f89b6579c/dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/tests/nonstop/GDBMultiNonStopRunControlTest.java

umairsair:
> Approved - subject to fixing up the API problems, which I think can be done like this:

what will happen if a sub-class is overriding `stepRemoteConnection` and doing connection in its own way? this may cause unexpected behaviors because now connection is being done in `attachDebuggerToProcess`..

jonahgraham:
Yes it may, hence N&N entry. I am certainly open to ideas that improve the situation, but the API contract of these methods is pretty undefined.

umairsair:
what could be the issues if we remove the API? I think its better that we get compile time error instead of getting strange runtime issue and others might be reporting issues..

jonahgraham:
You get runtime errors when removing API too. We can add an API exception and remove the methods.

It is a bit of a balancing act, and I'm not thrilled by the over exporting of API in this case that we are stuck with forever.

If we stay true to the API you should restructure the change. One idea is that we pull out the remaining method and call it from the other API methods with flags to say which behavior to do. But that seems like a lot of contrived code. 

The other option is to create a whole new FinalLaunchSequence and deprecate the existing one with no changes, like we did for gnumakefilebuilder, and then use the new FinalLaunchSequence.

umairsair:
BTW should I squash all commits or keep the revert commits separate?

jonahgraham:
You can squash them, or I will when I submit them. I'll keep the commit messages of all the squashed commits because the reference to when reverted commits are included is useful.

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
umairsair: kept the API wit deprecated annotation and its no more called..

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
umairsair: kept the API with deprecated annotation and its no more called..

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
```diff
@@ -133,11 +134,9 @@ protected String[] getExecutionOrder(String group) {
 					//
 					// "stepSetSourceLookupPath",   //$NON-NLS-1$
 
-					// For remote-attach launch only
```
Torbjorn-Svensson, line 139: I'm just thinking out loud now; Would it be safer to continue to call stepRemoteConnection until it is to be purged from the API?
With this change, it will be a "silent" error and it will be harder to find the reason why some other code that might extend this class stopped working. Sure - one should read the N&N, but in reality, this kind of change can easily be missed.

dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/tests/nonstop/MultiProcessRemoteTest.java
```diff
@@ -0,0 +1,239 @@
...
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+public class MultiProcessRemoteTest extends BaseParametrizedTestCase {
+	private static final String EXEC_NAME = "MultiThread.exe";
+
+	private DsfServicesTracker fServicesTracker;
+
+	private IGDBProcesses fGdbProcesses;
+	private ICommandControlService commandControl;
```
Torbjorn-Svensson, line 45: fCommandControl

dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/tests/nonstop/MultiProcessRemoteTest.java
```diff
@@ -0,0 +1,239 @@
...
+	public void launchGdbServerWithoutProgram_ConnectExecutable_And_DebugNewExecutable_noProgram() throws Throwable {
+		launchGdbServerWithProgram_ConnectExecutable_And_DebugNewExecutable();
+	}
+
+	private int getInitialProcessCount() {
+		return Boolean.valueOf((String) getLaunchAttribute(ITestConstants.LAUNCH_GDB_SERVER_WITHOUT_PROGRAM)) ? 0 : 1;
+	}
+
+	private Process launchApplication() throws IOException {
+		Process process = Runtime.getRuntime().exec(getBinary());
```
Torbjorn-Svensson, line 200: Wouldn't it be better to use the CDT ProcessFactory here?

dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/tests/nonstop/MultiProcessRemoteTest.java
umairsair: updated..

dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/tests/nonstop/MultiProcessRemoteTest.java
umairsair: no strong opinion.. we just need to launch a program and then attach to it.. IIUC, ProcessFactory is useful when there is some signal handling and pseudo-terminal related stuff is required..

Torbjorn-Svensson: I just wrote a long comment on this that github just trashed.

On Windows (I assume that the tests are executed on Windows as the extension is set to ".exe" for the test application), sending 2 signals to an application will just have the process vanish from the process list without letting it do cleanup of any shared resources. When calling Process.destroy() in java, the native call is going to be a call to [TerminateProcess](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-terminateprocess) and reading the documentation, it sounds just as bad as sending 2 signals.
To avoid having stale handles etc on the server, I suggest we use the CDT way instead that has been extensively tested over the years to handle this exact case.

For reference: [Process.destroy0 on Windows](https://github.com/openjdk/jdk17/blob/master/src/java.base/windows/native/libjava/ProcessHandleImpl_win.c#L369) and [Process.destroy0 on *nix](https://github.com/openjdk/jdk17/blob/74007890bb9a3fa3a65683a3f480e399f2b1a0b6/src/java.base/unix/native/libjava/ProcessHandleImpl_unix.c#L316)

jonahgraham: The tests haven't been maintained on Windows for a while, they once passed but not sure of current state. The `.exe` is explained here:

https://github.com/eclipse-cdt/cdt/blob/d2a5a57d01078a7ccf5919f2a429ba197dafec0a/dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/data/launch/src/Makefile#L17

That said, we should indeed be launching with process factory, here is an example of how gdbserver is launched in the base test case:

https://github.com/eclipse-cdt/cdt/blob/d2a5a57d01078a7ccf5919f2a429ba197dafec0a/dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/framework/BaseTestCase.java#L689

umairsair: > On Windows (I assume that the tests are executed on Windows as the extension is set to ".exe" for the test application)

They are run on linux only.. see `MultiProcessRemoteTest.beforeClass()` and `BaseTestCase.supportsNonStop()`.. regarding exe extension, see following

https://github.com/eclipse-cdt/cdt/blob/81581c2116f7b0a557493a2b3102bc68a4e070f9/dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/data/launch/src/Makefile#L17

> sending 2 signals to an application will just have the process vanish from the process list without letting it do cleanup of any shared resources. 

I need to understand how two signals are sent, even its just the case of windows (just for my knowledge)?

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
jonahgraham: Yes, this is what I recommended in https://github.com/eclipse-cdt/cdt/pull/336#pullrequestreview-1371709721 - it isn't just for the call, but code like this too: https://github.com/eclipse-embed-cdt/eclipse-plugins/blob/7115cc9957b44cabb495580f7a0e7452ce8647bf/plugins/org.eclipse.embedcdt.debug.gdbjtag.core/src/org/eclipse/embedcdt/debug/gdbjtag/core/dsf/GnuMcuFinalLaunchSequence.java#L64

.github/workflows/build-test.yml
```diff
@@ -37,6 +37,7 @@ jobs:
       run: |
         export DISPLAY=:99
         sudo Xvfb -ac :99 -screen 0 1280x1024x24 > /dev/null 2>&1 &
+        echo 0| sudo tee /proc/sys/kernel/yama/ptrace_scope
```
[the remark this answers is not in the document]
umairsair: it is working.. there was a problem in test case I believe.. we should first terminate the launch and it will kill the applications we launched and then check if there is still some application running, if yes, destroy it..

dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/tests/nonstop/MultiProcessRemoteTest.java
```diff
@@ -0,0 +1,242 @@
...
+		assertTrue("Process-2 is not started properly as no threads found for it", ctx != null && ctx.length > 0);
+
+		ctx = debugNewProcess(execPath, attributes);
+		assertTrue("Process-3 is not started properly as no threads found for it", ctx != null && ctx.length > 0);
+	}
+
+	@Test
+	public void launchGdbServerWithoutProgramAndStartMulipleNewExecutables_noProgram() throws Throwable {
+		launchGdbServerWithProgramAndStartMulipleNewExecutables();
+	}
```
jonahgraham, line 130: Was this test supposed to do more?

dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/tests/nonstop/MultiProcessRemoteTest.java
Torbjorn-Svensson: > > sending 2 signals to an application will just have the process vanish from the process list without letting it do cleanup of any shared resources.
> 
> I need to understand how two signals are sent, even its just the case of windows (just for my knowledge)?

https://github.com/eclipse-cdt/cdt/blob/7c6bd5bdcb4cf8c31050aeeb102250a27a157728/core/org.eclipse.cdt.core.native/src/org/eclipse/cdt/utils/spawner/Spawner.java#L334

As you can see in this method, both the `terminate()` and `kill()` methods are called and both of them sends a signal to the application.
If you want to see what happens, you can simply create a C/C++ application and have it register a signal handler and then sleep for  1 min or something. When running the application in a DOS prompt, try to hit `Ctrl + c`, wait a few sec and then hit `Ctrl + c` again and see what happens. To be more visual confirmation, you can have a printout in the signal handler to see if it's called or not.

jonahgraham (commented):
Thank you @umairsair for creating these tests.

dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/framework/BaseTestCase.java
```diff
@@ -680,7 +684,17 @@ private void launchGdbServer() throws Exception {
 				String server = (String) launchAttributes.get(ATTR_DEBUG_SERVER_NAME);
 				String port = (String) launchAttributes.get(IGDBLaunchConfigurationConstants.ATTR_PORT);
 				String program = (String) launchAttributes.get(ICDTLaunchConfigurationConstants.ATTR_PROGRAM_NAME);
-				String[] commandLine = { server, ":" + port, program };
+				boolean multi = Boolean.valueOf((String) launchAttributes.get(ITestConstants.LAUNCH_GDB_SERVER_MULTI));
+				boolean noprogram = Boolean
+						.valueOf((String) launchAttributes.get(ITestConstants.LAUNCH_GDB_SERVER_WITHOUT_PROGRAM));
+				List<String> commandArgs = new ArrayList<>();
+				commandArgs.add(server);
+				if (multi)
```
Torbjorn-Svensson, line 692: Can you add curly bracers around the single line statements to make a clear scope?
I know it's not mandatory, but most of the code in CDT does use it even when it's not mandatory to make the intended scope clearer.
This comment applies to other places in the PR too.

dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/tests/nonstop/MultiProcessRemoteTest.java
Torbjorn-Svensson: I think the answer to this question is here:

https://github.com/eclipse-cdt/cdt/pull/336/files#diff-4cfbb32f09206759612425a8f3253d379d272eeb8e2e9165d0cb3d355e3a4437R86 (line 86 in this file)

jonahgraham: Thanks @Torbjorn-Svensson for explaining. I wish we didn't need that, but resolving this now.

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
umairsair: the more I think about it, I get confused that why should we convolute our code for this case.. the current solution seems more inline with what we do for non-attach remote launches..

however, I have worked on your suggestion and kept the logic inside the two methods with some comments for future when removing the APIs.. but I personally don't like this approach :stuck_out_tongue:

dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/tests/nonstop/MultiProcessRemoteTest.java
umairsair: > have it register a signal handler

thats what I said in my first response that its useful when we are dealing with application which does special handling for signals like GDB and gdbserver does, but it was not the case for launching application here..

but thanks for the explanation.. I now understand how Spawner handles such cases..

dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF
```diff
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.cdt.dsf.gdb;singleton:=true
-Bundle-Version: 7.0.0.qualifier
+Bundle-Version: 7.1.0.qualifier
```
umairsair, line 6: should it remain on 7.1 or it should be 7.0.100 ?

dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/tests/ITestConstants.java
```diff
@@ -64,4 +64,9 @@ public class ITestConstants {
 
 	// Attribute that allows a test to request not to start gdbserver even if the session is a remote one
 	public static final String LAUNCH_GDB_SERVER = TestsPlugin.PLUGIN_ID + ".launchGdbServer";
+	// Attribute that tells whether gdbserver should be launched without program
+	public static final String LAUNCH_GDB_SERVER_WITHOUT_PROGRAM = TestsPlugin.PLUGIN_ID
+			+ ".launchGdbServerWithoutProgram";
+	// Attribute that tells whether gdbserver should be launched with --multi argument
+	public static final String LAUNCH_GDB_SERVER_MULTI = TestsPlugin.PLUGIN_ID + ".launchGdbServerMulti";
```
umairsair, line 71: at some point, I wish to add this option for gdbserver in `non-attach remote launch` so that we can also provide muliprocess support there as well.. currently multi-process support is only available for `attach remote launch` (just talking about remote debugging, its available for non-remote)

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
```diff
@@ -656,22 +662,14 @@ public void stepAttachToProcess(final RequestMonitor requestMonitor) {
 	 * If we are dealing with an remote attach session, perform the attach to debugger.
 	 * Bug 528145
 	 * @since 6.6
+	 *
+	 * When removing this method, do inline method refactoring
 	 */
-	@Execute
```
umairsair, line 660: probably we should keep Execute annotation, WDYS?

NewAndNoteworthy/CHANGELOG-API.md
```diff
@@ -665,3 +665,11 @@ The class BuiltinDetctionArgsGeneric will be removed. Use the correctly
 spelled BuiltinDetectionArgsGeneric instead.
 
 - org.eclipse.cdt.jsoncdb.core.participant.Arglets.BuiltinDetctionArgsGeneric
+
+## API Removals after June 2025
```
jonahgraham, line 669: I have simplified the API document by removing requirement to double-enter the contents - see #359

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
jonahgraham: Sorry @umairsair - I think there was some confusion as to what the request in this comment was. Your previous change removed the call to stepRemoteConnection, so I think all that was missing was this part of my comment:

```diff
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
@@ -134,6 +134,7 @@ public class FinalLaunchSequence extends ReflectionSequence {
                                        //
                                        // "stepSetSourceLookupPath",   //$NON-NLS-1$
 
+                                       "stepRemoteConnection", //$NON-NLS-1$
                                        // For all launches except attach ones
                                        "stepNewProcess", //$NON-NLS-1$
                                        // For all attach launch only
```


For API reasons the way you have done it now means that we won't break anything which is great and goes beyond what I was hoping you would do because the more I look around the more references to this code in extenders.

But it does leave the code more convoluted.

Also, the stepAttachRemoteToDebugger should still be in this list. The current version is a bit of a mishmash between both cases. I will provide an updated commit that I hope addresses this so we can get it merged in.

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
jonahgraham: yes

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
```diff
@@ -639,6 +643,8 @@ public void stepAttachToProcess(final RequestMonitor requestMonitor) {
 				fProcService.attachDebuggerToProcess(
 						fProcService.createProcessContext(fCommandControl.getContext(), Integer.toString(pid)),
 						new DataRequestMonitor<IDMContext>(getExecutor(), requestMonitor));
+			} else if (fGDBBackend.getSessionType() == SessionType.REMOTE) {
```
jonahgraham, line 646: Remove this and call the method from the launch sequence list.

dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF
jonahgraham: 7.1 because of new API method getBinary

jonahgraham (commented):
I am about to provide some new commits. @umairsair if you are around today and can review them that would be great, if not and the tests pass I am going to merge this so it can get in M1 today and we can pick up additional work later if required.

dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/src/org/eclipse/cdt/tests/dsf/gdb/tests/ITestConstants.java
jonahgraham: Sounds great.

dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/FinalLaunchSequence.java
```diff
@@ -134,12 +134,14 @@ protected String[] getExecutionOrder(String group) {
...
 
-					// For remote-attach launch only (deprecated)
+					// For remote-attach launch only (deprecated, see javadocs)
 					"stepRemoteConnection", //$NON-NLS-1$
 					// For all launches except attach ones
 					"stepNewProcess", //$NON-NLS-1$
 					// For all attach launch only
 					"stepAttachToProcess", //$NON-NLS-1$
+					// For remote attach launch only (deprecated, see javadocs)
+					"stepAttachRemoteToDebugger", //$NON-NLS-1$
```
umairsair, line 144: if we want to keep it then we need `fGDBBackend.getIsAttachSession()` check as well.. previously `stepAttachToProcess` was taking care of it..

umairsair: in fact also need `fGDBBackend.getSessionType() == SessionType.REMOTE` check..

umairsair:
LGTM.. Thanks @jonahgraham !

test failure seems a glitch.. can you please re-trigger it?

jonahgraham:
Thank you @umairsair for this work. And an extra thank you for writing the new tests, I think that will really help keep this stable as other use cases are worked on.

jonahgraham:
Just to confirm that I think we made the right decision with how this is implemented now I have come across this code that extends CDT (modified slightly for privacy):

```java

	@Override
	protected String[] getExecutionOrder(String group) {
		if (GROUP_TOP_LEVEL.equals(group)) {
			// Initialize the list with the steps from the base class
			List<String> orderList = new ArrayList<>(Arrays.asList(super.getExecutionOrder(GROUP_TOP_LEVEL)));
			orderList.add(orderList.indexOf("stepAttachRemoteToDebugger") + 1, "newstep")
```

In this case, "newstep" would have been added as the first item in the launch sequence had we removed `stepAttachRemoteToDebugger` fully.

PR #339, opened 2023-03-27
Do not error out failing to resolve function in dependend context

i-garrison:
Detect if function set failed to resolve while instantiating for template-id which still has dependend arguments. Since resolution still can succeed later via ADL when template-id is fully instantiated, do not error out immediately and return original evaluation to allow trying to instantiate later.

This partially fixes a problem with looking at GDB sources where `enum_flags` uses ADL to restrict args to binary operators over enumerations.


jonahgraham (approved):

jonahgraham:
Thanks @i-garrison

Issue #341, opened 2023-03-27
Support IEditorAssociationOverride


ghentschke:
We are currently working on a new [LSP based C/C++ Editor](https://github.com/Bachmann-electronic-GmbH/eclipse-cdt-lsp).
Currently I'am working on this [issue](https://github.com/Bachmann-electronic-GmbH/eclipse-cdt-lsp/issues/37)

The file association for C/C++ source or header file to an editor should be dependent on the project settings. Therefor the user can select to prefer the new LSP based editor for his project.

The current implementation to override the default association of an editor to a file is based on this extension point: `org.eclipse.ui.ide.editorAssociationOverride`

The problem is that the cdt method `org.eclipse.cdt.internal.ui.util.EditorUtility.getEditorID(IEditorInput, Object)`  does not care about the overriden default editor by the implementation of `org.eclipse.ui.ide.editorAssociationOverride`.

The fix are 5 lines of code in `org.eclipse.cdt.internal.ui.util.EditorUtility.getEditorID(IEditorInput, Object)`:
![image](https://user-images.githubusercontent.com/123444711/227997331-28b040ec-b5ff-4053-b352-1bbcf031a3ae.png)

The charm of this solution is, that the default C/C++ editor can be overriden by the  `org.eclipse.ui.ide.editorAssociationOverride` extension point.

jonahgraham:
This looks good to me (but may need a null check on return of getEditorDescriptor?). Can you provide a PR so we can proceed?

ghentschke:
I'll provide a pull request. Null check should not be necessary. The method `getEditorDescrioptor` will throw a `PartInitException` if no valid editor can be found.

jonahgraham:
Fixed by #345

PR #345, opened 2023-03-28
Support IEditorAssociationOverride (#341)

ghentschke:
Fix for #341


ghentschke:
@jonahgraham can you please review?

jonahgraham:
Sorry I haven't reviewed yet. I hope to get to it soon.

jonahgraham:
I have a concern that there is some weird corner case that the behaviour changes unexpectedly, but if the automated tests don't cover it I am happy to merge this. The code on selecting editors is way too complicated and I am suspicious at the value of most of that complexity. This new code will be the taken path most of the time now, which is good.

I force pushed to fix the whitespace error and I will merge once we get a clean run.

PR #367, opened 2023-04-14
Set environment for spawning GNU tool processes

jld01:


jld01:
I'm looking at the test failure.

jld01:
I think the test failure is unrelated to this PR. `PlainC_CompletionTests` passes on my desktop.

Issue #361, opened 2023-04-10
Support invoking objdump on binary files for 'MinGW GCC' and 'Cross GCC' toolchains


jld01:
When building CDT projects using a native _Linux GCC_ or _Cygwin GCC_ toolchain, the user can double click on the resulting binary file to invoke `objdump` and present the resulting output in a read-only editor part. It would be beneficial to provide equivalent behaviour when using _MinGW GCC_ toolchains and _Cross GCC_ toolchains.

For  _Linux GCC_, this capability is implemented via the `DefaultGnuToolFactory` and CDT expects to find `objdump` on the global `PATH`.

A few issues:
- _Cross GCC_ toolchains use the `org.eclipse.cdt.core.ELF` binary parser rather than the `org.eclipse.cdt.core.GNU_ELF` binary parser at present. Easy to fix.
- We could create an equivalent `org.eclipse.cdt.core.GNU_PE64` binary parser for _MinGW GCC_ toolchains. Easy to fix.
- _Cross GCC_ toolchains will typically provide a build of `objdump` with a prefix (eg `arm-none-eabi-objdump`) but the `IGnuToolFactory` implementation currently has no context from which to determine that prefix.
- It is much less common for GNU tools to be located on the global `PATH` on Windows hosts or when using a cross compiler. We need a mechanism to lookup the project containing the binary file and from there lookup the per-project `PATH`.

jantje:
my 2 cent

>It is much less common for GNU tools to be located on the global PATH on Windows hosts or when using a cross compiler. We need a mechanism to lookup the project containing the binary file and from there lookup the per-project PATH.

You are probably aware but to be sure: The current MBS allows to extend the path environment variable in the settings.
![afbeelding](https://user-images.githubusercontent.com/1175841/230973074-88aeba00-08dd-4b3b-a8d8-e23a146d6de7.png)

In Arduino world cross GCC, MinGW, MVC are called **tool providers**. These tool providers have a "tool path" property.
All commands are executed with FQN (in MBS talk ${provider tool path including /}${COMMAND}).
An environment variable provider can provide a value for the variable "provider tool path including /".

I have had no luck with the path extension used by MBS. I like the FQN approach.

jld01:
Thank you for your input, @jantje.

[...]

PR #368, opened 2023-04-15
Set tool prefix for spawning GNU tool processes

jld01:
Allows the GNU tool prefix to be specified by a CDT build variable.

Modifies the Cross GCC toolchain description to provide the GNU tool prefix.

Part of #361

### DIFF
--- a/core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/DefaultGnuToolFactory.java
+++ b/core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/DefaultGnuToolFactory.java
@@ -10,7 +10,7 @@
  *
  * Contributors:
  *     QNX Software Systems - initial API and implementation
- *     John Dallaway - set environment for spawning GNU tool processes (#361)
+ *     John Dallaway - set environment and tool prefix (#361)
  *******************************************************************************/
 package org.eclipse.cdt.utils;
 
@@ -19,11 +19,15 @@ import java.util.Arrays;
 
 import org.eclipse.cdt.core.CCorePlugin;
 import org.eclipse.cdt.core.ICExtension;
+import org.eclipse.cdt.core.cdtvariables.CdtVariableException;
+import org.eclipse.cdt.core.cdtvariables.ICdtVariable;
 import org.eclipse.cdt.core.envvar.IEnvironmentVariable;
 import org.eclipse.cdt.core.settings.model.ICConfigExtensionReference;
 import org.eclipse.cdt.core.settings.model.ICConfigurationDescription;
 import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
 
 public class DefaultGnuToolFactory implements IGnuToolFactory {
 	protected ICExtension fExtension;
@@ -94,7 +98,7 @@ public class DefaultGnuToolFactory implements IGnuToolFactory {
 		ICConfigExtensionReference ref = fExtension.getConfigExtensionReference();
 		String value = ref.getExtensionData("addr2line"); //$NON-NLS-1$
 		if (value == null || value.length() == 0) {
-			value = "addr2line"; //$NON-NLS-1$
+			value = getToolPrefix() + "addr2line"; //$NON-NLS-1$
 		}
 		return new Path(value);
 	}
@@ -103,7 +107,7 @@ public class DefaultGnuToolFactory implements IGnuToolFactory {
 		ICConfigExtensionReference ref = fExtension.getConfigExtensionReference();
 		String value = ref.getExtensionData("objdump"); //$NON-NLS-1$
 		if (value == null || value.length() == 0) {
-			value = "objdump"; //$NON-NLS-1$
+			value = getToolPrefix() + "objdump"; //$NON-NLS-1$
 		}
 		return new Path(value);
 	}
@@ -121,7 +125,7 @@ public class DefaultGnuToolFactory implements IGnuToolFactory {
 		ICConfigExtensionReference ref = fExtension.getConfigExtensionReference();
 		String value = ref.getExtensionData("c++filt"); //$NON-NLS-1$
 		if (value == null || value.length() == 0) {
-			value = "c++filt"; //$NON-NLS-1$
+			value = getToolPrefix() + "c++filt"; //$NON-NLS-1$
 		}
 		return new Path(value);
 	}
@@ -130,7 +134,7 @@ public class DefaultGnuToolFactory implements IGnuToolFactory {
 		ICConfigExtensionReference ref = fExtension.getConfigExtensionReference();
 		String value = ref.getExtensionData("strip"); //$NON-NLS-1$
 		if (value == null || value.length() == 0) {
-			value = "strip"; //$NON-NLS-1$
+			value = getToolPrefix() + "strip"; //$NON-NLS-1$
 		}
 		return new Path(value);
 	}
@@ -139,7 +143,7 @@ public class DefaultGnuToolFactory implements IGnuToolFactory {
 		ICConfigExtensionReference ref = fExtension.getConfigExtensionReference();
 		String value = ref.getExtensionData("nm"); //$NON-NLS-1$
 		if (value == null || value.length() == 0) {
-			value = "nm"; //$NON-NLS-1$
+			value = getToolPrefix() + "nm"; //$NON-NLS-1$
 		}
 		return new Path(value);
 	}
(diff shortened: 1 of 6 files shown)


cross/org.eclipse.cdt.build.crossgcc/src/org/eclipse/cdt/internal/build/crossgcc/CrossBuildMacroSupplier.java
```diff
@@ -0,0 +1,91 @@
...
+import org.eclipse.cdt.utils.IGnuToolFactory;
+import org.eclipse.core.runtime.Status;
+
+public class CrossBuildMacroSupplier implements IConfigurationBuildMacroSupplier {
+
+	/**
+	 * A CDT build variable defining the GNU tool prefix for use by
+	 * an {@link org.eclipse.cdt.utils.IGnuToolFactory} implementation.
+	 */
+	private static class GnuToolPrefixMacro implements IBuildMacro {
```
Torbjorn-Svensson, line 32: Would it be possible to put this class as a public class (public API) instead with the option id as a parameter to the constructor?
If left as is, then a downstream vendor with a custom toolchain definition would need to more or less copy this implementation and put all that in there code base. That by itself might not be an issue, but if the code base is not EPL-2.0, then it will be a mess for the vendor and perhaps would require them to have 1 additional plugin just to have the proper license on the single class.

core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/DefaultGnuToolFactory.java
```diff
@@ -161,4 +165,28 @@ protected String[] getEnvironment() {
...
+	/** @since 8.2 */
+	protected String getToolPrefix() {
+		ICConfigExtensionReference ref = fExtension.getConfigExtensionReference();
+		ICConfigurationDescription cfg = ref.getConfiguration();
+		ICdtVariable[] userVars = CCorePlugin.getUserVarSupplier().getMacros(cfg);
+		ICdtVariable var = Arrays.stream(userVars).filter(v -> v.getName().equals(GNU_TOOL_PREFIX_VARIABLE)).findFirst()
+				.orElse(null);
+
+		// if user-defined variable not found, look for system variable provided by toolchain integration
+		if (null == var) {
```
Torbjorn-Svensson, line 178: I'm not against using the "null == var" syntax rather than "var == null", but it looks like the file are using the latter way of writing the expression rather than what you used.
Do you think we should continue with the "var == null" order throughout the file or have a mix?
I'm okay regardless what you end up with!

cross/org.eclipse.cdt.build.crossgcc/plugin.xml
```diff
@@ -32,7 +34,7 @@
             targetTool="cdt.managedbuild.tool.gnu.cross.c.linker;cdt.managedbuild.tool.gnu.cross.cpp.linker;cdt.managedbuild.tool.gnu.archiver">
          <targetPlatform
                archList="all"
-               binaryParser="org.eclipse.cdt.core.ELF"
+               binaryParser="org.eclipse.cdt.core.GNU_ELF"
```
Torbjorn-Svensson, line 37: It looks like the binary parser attribute is written out in the .cproject file of every created project.
While this in general would be good, for this particular change, it looks like all the existing projects will be stuck on org.eclipse.cdt.core.ELF until a manual change is done. Do you have any idea how to automatically migrate existing projects to the org.eclipse.cdt.core.GNU_ELF parser instead?

Torbjorn-Svensson (commented):
Thanks for working on this John!

I see that you are handling the prefix here and I think that the path is handled with the environment provider, but would it make sense to also expose the path using a build variable?

cross/org.eclipse.cdt.build.crossgcc/src/org/eclipse/cdt/internal/build/crossgcc/CrossBuildMacroSupplier.java
jld01: How about an `AbstractGnuToolPrefixMacro` class which is missing the `getStringValue()` method tucked in to the Java package alongside `IGnuToolFactory`?

core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/DefaultGnuToolFactory.java
jld01: I use `null == var`  habitually, but I don't think this is worth arguing over. Will change for consistency.

cross/org.eclipse.cdt.build.crossgcc/plugin.xml
jld01: Will think about this. Perhaps as a separate issue. It is easy to switch the active  binary parser in the CDT project settings UI.

Do you think there could be any reason why someone would NOT wish to update their parser? I think it may be wise to prompt for confirmation.

Cc: @jonahgraham

jld01:
> Thanks for working on this John!
> 
> would it make sense to also expose the path using a build variable?

In the general case, a toolchain can end up on the PATH of a build configuration via many routes. The CDT builder just assumes that the ITool commands have made it onto the PATH by some route and I think we can do the same for the various GNU binutils. It seems unlikely that the user will wish to access these tools at a different location to the compiler and linker.

cross/org.eclipse.cdt.build.crossgcc/src/org/eclipse/cdt/internal/build/crossgcc/CrossBuildMacroSupplier.java
jld01: I've now looked at implementing an abstract class. The benefits are modest but probably worthwhile. Does this work for you?
```
package org.eclipse.cdt.managedbuilder.macros;

import org.eclipse.cdt.utils.IGnuToolFactory;
import org.eclipse.core.runtime.Status;

/**
 * A CDT build variable defining the GNU tool prefix for use by
 * an {@link org.eclipse.cdt.utils.IGnuToolFactory} implementation.
 *
 * @since 9.6
 */
public abstract class AbstractGnuToolPrefixMacro implements IBuildMacro {

	@Override
	public String getName() {
		return IGnuToolFactory.GNU_TOOL_PREFIX_VARIABLE;
	}

	@Override
	public int getValueType() {
		return IBuildMacro.VALUE_TEXT;
	}

	@Override
	public int getMacroValueType() {
		return getValueType();
	}

	@Override
	public String[] getStringListValue() throws BuildMacroException {
		throw new BuildMacroException(
				Status.error("Not a string list build macro: " + IGnuToolFactory.GNU_TOOL_PREFIX_VARIABLE)); //$NON-NLS-1$
	}

	@Override
	abstract public String getStringValue() throws BuildMacroException;
}
```

jld01: I thought about passing an `IOption` through the constructor, but this seems too limiting. Some implementations may not use an `IOption`. Perhaps they provide a constant value or obtain the value by other means such as an `IPreferenceService`.

jld01: We could provide a helper method (callable by the concrete implementation) in the abstract class:
```
public String getStringValue(IOption option) throws BuildMacroException {
	try {
		return option.getStringValue();
	} catch (BuildException e) {
		throw new BuildMacroException(Status.error(
				"Error getting build macro string value: " + IGnuToolFactory.GNU_TOOL_PREFIX_VARIABLE, e)); //$NON-NLS-1$
	}
}
```

jld01: @Torbjorn-Svensson, I've implemented the abstract class with `IOption` helper method and updated the PR accordingly.

core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/DefaultGnuToolFactory.java
jld01: Fixed.

cross/org.eclipse.cdt.build.crossgcc/plugin.xml
Torbjorn-Svensson: Right now, I do not see any reason why you would not want that.

Maybe we should invert the question and rather ask us why anyone would like to use the org.eclipse.cdt.core.ELF parser in the first place and if we have no clear answer to that, perhaps the org.eclipse.cdt.core.ELF and org.eclipse.cdt.core.GNU_ELF should be the same parser from now on?

Torbjorn-Svensson (approved):
Other than 2 more places of the "null" part being on the left side (that I do not really care about), I think this looks good.

jld01:
Thank you for the helpful reviews, @Torbjorn-Svensson. Perhaps we can discuss automatic update of the binary parser in legacy projects on the next CDT project call.

cross/org.eclipse.cdt.build.crossgcc/plugin.xml
jld01: There are, of course, non-GNU toolchains that generate ELF executable files. At present, the existence of the ELF parser and the GNU_ELF parser allows common code to determine whether use of GNU tools such as `objcopy` should be attempted. In theory, an extender could integrate equivalent non-GNU tools.

PR #374, opened 2023-04-21
Improved GCCToolChain.stripCommand()

ewaterlander:
GCCToolChain.stripCommand() assumed that all resources are at the end of the command, like in the old version of
GCCToolChain.getResourcesFromCommand() which was fixed in PR #311 (see commit a89ce59df23). Now stripCommand() is in line with getResourcesFromCommand().


jonahgraham:
The failing test is flaky and covered by #194

jonahgraham (approved):
I think this is fine - AFAICT there are no existing tests for any of this functionality so it is hard to verify if anything regresses because of this change.

If you want to reduce the chance of future regressions adding a unit test or two would be very welcome.

PR #380, opened 2023-04-25
Wrong active build configuration for Core Build projects.

### LINKED ISSUE #378: Launch configuration runs wrong build configuration of Core Build project.
The Core Build Makefile/CMake projects create too simple launch configurations. To run the binary of my Core Build Makefile project I need to create a "C/C++ Application" launch configuration to be able to set options for the launching. For debug I create a "C/C++ Remote Application" launch configuration.

The problem is that a normal "C/C++ Application" launch configuration can't control which build configuration of a Core Build project to run. Therefore the incremental build can build with wrong options.
Once the project was once compiled for Debug, using the Core Build launch configuration, the  "C/C++ Application" will always compile for debug, no matter what the launch type setting in the launch bar is.

**To Reproduce**
1. Create a new (Experimental) Makefile project with a hello world program.
2. Build it, run it. With the automatically created launch configuration.
3. Set launch type to Debug in the LaunchBar
4. Build for Debug, and run a debug session.
5. Set launch type back to Run in the LaunchBar.
6. Clean the project.
7. Build and run the project in Run mode.
8. Create a "C/C++ Application" launch configuration, select the Run binary, without debug.
9. Select the new launch configuration.
10. Run the new launch configuration.

You can see the project is build with flag -g. The expected flag would be -O2.

image

The normal (not Core Build) launch configurations have no option to select a specific build configuration of a Core Build project that could be used to force a specific build configuration (which can be done for managed build projects).

image

**Expected behavior**
The proper build configuration should be built based on the launch type set in the launch bar.

Workaround: Select the core build launch config, build for Run, select the "C/C++ Application" launch config, and launch.

OS: Linux.

ewaterlander:
The CoreBuildLaunchBar tracker always made a non default build configuration the active build configuration. In other words, it always made the debug build configuration active.
This caused wrong build flags if a non core build launch configuration was used to launch a core build project binary.

Fixed the CoreBuildLaunchBar tracker to set the build configuration to active that matches the launchBar mode.

Fixes #378

### DIFF
--- a/debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchBarTracker.java
+++ b/debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchBarTracker.java
@@ -100,6 +100,7 @@ public class CoreBuildLaunchBarTracker implements ILaunchBarListener, ILaunchTar
 		}
 
 		if (project == null || !configManager.supports(project)) {
+			// The project is not a Core Build project.
 			return;
 		}
 
@@ -109,7 +110,23 @@ public class CoreBuildLaunchBarTracker implements ILaunchBarListener, ILaunchTar
 		lastDescriptor = descriptor;
 		lastTarget = target;
 
-		// Pick build config based on toolchain for target
+		/*
+		 * Core build projects do not work with the concept of active build
+		 * configurations, like managed build projects. Instead they rely
+		 * on the launch mode set in the launchBar. A core build launch
+		 * configuration looks at the launchBar mode to pick the build
+		 * configuration. Core build projects still have an active build
+		 * configuration, but it is hidden for the user and ignored for
+		 * launching.
+		 *
+		 * Core build launch configurations have no options to set, so it is
+		 * likely that users will use non core build launch configurations to
+		 * launch a core build project binary. Non core build launch
+		 * configurations typically launch the active build configuration. The
+		 * active build configuration needs to match the launchBar mode.
+		 */
+
+		// Pick core build config based on launch mode and toolchain for target
 		// Since this may create a new config, need to run it in a Job
 		Job job = new Job(InternalDebugCoreMessages.CoreBuildLaunchBarTracker_Job) {
 			@Override
@@ -120,11 +137,12 @@ public class CoreBuildLaunchBarTracker implements ILaunchBarListener, ILaunchTar
 					Collection<IToolChain> tcs = toolChainManager.getToolChainsMatching(properties);
 					ICBuildConfiguration buildConfig = null;
 					if (!tcs.isEmpty()) {
-						// First, see if any existing non default build configs match the target properties
 						configs: for (IBuildConfiguration config : finalProject.getBuildConfigs()) {
-							if (!config.getName().equals(IBuildConfiguration.DEFAULT_CONFIG_NAME)) {
-								ICBuildConfiguration testConfig = configManager.getBuildConfiguration(config);
-								if (testConfig != null && !(testConfig instanceof ErrorBuildConfiguration)) {
+							ICBuildConfiguration testConfig = configManager.getBuildConfiguration(config);
+							if (testConfig != null && !(testConfig instanceof ErrorBuildConfiguration)) {
+								// Match launch mode run/debug.
+								if (testConfig.getLaunchMode().equals(lastMode.getIdentifier())) {
+									// Match toolchain.
 									for (IToolChain tc : tcs) {
 										if (testConfig.getToolChain().equals(tc)) {
 											buildConfig = testConfig;
--- a/debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchBarTracker.java
+++ b/debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchBarTracker.java
@@ -113,17 +113,18 @@ public class CoreBuildLaunchBarTracker implements ILaunchBarListener, ILaunchTar
 		/*
 		 * Core build projects do not work with the concept of active build
 		 * configurations, like managed build projects. Instead they rely
-		 * on the launch mode set in the launchBar. A core build launch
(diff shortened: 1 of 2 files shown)


ewaterlander:
How to fix the test problems? I see no relation with my change.

jonahgraham:
> How to fix the test problems? I see no relation with my change.

You are making the first change to that bundle since our last release so you need to bump the version in MANIFEST.MF.

ewaterlander:
> > How to fix the test problems? I see no relation with my change.
> 
> You are making the first change to that bundle since our last release so you need to bump the version in MANIFEST.MF.

OK. I'm just closing down. I'm back on Friday.
best regards,
Erwin

jonahgraham:
> > > How to fix the test problems? I see no relation with my change.
> > 
> > 
> > You are making the first change to that bundle since our last release so you need to bump the version in MANIFEST.MF.
> 
> OK. I'm just closing down. I'm back on Friday. best regards, Erwin

np. If there is no other issue other than the version bump I can take care of it when I get to reviewing it.

jonahgraham (commented):
This looks reasonable to me - but I fear I don't have enough knowledge of this.

@betamaxbandit or @mbooth101 can one of you review this change please? I think both of you have much more knowledge of core build + launchbar in this case and it will also help to ensure that this change does not regress your use cases.

debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchBarTracker.java
```diff
@@ -109,7 +110,23 @@ private void setActiveBuildConfig(ILaunchMode mode, ILaunchDescriptor descriptor
...
 
-		// Pick build config based on toolchain for target
+		/*
+		 * Core build projects do not work with the concept of active build
+		 * configurations, like managed build projects. Instead they rely
+		 * on the launch mode set in the launchBar. A core build launch
+		 * configuration looks at the launchBar mode to pick the build
+		 * configuration. Core build projects still have an active build
+		 * configuration, but it is hidden for the user and ignored for
+		 * launching.
```
betamaxbandit, line 120: * Core build projects do not work with the concept of active build
		 * configurations, like managed build projects. Instead they rely
		 * on the launch mode and launch target set in the launchBar. A core build launch
		 * configuration looks at the launchBar launch mode and then the set of toolchains
		 * associated with the launch target to pick the build configuration.
		 * Core build projects still have an active build
		 * configuration, but it is hidden for the user and ignored for
		 * launching.

debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchBarTracker.java
```diff
@@ -109,7 +110,23 @@ private void setActiveBuildConfig(ILaunchMode mode, ILaunchDescriptor descriptor
...
+		 *
+		 * Core build launch configurations have no options to set, so it is
+		 * likely that users will use non core build launch configurations to
+		 * launch a core build project binary. Non core build launch
+		 * configurations typically launch the active build configuration. The
+		 * active build configuration needs to match the launchBar mode.
+		 */
```
betamaxbandit, line 127: * Core build launch configurations have no options to set, so it is
		 * possible that users will use non core build launch configurations to
		 * launch a core build project binary. Non core build launch
		 * configurations typically launch the active build configuration. The
		 * active build configuration needs to match the launchBar launch mode.
		 */

debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchBarTracker.java
betamaxbandit, line 1: I am able to reproduce the steps and problem in https://github.com/eclipse-cdt/cdt/issues/378#issue-1680959919.
I have executed the patch set in my own environment (on Windows) and see no problems.
For the case where non core build launch configurations are used this looks like a positive improvement.

debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchBarTracker.java
jonahgraham, line 127:
```suggestion
		/*
		 * Core build projects do not work with the concept of active build
		 * configurations, like managed build projects. Instead they rely
		 * on the launch mode and launch target set in the launchBar. A core
		 * build launch configuration looks at the launchBar launch mode
		 * and then the set of toolchains associated with the launch target
		 * to pick the build configuration. Core build projects still have an
		 * active build configuration, but it is hidden for the user and ignored for
		 * launching.
		 *
		 * Core build launch configurations have no options to set, so it is
		 * possible that users will use non core build launch configurations to
		 * launch a core build project binary. Non core build launch
		 * configurations typically launch the active build configuration. The
		 * active build configuration needs to match the launchBar launch mode.
		 */
```

debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchBarTracker.java
jonahgraham: I converted this comment to a suggested code change by using this button in the review:

image

This is one of the few features that makes github reviewing better than gerrit IMHO.

debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchBarTracker.java
jonahgraham: same as above.

jonahgraham (changes requested):

ewaterlander (commented):
Thanks for the review.

ewaterlander:
Do I need to do something to unblock merging?

jonahgraham:
No. I was waiting for build to complete before I merged.

jonahgraham (approved):

Issue #354, opened 2023-04-04
LSP / clangd support for C/C++ in CDT


jonahgraham:
An important part of the future of Eclipse CDT is improving language support. The long term future of this is expected to be done by integrating clangd via LSP. 

@ghentschke [announced](https://www.eclipse.org/lists/cdt-dev/msg35445.html) in February that she and others (Bachmann electronic, Renesas and Espressif Systems) are working on that in https://github.com/Bachmann-electronic-GmbH/eclipse-cdt-lsp repo.

This was also extensively [discussed](https://www.eclipse.org/lists/cdt-dev/msg35452.html) at the last monthly CDT call.

This issue is to track progress on migrating the work started in  https://github.com/Bachmann-electronic-GmbH/eclipse-cdt-lsp to be hosted by Eclipse CDT itself. Steps to migrate the work:

- [x] IP Approval as the code change is > 1000 lines.
- [x] Request EF Webmaster move the repo to the eclipse-cdt GitHub organization
- [x] Submit committer elections
- [x] Update links from https://github.com/Bachmann-electronic-GmbH/eclipse-cdt-lsp -> https://github.com/eclipse-cdt/cdt-lsp/
- [x] Create initial pre-releases of the code

jonahgraham:
IP approval requested in https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/7767

jonahgraham:
> IP approval requested in https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/7767

Done.

The request to transfer has been started here: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3048

@ghentschke can you please comment on that issue so that Eclipse Webmaster has your confirmation and contact details please?

jonahgraham:
> The request to transfer has been started here: https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3048

Transfer complete  - https://github.com/eclipse-cdt/cdt-lsp

Election has now started for @ghentschke   https://projects.eclipse.org/projects/tools.cdt/elections/election-gesa-hentschke-committer-eclipse-cdt%E2%84%A2-cc-development-tooling

jonahgraham:
On the CDT call today we agreed to aim for a May 21 pre-release of the LSP work in conjunction with 2023-06 M3.

fgelm01:
[...]

PR #383, opened 2023-04-25
Enable toggling breakpoints in LSP CEditor

jonahgraham:
This is an example of where we have somewhat an inversion of dependencies. The existing CDT code assumes it knows about all types of editors at compile time. In this case the LSP C Editor is a new type. However, rather than creating a new extension mechanism here we are simply adding the LSP C Editor to the known list as the LSP C Editor is (or will soon be) part of CDT itself anyway (see #354)

By itself this change doesn't do anything, it needs the change in https://github.com/Bachmann-electronic-GmbH/eclipse-cdt-lsp/pull/46/

Also-by: Gesa HENTSCHKE <Gesa.HENTSCHKE@bachmann.info>


jonahgraham:
Pushed to wrong origin. Replaced with #384

PR #386, opened 2023-04-27
[#385] Employ Eclipse Dash license check

ruspl-afed:
Add yml configuration to check licenses

Fixes #385

PR #387, opened 2023-05-01
Update to using direct from maven dependencies where possible

jonahgraham:
This is the next step in modernization and moving on from Orbit.

All the bundles that have fully OSGi bundles in maven are included in this update.

This PR contains the other update needed for 2023-06 M2

Part of #320

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
 			 runs with. It may work with older versions, but this is not
 			 tested or supported. -->
 		<required-maven-version>3.6.3</required-maven-version>
-		<tycho-version>3.0.3</tycho-version>
+		<tycho-version>3.0.5</tycho-version>
 		<cbi-plugins.version>1.3.4</cbi-plugins.version>
 		<sonar.core.codeCoveragePlugin>jacoco</sonar.core.codeCoveragePlugin>
 		<cdt-site>http://ci.eclipse.org/cdt/job/cdt-master/lastSuccessfulBuild/artifact/releng/org.eclipse.cdt.repo/target/repository</cdt-site>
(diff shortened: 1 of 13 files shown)


ruspl-afed:
Thank you for working on this @jonahgraham 
Do you plan to change them all in one PR?

jonahgraham:
> Do you plan to change them all in one PR?

Yes. I need to figure out how to document the updated steps now - but we have always updated all the dependencies in one PR, it just used to be by changing the Orbit version. As most of the versions in the target file were `"0.0.0"` the dependency version would change automagically.

jonahgraham:
I don't understand why license check is failing here for icu, it passes in platform for AFAICT the same content.

jonahgraham:
I don't see how it is passing for Eclipse Platform (e.g.  https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/actions/runs/4846605691/jobs/8636246851) - because I don't see license vetting having been done for 73.1 in iplab. 

The 72.1 is there in https://gitlab.eclipse.org/eclipsefdn/emo-team/iplab/-/issues/4354

We aren't releasing for a couple of months still, so this issue on its own isn't going to hold up merging this. But I would like to understand the flow of what is going wrong with this new check added in #386. Until now I have run license tool and manually verified the not automatically verified parts (e.g. https://ci.eclipse.org/cdt/job/run-dash-licenses/job/cdt_11_1/1/ for 11.1)

jonahgraham:
Before this can be merged we also need the PGP signing which has a dependency on [EF Helpdesk #3070](https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3070)

pom.xml
```diff
@@ -990,6 +990,22 @@
...
+				<artifactId>tycho-gpg-plugin</artifactId>
+				<version>${tycho-version}</version>
+				<executions>
+					<execution>
+						<id>pgpsigner</id>
+						<goals>
+						<goal>sign-p2-artifacts</goal>
+						</goals>
+						<configuration>
+							<keyname>4F23165B6AC51B15</keyname>
```
ruspl-afed, line 1004: Will this magic work for other Eclipse projects as well or this key is individual for every project @jonahgraham ?

jonahgraham: It is specific to CDT's key. See https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/3070

jonahgraham:
Current state of this branch (maven-deps) is that I cannot sign because of `gpg: signing failed: No pinentry` error which according to https://wiki.eclipse.org/Jenkins#Common_issues_with_GPG indicates pinentry is not installed. But as pinentry is installed it is probably a custom image problem.

jonahgraham:
> custom image problem.

So I am working on refreshing our custom images, but for 2023-06 I won't get this integrated.

PR #393, opened 2023-05-12
Cleanup GCCToolChain

ewaterlander:
Fixed some null pointer issues.


build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/GCCToolChain.java
```diff
@@ -651,11 +659,21 @@ private Set<String> getResourcesFileExtensions() {
...
+				fileExts.addAll(Arrays.asList(contentTypeCpp.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));
 
-		IContentType contentTypeC = manager.getContentType(CCorePlugin.CONTENT_TYPE_CSOURCE);
-		fileExts.addAll(Arrays.asList(contentTypeC.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));
+			IContentType contentTypeC = manager.getContentType(CCorePlugin.CONTENT_TYPE_CSOURCE);
+			if (contentTypeC != null)
+				fileExts.addAll(Arrays.asList(contentTypeC.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));
+		}
+
+		if (fileExts.size() == 0) {
```
Torbjorn-Svensson, line 672:
```suggestion
		if (fileExts.isEmpty()) {
```

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/GCCToolChain.java
```diff
@@ -651,11 +659,21 @@ private Set<String> getResourcesFileExtensions() {
 
 		Set<String> fileExts = new TreeSet<String>(String.CASE_INSENSITIVE_ORDER);
 
-		IContentType contentTypeCpp = manager.getContentType(CCorePlugin.CONTENT_TYPE_CXXSOURCE);
-		fileExts.addAll(Arrays.asList(contentTypeCpp.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));
+		if (manager != null) {
+			IContentType contentTypeCpp = manager.getContentType(CCorePlugin.CONTENT_TYPE_CXXSOURCE);
+			if (contentTypeCpp != null)
+				fileExts.addAll(Arrays.asList(contentTypeCpp.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));
```
Torbjorn-Svensson, line 665:
```suggestion
			if (contentTypeCpp != null) {
				fileExts.addAll(Arrays.asList(contentTypeCpp.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));
			}
```

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/GCCToolChain.java
```diff
@@ -651,11 +659,21 @@ private Set<String> getResourcesFileExtensions() {
...
 
-		IContentType contentTypeC = manager.getContentType(CCorePlugin.CONTENT_TYPE_CSOURCE);
-		fileExts.addAll(Arrays.asList(contentTypeC.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));
+			IContentType contentTypeC = manager.getContentType(CCorePlugin.CONTENT_TYPE_CSOURCE);
+			if (contentTypeC != null)
+				fileExts.addAll(Arrays.asList(contentTypeC.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));
```
Torbjorn-Svensson, line 669:
```suggestion
			if (contentTypeC != null) {
				fileExts.addAll(Arrays.asList(contentTypeC.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));
			}
```

ewaterlander:
> Please extend the commit message with when the problem happens and how it can be verified. The current commit message is way to generic to be useful. Based on the info given so far in this PR, I can't say if this is a good change or bad one.

These were possible null pointer uses, detected by static code analysis. I updated the commit message.

ewaterlander:
Test failed due to some package that could not be fetched. Is there a way to re-run tests without a new push?

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/GCCToolChain.java
```diff
@@ -649,13 +658,25 @@ public String[] getCompileCommands(ILanguage language) {
...
+			}
+		}
 
-		IContentType contentTypeC = manager.getContentType(CCorePlugin.CONTENT_TYPE_CSOURCE);
-		fileExts.addAll(Arrays.asList(contentTypeC.getFileSpecs(IContentType.FILE_EXTENSION_SPEC)));
+		if (fileExts.isEmpty()) {
+			Activator.log(new Status(Status.ERROR, Activator.getId(),
+					"Error: cannot determine resources file extensions.\n"));
+			fileExts.addAll(Arrays.asList(new String[] { "C", "c", "c++", "cc", "cpp", "cxx" })); //$NON-NLS-1$ //$NON-NLS-2$ //$NON-NLS-3$ //$NON-NLS-4$ //$NON-NLS-5$ //$NON-NLS-6$
```
Torbjorn-Svensson, line 678: I'm a bit torn here. I don't know if we should hard code the file extensions like this or if we should return the empty set or perhaps throw some kind of exception instead.
@jonahgraham: What's your opinion here?

ewaterlander: The set is filled at the creation of GCCToolChain. An empty set disables the C/C++ indexing, because no resources will be found in the command lines. An exception will stop the creation of GCCToolChain which will disrupt the whole Core Build Makefile flow.

jonahgraham: > @jonahgraham: What's your opinion here?

I don't understand the use case that leads to fileExts being empty at this point in the code. Is this a theoretical problem or is there a use case? If it is theoretical I would probably do it as @ewaterlander has it now as it seems a fairly safe fallback way to fail.

ewaterlander: > I don't understand the use case that leads to fileExts being empty at this point in the code. Is this a theoretical problem or is there a use case? If it is theoretical I would probably do it as @ewaterlander has it now as it seems a fairly safe fallback way to fail.

I'm building a plugin based on Core Build Makefile projects. I'm running static code analysis on my own files and a few CDT files that I needed to modify. I'm currently stuck on CDT 10.5. I created the fallback to get rid of a null pointer issue, because code has to be clean. So you can say it is theoretical. When I move to CDT 11.2 which includes all my other changes I don't need the modified GCCToolChain anymore. But due to constraints I can't move to CDT 11.x for the time being. I offered this change as a code cleanup. I do not depend on it. So you can also choose to remove the fallback. Or I can do it if you want.

jonahgraham (changes requested):

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/GCCToolChain.java
```diff
@@ -535,7 +542,9 @@ public void run() {
 		} catch (InterruptedException e) {
 			Activator.log(e);
 		}
-		Files.delete(tmpFile);
+		if (tmpFile != null) {
+			Files.delete(tmpFile);
+		}
```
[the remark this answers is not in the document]
ewaterlander: Thanks for the review.

jonahgraham (approved):

jonahgraham:
As @Torbjorn-Svensson has an outstanding "change requested" marked on here I will wait a bit before merging to give him a chance to weigh in.

FWIW this is what I see in the UI, because we have allow override for branch protection I can submit this, but I have to tick the big red checkmark first:

image

Torbjorn-Svensson (approved):

jonahgraham:
11.3 — I'm preparing branches/etc now ahead of final build on Monday

PR #399, opened 2023-05-19
Some old codon markers are left after successful indexing

umairsair:


codan/org.eclipse.cdt.codan.core/src/org/eclipse/cdt/codan/internal/core/model/CodanMarkerProblemReporter.java
```diff
@@ -139,17 +136,18 @@ protected Collection<IMarker> findResourceMarkers(IResource resource, IChecker c
...
-		for (IMarker m : markers) {
-			String id = m.getAttribute(ICodanProblemMarker.ID, ""); //$NON-NLS-1$
-			for (IProblem problem : problems) {
-				if (problem.getId().equals(id)) {
-					res.add(m);
-				}
-			}
-		}
-		return res;
+		return Arrays.asList(markers);
```
umairsair, line 139: @jonahgraham , others!

I need codon expert opinion.. sometimes I face issue that some error markers are left after project creation.. if I change code as following, I can see that these are the markers for which the problem doesn't exist in `reg.getRefProblems(checker)`; which means that these markers doesn't actually belong to any problem and should ideally be deleted but this code simply neglects them.. I am trying to understand why are doing filtering here? why don't we return all codon errors for the resource and caller can take of remaining; keep or delete it..

```
		for (IMarker m : markers) {
			String id = m.getAttribute(ICodanProblemMarker.ID, ""); //$NON-NLS-1$
			for (IProblem problem : problems) {
				if (problem.getId().equals(id)) {
					res.add(m);
				}
			}
			if (!res.contains(m))
				System.out.println(m); // It prints exactly those markers which I see after project creation..
		}
		System.out.println("*****************");
```

jonahgraham:
@elaskavaia Can you look at this - your expertise is very welcome in this area.

codan/org.eclipse.cdt.codan.core/src/org/eclipse/cdt/codan/internal/core/model/CodanMarkerProblemReporter.java
ddscharfe: > I need codon expert opinion..

Disclaimer: I'm not an expert on this

> which means that these markers doesn't actually belong to any problem and should ideally be deleted but this code simply neglects them.. I am trying to understand why are we doing filtering here? why don't we return all codon errors for the resource and caller can take care of remaining; keep or delete it..

As far as I understand the implementation, a `CodanMarkerProblemReporter` is responsible for one checker. The implementation you propose would lead to deleting the problem markers for all checkers, when `AbstractChecker::before(IResource resource)` is called. I would guess this will lead to other problems, besides fixing the bug you found.

You already mentioned, you only *sometimes* face this issue, but can you try to provide a way to reproduce the behavior you are experiencing?

umairsair: thanks @ddscharfe , you are right! I understand it now..

I have looked into the problem further.. I noticed that when I face the issue, the indexer reports more problems, so apparently the markers are correct.. but the interesting thing is that if I open the declaration, it opens fine.. so now I need to investigate why indexer randomly starts reporting more problems..

```

!MESSAGE Indexed 'p1' (9 sources, 12 headers) in 0.197 sec: 505 declarations; 1,130 references; 33 unresolved inclusions; 74 syntax errors; 176 unresolved names (9.7%)
!MESSAGE Indexed 'p2' (9 sources, 445 headers) in 2.53 sec: 18,766 declarations; 37,834 references; 1 unresolved inclusions; 51 syntax errors; 202 unresolved names (0.36%)
```

jonahgraham: Thanks @umairsair for providing that useful feedback!

Issue #402, opened 2023-05-24
Rename project does change include directories pointing to this project


jantje:
**Describe the bug**
When I rename a managed build project that has a include directory that points somewhere inside the project the include directory does not get updated.

**To Reproduce**
1. Create a managed Build C++ executable project
2. Create the folder include in the source folder
3. Add the src/include folder to the include directories
4. Rename the project
5. The include directories is not updated

**Expected behavior**
The include directories should point to the renamed project not to the old project name.
Note that the actual project folder was renamed so the "include directory" should be updated.
When I try to add the folder again the correct name is provided.

**Screenshots**
![afbeelding](https://github.com/eclipse-cdt/cdt/assets/1175841/78b234e4-7137-4e81-90d9-a03f4682cf64)

**Desktop (please complete the following information):**
 - OS: windows

**Additional context**
I was investigating a Sloeber rename issue where I found that the rename failed.
So I thought this must be a CDT issue. 
Managed make did fail but did not rename the include directories.

https://github.com/Sloeber/arduino-eclipse-plugin/issues/1567

jantje:
Update:
The sloeber project rename fails due to a linked folder.
Which I could not yet reproduce in naked managed Build.
So the Sloeber fail is probably unrelated to the include directories not being updated.

jonahgraham:
I cannot reproduce the behaviour you see, but I think the difference is in:

> Add the src/include folder to the include directories

I am not sure what you mean by that exactly. This is how I do it:

![Peek 2023-05-29 12-49](https://github.com/eclipse-cdt/cdt/assets/679236/36d65387-1765-42cf-8dac-ad5cff7d7c32)

which leads to the project name *not* being hardcoded in the settings anywhere and the result in the C/C++ General -> Paths and Symbols being the `${ProjName}`:

![image](https://github.com/eclipse-cdt/cdt/assets/679236/1c53c11a-867c-48ed-90e1-9ad1c73ecf2c)

jantje:
I'm adding the includes in Paths and folders in what I assume is the CDT way not the MBS way.
Both ways seems to result in having the includes at both places.
![afbeelding](https://github.com/eclipse-cdt/cdt/assets/1175841/94bf86e1-9452-4c68-aeec-5ae2b44fe5e8)
resulting in a references that does contain the name of the project.
[...]

PR #405, opened 2023-05-31
Use ${ProjName} for workspace includes referencing folders from the project to better support project renames #402

jantje:
I don't really like this solution but it is the best I could come up with.
If there are better ways please tell me how.

### DIFF
--- a/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java
+++ b/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java
@@ -585,9 +585,9 @@ public abstract class AbstractCPropertyTab implements ICPropertyTab {
 		if (dialog.open() == Window.OK) {
 			IResource resource = (IResource) dialog.getFirstResult();
 			if (resource != null) {
-				if (resource.getProject() == prj) {
+				if (resource.getProject().equals(prj)) {
 					StringBuilder buf = new StringBuilder();
-					return buf.append("${").append("workspace_loc:${ProjName}/") //$NON-NLS-1$//$NON-NLS-2$
+					return buf.append("${").append("workspace_loc:/${ProjName}/") //$NON-NLS-1$//$NON-NLS-2$
 							.append(resource.getProjectRelativePath()).append("}").toString(); //$NON-NLS-1$
 				}
 				StringBuilder buf = new StringBuilder();
--- a/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java
+++ b/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java
@@ -496,28 +496,37 @@ public abstract class AbstractCPropertyTab implements ICPropertyTab {
 	}
 
 	public static String getWorkspaceDirDialog(Shell shell, String text) {
-		return getWorkspaceDialog(shell, text, true, null);
+		return getWorkspaceDialog(shell, text, true, true, null);
 	}
 
 	public static String getWorkspaceFileDialog(Shell shell, String text) {
-		return getWorkspaceDialog(shell, text, false, null);
+		return getWorkspaceDialog(shell, text, false, true, null);
+	}
+
+	public static String getWorkspaceDirDialog(Shell shell, String text, IProject prj) {
+		return getWorkspaceDialog(shell, text, true, true, prj);
+	}
+
+	public static String getWorkspaceFileDialog(Shell shell, String text, IProject prj) {
+		return getWorkspaceDialog(shell, text, false, true, prj);
 	}
 
 	/**
 	 * @since 5.4
 	 */
 	public static String getProjectDirDialog(Shell shell, String text, IProject prj) {
-		return getWorkspaceDialog(shell, text, true, prj);
+		return getWorkspaceDialog(shell, text, true, false, prj);
 	}
 
 	/**
 	 * @since 5.4
 	 */
 	public static String getProjectFileDialog(Shell shell, String text, IProject prj) {
-		return getWorkspaceDialog(shell, text, false, prj);
+		return getWorkspaceDialog(shell, text, false, false, prj);
 	}
 
-	private static String getWorkspaceDialog(Shell shell, String text, boolean dir, IProject prj) {
+	private static String getWorkspaceDialog(Shell shell, String text, boolean dir, boolean selectFromRoot,
+			IProject prj) {
 		String currentPathText;
 		IPath path;
 		currentPathText = text;
@@ -528,7 +537,7 @@ public abstract class AbstractCPropertyTab implements ICPropertyTab {
 		ElementTreeSelectionDialog dialog = new ElementTreeSelectionDialog(shell, new WorkbenchLabelProvider(),
 				new WorkbenchContentProvider());
 
-		if (prj == null)
+		if (prj == null || selectFromRoot)
 			dialog.setInput(ResourcesPlugin.getWorkspace().getRoot());
 		else
 			dialog.setInput(prj);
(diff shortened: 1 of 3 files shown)


jonahgraham (changes requested):
This code has a bug if the user selects a project that starts with the same name as the current project you get an error. e.g. create two projects, `simple` and `simpler`. In `simple` press Add and then choose a directory in `simpler`, the resulting value you get is: `/${ProjName}r/src/include` instead of the expected `/simpler/src/include`

I think messing around with the string at this point is too late, instead I think the code that creates the string in the first place could be modified:

https://github.com/eclipse-cdt/cdt/blob/7c8bb9f00ef0d5e3ff493bc0c944519a3e476da2/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java#L576-L582

using the same algorithm as discussed in https://github.com/eclipse-cdt/cdt/issues/402#issuecomment-1567442621

If that proves impossible/impractical, the current solution looks fine to me, but does it need some null checks around get the project name from `cfgd`?

Please also provide a self contained commit message / PR title. If you add to the body of the commit message and PR something like `Fixes #402` then merging the PR will close the issue with the appropriate cross reference. See [help for more info on that.](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

jantje:
>I think messing around with the string at this point is too late, instead I think the code that creates the string in the first place could be modified:

I fully agree. That is why I stated : "I don't really like this solution but it is the best I could come up with."

I tried to modify [cdt/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java](https://github.com/eclipse-cdt/cdt/blob/7c8bb9f00ef0d5e3ff493bc0c944519a3e476da2/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java#L576-L582)
But the provided project is null.
So I tried getResDesc to get the project but the method is static so getResDesc is not available.
So I changed the method to not be static but it is called by static methods.
I didn't feel confident enough to remove the static from all required methods.

So I opted for a quick and dirty method.

If you are ok with removing the statics I'll do it.

jonahgraham:
If it were me I would pass in the project as an additional parameter and leave it static.

jantje:
A bit confusing as it already gets a IProject as a parameter
https://github.com/eclipse-cdt/cdt/blob/7c8bb9f00ef0d5e3ff493bc0c944519a3e476da2/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java#L520
And a explicit null is provided and a check for null is done.
https://github.com/eclipse-cdt/cdt/blob/7c8bb9f00ef0d5e3ff493bc0c944519a3e476da2/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java#L531

jonahgraham:
Yeah, bit annoying to have two iproject in one signature. But they have distinctly different purposes at least.

Or add another boolean parameter like "limit selection to project" and always pass the project in. 

The other alternative is to add a new button called project and that would limit selection to the project.

Or fix the bug in your initial implementation.

jantje:
>Or add another boolean parameter like "limit selection to project" and always pass the project in.

I'll try this one

jantje:
This seems to fail as well as the project folder is added when getting back to the properties pages.
Suppose the project is called "C" and the folder "folder" You now get ${ProjName}/folder but then clicking OK again you get C/${ProjName}/folder
I tried to find where the project got added but I failed.

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java
```diff
@@ -576,8 +585,14 @@ public IStatus validate(Object[] selection) {
 		if (dialog.open() == Window.OK) {
 			IResource resource = (IResource) dialog.getFirstResult();
 			if (resource != null) {
+				if (resource.getProject() == prj) {
```
jonahgraham, line 588: I think this should use .equals? ref: https://github.com/eclipse-cdt/cdt/blob/7c8bb9f00ef0d5e3ff493bc0c944519a3e476da2/core/org.eclipse.cdt.ui/utils.ui/org/eclipse/cdt/utils/ui/controls/FileListControl.java#L325

```suggestion
				if (resource.getProject().equals(prj)) {
```

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java
```diff
@@ -496,28 +496,37 @@ public static String getVariableDialog(Shell shell, ICConfigurationDescription c
...
-		return getWorkspaceDialog(shell, text, true, null);
+		return getWorkspaceDialog(shell, text, true, true, null);
 	}
 
 	public static String getWorkspaceFileDialog(Shell shell, String text) {
-		return getWorkspaceDialog(shell, text, false, null);
+		return getWorkspaceDialog(shell, text, false, true, null);
+	}
+
+	public static String getWorkspaceDirDialog(Shell shell, String text, IProject prj) {
```
jonahgraham, line 506: Needs `@since` - but we can resolve that later in the review process

jonahgraham (changes requested):
First pass review before I try to run it.

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java
jantje: Which version number is to be used?

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java
```diff
@@ -576,8 +585,14 @@ public IStatus validate(Object[] selection) {
 		if (dialog.open() == Window.OK) {
 			IResource resource = (IResource) dialog.getFirstResult();
 			if (resource != null) {
+				if (resource.getProject() == prj) {
+					StringBuilder buf = new StringBuilder();
+					return buf.append("${").append("workspace_loc:${ProjName}/") //$NON-NLS-1$//$NON-NLS-2$
```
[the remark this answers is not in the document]
jantje: That seems to do the trick though I would expect it to be relative path as it is a workspace location after all 😕

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java
jantje: I wasn't aware of the equals. Thanks.

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java
jonahgraham: The minor version in MANIFEST.MF needs to be increased and use that number. The auto-fix in Eclipse should do that if you have API baseline setup. If you don't have it setup I think it will be `8.1`

I can handle that as part of the merging if it isn't straightforward for you.

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/newui/AbstractCPropertyTab.java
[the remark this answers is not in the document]
jonahgraham: I think the "workspace location" defines what the root is, i.e `/` is the workspace root as opposed to the file system root.

jonahgraham (changes requested):
The code change looks good - can you please address this:

> Please also provide a self contained commit message / PR title. If you add to the body of the commit message and PR something like Fixes #402 then merging the PR will close the issue with the appropriate cross reference. See [help for more info on that.](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

jantje:
1) Fixes #402 
I do not really want to say this fixes 402 as it only makes the problem less annoying.
Basically a project rename listener(?) should handle includes which use the project name.
2)  MANIFEST.MF
I remember having had these API incompatibility errors and quick fixes to fix them. But now no errors appear. That may be because I had to clone at GIT level. Anyway I would prefer you to do so as I do not know whether all manifests need updating or only the touched project.

jonahgraham:
See https://github.com/eclipse-cdt/cdt/blob/main/POLICY.md#Using-API-Tooling for instructions on setting up and using API tooling

jonahgraham:
I'm fixing API errors, doing last check and merging this now.

jonahgraham:
I found a bug related to how the new code was handling slashes if the project was the selection as opposed to a folder in a project. In commit bde076107a6cacb7b9e8ce31b5a7b8365567cfa0 I fixed the handling to use Path class to assemble the path as the reference code did that I mentioned in https://github.com/eclipse-cdt/cdt/issues/402#issuecomment-1567442621

jonahgraham (approved):

jonahgraham:
> I do not really want to say this fixes 402 as it only makes the problem less annoying.

I understand what you mean - for now it is marked as fixed, at least fixed in the same way as other places using the same methodology.

PR #417, opened 2023-06-12
Provide hex dump of binary files in Default Binary File Editor

### LINKED ISSUE #416: Support hex dump of binary files in the Default Binary File Editor
The CDT _Default Binary File Editor_ will invoke `objdump` on binary files in cases where the associated build configuration uses a binary parser that adapts to a GNU tool factory. Otherwise, it defaults to presenting the binary file as raw text. It would be preferable to present a hex dump of the binary file in cases where a GNU tool factory is unavailable:
```
00000000 7F 45 4C 46 01 01 01 00 00 00 00 00 00 00 00 00 .ELF............
00000010 02 00 28 00 01 00 00 00 6C 81 00 00 34 00 00 00 ..(.....l...4...
00000020 64 7A 01 00 00 02 00 05 34 00 20 00 03 00 28 00 dz......4. ...(.
00000030 1A 00 19 00 01 00 00 70 7C B6 00 00 7C B6 00 00 .......p|...|...
```

jld01:
Resolves #416

### DIFF
--- a/core/org.eclipse.cdt.ui/META-INF/MANIFEST.MF
+++ b/core/org.eclipse.cdt.ui/META-INF/MANIFEST.MF
@@ -102,6 +102,7 @@ Export-Package: org.eclipse.cdt.internal.corext;x-internal:=true,
  org.eclipse.cdt.ui.wizards,
  org.eclipse.cdt.ui.wizards.conversion,
  org.eclipse.cdt.utils.ui.controls
+Import-Package: org.apache.commons.io;version="[2.8.0,3.0.0)"
 Require-Bundle: org.eclipse.cdt.core;bundle-version="[8.0.0,9.0.0)",
  org.eclipse.compare;bundle-version="[3.8.500,4.0.0)",
  org.eclipse.core.expressions;bundle-version="[3.8.200,4.0.0)",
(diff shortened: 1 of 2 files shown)


core/org.eclipse.cdt.ui/META-INF/MANIFEST.MF
```diff
@@ -125,7 +125,8 @@ Require-Bundle: org.eclipse.cdt.core;bundle-version="[8.0.0,9.0.0)",
  org.eclipse.ui.views;bundle-version="[3.11.300,4.0.0)",
  org.eclipse.ui.workbench.texteditor;bundle-version="[3.16.600,4.0.0)",
  org.eclipse.e4.ui.css.swt.theme,
- org.eclipse.tools.templates.ui;bundle-version="[1.4.0,2.0.0)"
+ org.eclipse.tools.templates.ui;bundle-version="[1.4.0,2.0.0)",
+ org.apache.commons.io;bundle-version="[2.8.0,3.0.0)"
```
jld01, line 129: @jonahgraham is this use of maximum version correct for non-Eclipse bundles?

ruspl-afed, line 129: Please prefer `Import-Package` for 3rd party libs, it simplifies releng work for SimRel

jonahgraham: Yes - org.apache.commons.io tends to follow semantic versioning

Issue #394, opened 2023-05-15
Add additional build artifacts, such as object files in .o format


baoruiqi:
Currently, we can select the type when creating a managed c/c++ project, such as executable files, or modify the build artifact of a project on the settings tab of the property page. However, there seems no way to get an unlinked object file by modifying these settings. For example, I can later disassemble this object file.

The following is a screenshot showing the project properties page. May it be possible to add some options to these entry in build artifact tab and implement the specific build process in the background?

![image](https://github.com/eclipse-cdt/cdt/assets/128341204/cb2b5a8f-f168-4cd4-8d7d-844d464ff64c)

Does such feature exist already and just not work since I'm not familiar with cdt. If not I hope to implement this feature myself.
By searching the source code, I seem to find some content related to the type of obejct file, but due to lack of familiarity with the source code, I cannot fully discover the connection and realize the feature I need. please give me some suggestions. Thanks!

![image](https://github.com/eclipse-cdt/cdt/assets/128341204/604cf0ad-059c-4021-81e4-1bf855b23a23)

jantje:
My 2 cent
As you noticed the project type selected during project creation is linked to build artifact->artifact type project property.
You can find these in the plugin.xml files for the extension point org.eclipse.cdt.managedbuilder.core.buildProperties the key org.eclipse.cdt.build.core.buildArtefactType

The exe artifact type is identified by the key org.eclipse.cdt.build.core.buildArtefactType.exe
The static lib artifact type is identified by the key org.eclipse.cdt.build.core.buildArtefactType.staticLib
The dynamic lib type is identified by the key org.eclipse.cdt.build.core.buildArtefactType.sharedLib
![afbeelding](https://github.com/eclipse-cdt/cdt/assets/1175841/cdffb7db-11e1-4238-a6b5-04fe3a50696f)
As you can see there is no "unlinked object file" there

These keys are then used in the project type definitions as you can see below
![afbeelding](https://github.com/eclipse-cdt/cdt/assets/1175841/87efbd27-50a5-471d-914e-c019733fe61b)

So if you would want to add a "unlinked object file" target the first thing to do is to get a basic understanding of how plugin.xml works in eclipse and then a better understanding of the extension points defined in managed build

This is no small feat and before I would start to tackle this I would make sure the existing CDT can not deliver.
[...]

PR #418, opened 2023-06-13
Implement the object file buid artifact mentioned in #394

baoruiqi:
Hi, all

As mentioned in #394 , an entry is left for the object file just with a name, but the specific implementation of such build artifact of is not given.

Although, I don't think it's a necessary feature, but it might serve some purpose. For example, compiling only can save some link time while checking for errors, and sometimes users may want to use object files to link external libraries.

My idea is rather simple. 
First, create the connection between property name and buildArtefactType by adding an entry to the configuration file of the managedbuild.core module,
Then, adapt the project property page to support the new buildArtefactType,
Finally, in GnuMakefileGenerator, skip the subsequent generation steps if buildArtefactType is judged to be an object file.

Since I don't have a deep understanding of cdt, this is a relatively simple solution I can think of so far. If there is a better implementation solution, please discuss and improve it together.

Thank you!

### DIFF
--- a/build/org.eclipse.cdt.managedbuilder.core/plugin.xml
+++ b/build/org.eclipse.cdt.managedbuilder.core/plugin.xml
@@ -296,6 +296,10 @@
             property="org.eclipse.cdt.build.core.buildArtefactType" 
             id="org.eclipse.cdt.build.core.buildArtefactType.sharedLib" 
             name="%BuildProperty.type.name.sharedLibrary"/>
+      <propertyValue
+            property="org.eclipse.cdt.build.core.buildArtefactType"
+            id="org.eclipse.cdt.build.core.buildArtefactType.objectFile"
+            name="%BuildProperty.type.name.objectFile"/>
 
       </extension>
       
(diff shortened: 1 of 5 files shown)


jonahgraham:
@baoruiqi - I have seen this PR, and I look forward to reviewing it in the near future. In the meantime if any other of the CDT committers want to review, please go ahead as I have no special knowledge in this area anyway.

jld01 (commented):
@baoruiqi a few thoughts on this:
- The key difference between an object file as a build artefact and the existing build artefacts is that the `ITool` that generates an object file does not transform multiple inputs into a single output. For all other tools that generate a build artefact, we can find `<inputType multipleOfType="true">` in the declaring XML and there is a single build artefact file.
- If we are to accommodate an object file as a build artefact, we would need a generic mechanism within the managed build system (MBS) that would trigger the building of all these artefacts. I note that `IConfiguration.getArtifactName()` and `IConfiguration.getArtifactExtension()` assume that the artefact is a single file. It might be possible for the `IConfiguration` implementation to return a virtual identifier rather than a file name (similar to a _phony_ target in a GNU makefile), but I have not investigated this.
- Finally, a solution would need to work for the CDT _Internal Builder_ as well as the _External Builder_. In the case of the _Internal Builder_, this implies some work in determining which of the multiple build artefacts are stale and require rebuilding.

jonahgraham:
Thanks @jld01 for reviewing. @baoruiqi I can try to provide any additional info you may need, but I hope that @jld01's review is enough to get you started.

baoruiqi:
@jonahgraham @jld01 Thank you for your suggestions. I was busy at work a while ago, and I will try to realize this feature in a more seamless manner if possible with cdt when I have time recently.

jonahgraham:
I'm going to move this into draft until you are ready to return to this.

jonahgraham:
I am cleaning up our backlog of open Pull Requests. Please re-open this PR if you plan to pick up this work again soon.

PR #448, opened 2023-07-07
Bug 559318 - construct via initialisation list with single item

davmac314:
Add the special case handling from [dcl.list.init] so that a class object can be constructed using an initialisation list with a single item of the same type (or subclass type).

This is not a complete fix for all cases of initialisation lists (in particular string literals aren't handled) but it should help for the case identified in the bug.

### DIFF
--- a/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/core/parser/tests/ast2/AST2CPPTests.java
+++ b/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/core/parser/tests/ast2/AST2CPPTests.java
@@ -8848,7 +8848,7 @@ public class AST2CPPTests extends AST2CPPTestBase {
 	//		h({ "foo" }); // OK: h(C(std::string("foo")))
 	//		i({ { 1, 2 }, { "bar" } }); // OK: i(D(A(std::initializer_list<int>{1,2}),C(std::string("bar"))))
 	//      X x1;
-	//      x({x1});  // no matching constructor
+	//      x({x1});  // calls copy constructor
 	//	}
 	public void testListInitialization_302412c() throws Exception {
 		String code = getAboveComment();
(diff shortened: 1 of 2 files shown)


davmac314:
The test failures don't seem to have anything to do with the change, which is relatively small :(

davmac314:
Thanks @jonahgraham . I have a number of other fixes lined up as well, but I will hold off on creating PRs for those until these have gone through, to keep it easier to manage (also for myself).

davmac314:
Hi @jonahgraham - I have rebased on current main. There is still a test failure, but it still looks unrelated to my changes (though different from last time).

i-garrison:
I think I have similar fix but made differently: if you drop constructor special case from here https://github.com/eclipse-cdt/cdt/pull/448/files#diff-43fc87d9b8201b05f77f7b8f2edcb7b04a31d029dc1011ebe2f9c22385797c77L617-L632 and then restrict cost to NO_CONVERSION if checkImplicitConversionSequence indicates narrowing conversion but context demands no narrowing here https://github.com/eclipse-cdt/cdt/pull/448/files#diff-43fc87d9b8201b05f77f7b8f2edcb7b04a31d029dc1011ebe2f9c22385797c77L701 and filter out other constructors in list of potentialCosts in `resolveFunctionImpl` due to no narrowing conversion requested, then `testListInitialization_302412c` should allow `x({x1})` as initialization from lvalue reference to x1.
Unfortunately I do not have much time to finish my change atm.

i-garrison:
I now added a PR which fixes this particular issue with single list item of same class type, and also handles other places where list-initialization sequence needs more careful processing. Please compare https://github.com/eclipse-cdt/cdt/pull/581 - there are few more tests added too.

davmac314:
I'm happy for this PR to be closed in preference to #581

davmac314:
Closing this in favour of #581

PR #470, opened 2023-07-20
Use latest Platform version so LinuxTools can resolve

jonahgraham:


jonahgraham:
This doesn't build because of issues in 3rd party dependencies - see See https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/issues/232 and instead I'm temporarily going to the last release linuxtools in #471

jonahgraham:
Replaced with #471 and #472

PR #471, opened 2023-07-20
Use last released LinuxTools docker

jonahgraham:
At the moment there are some resolution issues on the nightly versions of LinuxTools docker components, so for now use the latest release rather than nightly.

See https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/issues/232


jonahgraham:
@jld01 this PR should fix the issue of plug-in validation when using cdt as target platform too.

PR #474, opened 2023-07-21
Move entirely from the old Orbit repos to the new aggregated ones

jonahgraham:
I have also removed the now unused dependencies.

### DIFF
--- a/releng/org.eclipse.cdt.target/cdt.target
+++ b/releng/org.eclipse.cdt.target/cdt.target
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <?pde version="3.8"?>
-<target name="cdt" sequenceNumber="142">
+<target name="cdt" sequenceNumber="143">
 	<locations>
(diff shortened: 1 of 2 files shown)


releng/org.eclipse.cdt.target/cdt.target
```diff
@@ -202,6 +192,13 @@
 			     See https://github.com/eclipse-pde/eclipse.pde/issues/675
 			     -->
 			<unit id="com.google.guava.failureaccess" version="0.0.0" />
+
+			<unit id="org.junit" version="4.13.2.v20211018-1956" />
+		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/tools/orbit/simrel/maven-bnd/nightly/latest"/>
+			<!-- may be a temporary problem but https://download.eclipse.org/tools/orbit/simrel/orbit-aggregation/nightly is missing the update version of this bundle -->
```
jonahgraham, line 200: I think once https://github.com/eclipse-orbit/orbit-simrel/pull/3 is resolved then the out of sync of the two orbit repos will be resolved.

PR #371, opened 2023-04-19
Improve encapsulation within AbstractGnuToolPrefixMacro

jld01:
Part of #361

Issue #454, opened 2023-07-11
Test results have recently regressed


jonahgraham:
The last few builds, some before the version updates for #420 and more since has left the test suite looking poor. This issue is to track what is going on here.

jonahgraham:
See #498 for one of the cases.

jonahgraham:
Most of the test regressions come back to #422

PR #483, opened 2023-08-02
Fix version dependency on o.e.c.managedbuilder.core

jld01:
Discussion at #371


jonahgraham (approved):
I kicked off the build again because it failed in a weird way which I suspect was something transient. 

The test failures are covered broadly by https://github.com/eclipse-cdt/cdt/issues/454

jonahgraham:
I rebased the change to pick up #486 which fixes the broken build.

PR #485, opened 2023-08-02
[#484] Add extension point for custom icons of C/C++ source files

### LINKED ISSUE #484: Support custom icons for C/C++ source files
As a CDT User I want to see easily which C/C++ Editor is used by my project (LSP based C/C++ Editor or the old one). This can be achieved by using custom icons for C/C++ files in the Project Explorer view: 
image

A solution would be, that CDT could provide an extension point for ImageDescriptors e.g. `CFileImageProvider`

I can provide a PR for this.

ghentschke:
fixes #484


jonahgraham (commented):
This mostly looks fine - one small problem with it is that it implies that "anyone" can provide such an image descriptor, but the current code will select the first descriptor that returns isEnabled true.

Rather than adding a lot of complications that won't be used (like priorities) and as this change is primarily to prevent CDT core/ui from depending on CDT-LSP, maybe we should mark the API as such - i.e. make it x-friends to only cdt-lsp and add similar comment to the exsd.

PS #486 resolves build failure and it should be merged soon.

jonahgraham:
> make it x-friends 

I think that means putting the new interface in a new package.

jonahgraham (approved):

PR #488, opened 2023-08-03
Update to Tycho 4.0.1

jonahgraham:
The error handling in Tycho 4 identifies when bundles are listed in category.xml but not available. The removed bundles were not in the output, and with this change the error is resolved in the build.

Updated setting for maven enforcer to match Tycho requirements.


akurtakov:
Tych0 4.0.1 is out so it makes sense to go for it directly.

jonahgraham:
> Tych0 4.0.1 is out so it makes sense to go for it directly.

Done.

PR #490, opened 2023-08-03
Resolve cast exception - fixup for inferring generic types

jonahgraham:
Long ago in 334702ee05048a95e131530ddb80fbf2654dfef5 generic types were added. However in at least one place a previous assumption that existed was broken. That assumption was the true type of ResourceInfo.getTools() would be Tool[] as opposed to ITool[]. That commit broke it for ResourceConfiguration.getTools()

Internally in CDT ITool is in practice always Tool, and as a result there are lots of downcast from ITool -> Tool. However ITool[] cannot be downcast to Tool[] unless the original type was Tool[].

Instead we use Arrays.copyOf to copy the array into the correct array type.

Steps to reproduce the problem now fixed:

1. Create a Managed Build executable project.
2. Open the properties window of the source file (as opposed to project)
3. At the Properties page, navigate to "C/C++ Build/Tool Chain Editor".
4. Click the "Apply" button.

### DIFF
--- a/build/org.eclipse.cdt.managedbuilder.core/META-INF/MANIFEST.MF
+++ b/build/org.eclipse.cdt.managedbuilder.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.cdt.managedbuilder.core; singleton:=true
-Bundle-Version: 9.6.0.qualifier
+Bundle-Version: 9.6.100.qualifier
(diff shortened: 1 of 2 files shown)


jonahgraham:
I have an alternative fix for this that I am trying, which is to try to get the generic types around ITool vs Tool correct in the first place so that the casting can be "safer" everywhere.

jonahgraham:
> I have an alternative fix for this that I am trying, which is to try to get the generic types around ITool vs Tool correct in the first place so that the casting can be "safer" everywhere.

Actually I don't - the change to using `Tool` internally everywhere, while I believe it is correct, is way to extensive to bother trying to complete and review.

jld01 (approved):
Looks good to.me.

Issue #502, opened 2023-08-11
Support for creating and opening projects in pre CDT 4.0 format will be removed


jonahgraham:
There is a lot of code in CDT to support old CDT project versions, those before the current `.cproject` file format was finalized at version 4.0.0 (as you can see in `.cproject` files with the `<?fileVersion 4.0.0?>` tag). This format was introduced in 2007.

This is a tracking bug to handle removing this code in a future version. Some of the code is public API so is subject to the [API policy](https://github.com/eclipse-cdt/cdt/blob/main/POLICY.md#deprecating-and-deleting-api) and therefore won't be deleted until either the next major version of CDT, or 2 years from now.

PR #504, opened 2023-08-11
Log errors when processing .cproject files

jonahgraham:
This relates to #502 where we are getting rid of support for all pre-4.0 project structures. In the past if we failed to handle .cproject successfully we would default to old settings.

This change still has the fallback, but the error message is exposed to the Error Log instead of being suppressed.


jonahgraham:
The knock on effect of changing this so many years later is too big as lots of code was relying on this to silently fail, especially when making a new project.

PR #507, opened 2023-08-14
Add support for Context when running Boost test runners

### LINKED ISSUE #459: C/C++Testing: XML parse error when using Boost Test Runners (<Context> is not accepted)
**Describe the bug**
In out project we use BOOST test environment for unit and integrationtests. We have defined a "Run Configuration" in eclipse and choose "Boost Test Runner" as "Testrunner". We use data test cases (BOOST_DATA_TEST_CASE_F) in a unit test. When using the boost test runner in eclipse the following error will be reported in "C/C++ Unit"-section:

**XML parse error: Invalid XML format: Element "Context" is not accepted!**

The following program arguements are set in run configuration of boost test runner: "--output_format=XML --log_level=all --report_level=no --"

**To Reproduce**
1. Use/Write any data set test case in boost
2. Run this testcase with "boost test runner" configuration

**Expected behavior**
The boost data testcase are runable when calling them by standard "C/C++ Application" (.../Verification.exe). We expect that eclipse can handle XML-Element `<Context>`

**Screenshots**
image

 - OS: Windows 
 - Browser chrome
 - Version 

**Smartphone (please complete the following information):**
 - Not relevant

**Additional context**
In the following screenshot from xml file you can see the` <Context>`-Element due to data test case:
image

jonahgraham:
Fixes #459


jonahgraham:
Setting to draft until discussion in #459 is progressed/concluded.

PR #508, opened 2023-08-15
Bug#494: Fix for ability to set terminal tab name programmatically

### LINKED ISSUE #494: Ability to set terminal tab name programmatically and not to reset it with ANSI command
### Discussed in https://github.com/eclipse-cdt/cdt/discussions/492

<div type='discussions-op-text'>

<sup>Originally posted by **mmx85** August  4, 2023</sup>
I have the code that setups terminal and opens it with custom tab name after some user actions. It looks like this:

HashMap<String, Object> params = new HashMap<String, Object>();
params.put(ITerminalsConnectorConstants.PROP_TITLE, "some name"));
params.put(ITerminalsConnectorConstants.PROP_DELEGATE_ID, delegateId);
params.put(ITerminalsConnectorConstants.PROP_PROCESS_PATH, bashShellExecutablePath);
params.put(ITerminalsConnectorConstants.PROP_PROCESS_WORKING_DIR, workingDir);
...
TerminalServiceFactory.getService().openConsole(params, null);

It worked well for me with CDT 10.0.

After updating to CDT 10.2 and higher the code above sets the correct name but VT100Emulator.processNewText() overrides it reading a startup sequence that Bash outputs with ANSI command to change terminal title to current working directory value. When VT100Emulator reads this command it overrides terminal tab title with current working directory value calling: processNewText()->processAnsiOsCommand()->terminal.setTerminalTitle(ansiOsCommand.substring(2)) in VT100Emulator class. 
</div>

This regression is cased by the following changes: https://git.eclipse.org/c/cdt/org.eclipse.cdt.git/commit/?id=fe003208544902646096db658f63d2038a80d40a

I need to have ability to keep the value that is set with ITerminalsConnectorConstants.PROP_TITLE parameter regardless what is set with "\e]0;...\u0007" ANSI command and when but this does not meant that feature that caused regression should be reverted or something like this.

I propose to add a ITerminalsConnectorConstants.PROP_TITLE_UPDATE constant with options:
1. PROP_TITLE_UPDATE_API (to enable title change from menu and with value set to ITerminalsConnectorConstants.PROP_TITLE parameter but disable ability to change title with ANSI command "\e]0;...\u0007") and
2. PROP_TITLE_UPDATE_MIXED (keep current behavior).

mmx85:
setTerminalTitle method sets terminal title unconditionally. When you try to set title programmatically using the following code:
HashMap<String, Object> params = new HashMap<String, Object>();
params.put(ITerminalsConnectorConstants.PROP_TITLE, "some name"));
params.put(ITerminalsConnectorConstants.PROP_DELEGATE_ID, delegateId);
params.put(ITerminalsConnectorConstants.PROP_PROCESS_PATH, bashShellExecutablePath);
params.put(ITerminalsConnectorConstants.PROP_PROCESS_WORKING_DIR, workingDir);
For some terminals like Bash "\e]0;...\u0007" ANSI command is issued when terminal is initialized. It redefines the terminal name that is set by params.put(ITerminalsConnectorConstants.PROP_TITLE, "some name"));.

The proposed fix adds a new parameter "String requestor" for setTerminalTitle method to track originator of the change and a new property flag PROP_TITLE_UPDATE_API. When developer sets "true" for PROP_TITLE_UPDATE_API flag then ANSI command is ignored and terminal title is not updated when user executes "\e]0;<New Title>\u0007" in terminal, but title can be changed from a pop-up menu or programmatically. If developer sets "false" for PROP_TITLE_UPDATE_API flag or does not set anything for the flag (in this case PROP_TITLE_UPDATE_API flag is set to "false" implicitly) then existing behavior is preserved. 

The update allows to add different behavior for different originators in future by updating setTerminalTitle logic to process different requestors with minimal changes in the other places.

Fixes #494


jonahgraham:
Thanks for fixing that other code cleanliness issue :-)

>  have approved to run again.

PS Once your first commit is merged into the repo the workflows will run automatically on future PRs. For now I will just continue to press Approve and Run as soon as I see the new commits.

mmx85:
Thanks @jonahgraham. Now another failure has happened. From console output it is unclear what exactly has happened.
BTW when I ty to setup build environment automatically following https://github.com/eclipse-cdt/cdt/blob/main/CONTRIBUTING.md#setup-cdt-for-development-with-oomph some error happens at the end of the process that prevents successful completion of dev environment setup.

jonahgraham:
I added `Fixes #494` to the body of the message which causes GitHub to auto-link and auto-close the #494 issue once I merge this

I have applied the (probable) target platform fix to this branch too (from #495) so we can get workflows running.

terminal/plugins/org.eclipse.tm.terminal.control/src/org/eclipse/tm/internal/terminal/control/ITerminalListener.java
```diff
@@ -30,6 +30,7 @@ public interface ITerminalListener {
 	/**
 	 * Set the title of the terminal.
 	 * @param title
+	 * @param requestor Who requests update.
 	 */
-	void setTerminalTitle(String title);
```
jonahgraham, line 34: This is somewhat frustrating (and not something I realized until now). 

When this code was converted to API in dea02eb9f98dbdbcea414adf68894e8af04f23f1 it seems that there was a decision not to rename these packages. So even though they have "internal" in their package names, they are API of the bundle:

https://github.com/eclipse-cdt/cdt/blob/b2666eb57819b3e22b22bf05f586caa95dc34df4/terminal/plugins/org.eclipse.tm.terminal.control/META-INF/MANIFEST.MF#L15-L18

That means we need to [evolve this API](https://github.com/eclipse-cdt/cdt/blob/main/POLICY.md#api) in a non-breaking way.

Do you think you can evolve the API? Do you know about the API rules?

terminal/plugins/org.eclipse.tm.terminal.test/src/org/eclipse/tm/internal/terminal/emulator/MockTerminalControlForText.java
```diff
@@ -32,8 +32,10 @@ public void setState(TerminalState state) {
 	}
 
 	@Override
-	public void setTerminalTitle(String title) {
-		allTitles.add(title);
+	public void setTerminalTitle(String title, String requestor) {
+		if (requestor.equals("ANSI")) {
```
jonahgraham, line 36: This constant should be shared so that future readers of the code can track down how these things connect.

Have you considered using an `enum` to handle the requestor?

terminal/plugins/org.eclipse.tm.terminal.view.ui/src/org/eclipse/tm/terminal/view/ui/tabs/TabTerminalListener.java
```diff
@@ -161,10 +163,38 @@ protected String getTerminalConsoleTabTitle(TerminalState state) {
...
+			// Get the original terminal properties associated with the tab item
+			Map<String, Object> properties = (Map<String, Object>) item.getData("properties"); //$NON-NLS-1$
+			if (properties.containsKey(ITerminalsConnectorConstants.PROP_TITLE_UPDATE_API)) {
+				flag = (Boolean) properties.get(ITerminalsConnectorConstants.PROP_TITLE_UPDATE_API);
+			}
+			// Check if terminal should be updated either using API only (flag == true)
+			// or using API and ANSI command (flag == false).
+			if (flag == true && requestor != null && requestor.equals("ANSI")) {
+				return;
+			}
```
jonahgraham, line 193: This looks mostly like the inlined version of `updateTitle`. You can refactor private methods so we don't need to duplicate this code.

jonahgraham (changes requested):
The change itself looks pretty good to me. We are bitten by this code being API and we cannot break public API.

Please let me know how much guidance you would like me to provide on how to handle evolving the API. I can leave you to it, answer questions, or even provide some code. Let me know what you need as I am keen to get this done, ideally by Monday so we can get it into 2023-09 M3 build next week.

terminal/plugins/org.eclipse.tm.terminal.test/src/org/eclipse/tm/internal/terminal/emulator/MockTerminalControlForText.java
mmx85: Agree. What do you think is the best place for this enum?

terminal/plugins/org.eclipse.tm.terminal.control/src/org/eclipse/tm/internal/terminal/control/ITerminalListener.java
mmx85: Did not know about that. For now I do not have better idea how to implement this fix. Need to think more about it. Do you have any ideas?

mmx85: Well, we can try to pass complete ANSI string in https://github.com/eclipse-cdt/cdt/blob/af7643e0d0b7fd723ea1aaf00bde78ea9eeb640d/terminal/plugins/org.eclipse.tm.terminal.control/src/org/eclipse/tm/internal/terminal/emulator/VT100Emulator.java#L489 and analyze it in updateTitle for leading "0;" symbols and then consider it as passed using ANSI command but I do not think it will be a robust solution. Probably setting complete "\e]0;" prefix and passing whole string to setTerminalTitle and further string analysis in updateTitle will be more robust. This will allow us to keep existing implementation of setTerminalTitle method.

jonahgraham: I'll try to throw a prototype together of how to write your code using an evolved API.

jonahgraham: I force pushed with my proposal on how to handle it - there is still some reviewing and updating you need to do, but see https://github.com/eclipse-cdt/cdt/pull/508/commits/18634bb4971fcf78eee467ebf5e173ef023a011b for a summary of what I propose.

terminal/plugins/org.eclipse.tm.terminal.test/src/org/eclipse/tm/internal/terminal/emulator/MockTerminalControlForText.java
jonahgraham: 1. The enum can go right next to the main declaration i.e. an inner enum in org.eclipse.tm.internal.terminal.emulator.VT100TerminalControl.setTerminalTitle(String, String)
2. be clear in the API if null is allowed for the requestor (you have it documented in the implementation class, but that needs to be in the interface because that is the API)
3. All the places that you updated to call the new API don't really seem necessary, e.g. you don't need to update the one calling with "RCM" at all with the evolved API.

jonahgraham:
Great timing on pushing the new code - I am just now looking at terminal issues ahead of M3 build this weekend! Let me know when you are ready for me to review your updates.

jonahgraham:
> We are bitten by this code being API and we cannot break public API.

Since you first created this PR I have added a feature to the build so that the "Code Cleanliness Checks" verifies the API changes are ok in #514

mmx85:
Great that I managed to meet the deadline with new push. I have just fixed the formatting. Please review. I hope I understood your idea for fixing correctly and we are both on the same page.

mmx85:
I do not understand this error:
The following bundles are missing a service segment version bump:
  - org.eclipse.tm.terminal.view.ui
Please bump service segment by [10](https://github.com/eclipse-cdt/cdt/actions/runs/5906303672/job/16022131310?pr=508#step:9:11)0 if on main branch

jonahgraham (approved):
I have reviewed and applied the requested changes as I really want to get this in to 11.3.

Aside from the comments I made I also renamed the new flag as I found the wording confusing as to what it was doing. I hope that the rename isn't too disruptive to you.

I squashed your previous code into one commit (with a more expansive commit message) and added a new commit 7d6039a295  that has all my changes. I will squash that ahead of merging the change.

If you have a chance to look at this before (or even after) I merge that would be great. The build is running now and I will merge it before producing the M3 build assuming no blocking comments.


> I do not understand this error: The following bundles are missing a service segment version bump:
> 
> * org.eclipse.tm.terminal.view.ui
>   Please bump service segment by [10](https://github.com/eclipse-cdt/cdt/actions/runs/5906303672/job/16022131310?pr=508#step:9:11)0 if on main branch

Please see https://wiki.eclipse.org/Version_Numbering#When_to_change_the_service_segment for details. I have made the change.

mmx85:
Thanks @jonahgraham. I'm OK with the changes you did. My initial idea for vars and consts naming was that we may want to disable not only ANSI command to update title but other, but eventually this all narrows to ANSI command only, for now.

jonahgraham:
Ahh. I understand your naming. We can add additional consts as we need more options in the future. But if you have anything else you would like to add now we should try to get it in now., so feel free to let me know.

mmx85:
It would be much easier to work on fixes when #511 is resolved. Please prioritize this issue for resolving soon.

jonahgraham:
#511 is a priority, but I'm held up by my number one priority which is the builds are all failing on Jenkins for the last few days!

mmx85:
> But if you have anything else you would like to add now we should try to get it in now., so feel free to let me know.

No. What we have now is fine.

PR #512, opened 2023-08-16
Update build-test.yml

eling22:


jonahgraham:
Hi @eling22 - in general I have no problem with what you are proposing here and have wondered a couple of times if we should do it as we used to have something similar when we used Jenkins + gerrit to verify new code. It should have a `retention-days` setting to something fairly short (maybe a day?) so that we don't use up all our space.

It would be great if you could enable actions on your fork, test this out and provide a final PR to merge :)

eling22:
Hi @jonahgraham ,
In this pull request, the main goal I want to achieve is to create a file that can be directly installed by the user. However, after downloading `releng/org.eclipse.cdt.repo/target/repository`, I only see the p2 files. I am aware that p2 files are used for updating software from Eclipse IDE. However, is there a way to obtain something like a dmg file for Mac OS and an exe file for Windows ?

jonahgraham:
> However, is there a way to obtain something like a dmg file for Mac OS and an exe file for Windows ?

Not from CDT's build. CDT is currently fed into the SimRel and the [Eclipse Packaging project](https://github.com/eclipse-packaging/packages) builds the downloads on https://www.eclipse.org/downloads/packages/

Issue #515, opened 2023-08-17
Memory Search GUI improvement 


LizzMre:
**Describe the issue**
Selecting a memory address from the memory search results list does not refocus on Memory Browser/ Memory view. 

**To Reproduce**
1.	Open “Memory Browser” or “Memory view”. 
2.	Open the context menu and select “Find and Replace”. 
3.	Add value to the “Find” field and then click on “Find All”. 
4.	“Search” view will be opened in the same tab group as the memory view. 
5.	Click on a memory address.

**Expected behavior**
Focus on Memory Browser/ Memory view and jump to the selected memory address.

**Potential Fix**
Improve selection changed listener implementation: https://github.com/eclipse-cdt/cdt/blob/main/memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java#L172.  

**Additional context**
The current behavior does not include refocusing on the memory views, thus it is not intuitive for the user that selecting an address from the memory search result list has any effect. 
To see the effect, the user must switch to the memory views manually.

LizzMre:
Here you can find my proposed solution: https://github.com/LizzMre/cdt/commit/3c792b18bddcdfe882028b3f6c6abade29bcf329

jonahgraham:
> Here you can find my proposed solution: https://github.com/LizzMre/cdt/commit/3c792b18bddcdfe882028b3f6c6abade29bcf329

Can you make a PR with that change so it can be reviewed?

jonahgraham:
PS Please merge or rebase you change on the main branch  - I just merged some fixes to the build reporting.

PR #516, opened 2023-08-17
Improve memory search results page

LizzMre:
The Memory Browser and Memory views will jump to the memory address after a user selects a value from the Memory Search Result list. If the views belong to the same tab group as the Search view, then it might not be very intuitive for the user to switch back to one of the Memory views and see the effect of the selection. 
The current commit adds on the the double click listener(from MemorySearchResultsPage.java) the behavior to refocus on the Memory Browser/ Memory view.
Resolves: #515

### DIFF
--- a/memory/org.eclipse.cdt.debug.ui.memory.search/META-INF/MANIFEST.MF
+++ b/memory/org.eclipse.cdt.debug.ui.memory.search/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.cdt.debug.ui.memory.search;singleton:=true
-Bundle-Version: 1.5.0.qualifier
+Bundle-Version: 1.5.100.qualifier
 Bundle-Activator: org.eclipse.cdt.debug.ui.memory.search.MemorySearchPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
--- a/memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
+++ b/memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
@@ -110,8 +110,8 @@ public class MemorySearchResultsPage extends Page implements ISearchResultPage, 
 
 	@Override
 	public void setInput(ISearchResult search, Object uiState) {
-		if (search instanceof MemorySearchResult) {
-			((MemorySearchResult) search).addListener(new ISearchResultListener() {
+		if (search instanceof MemorySearchResult result) {
+			result.addListener(new ISearchResultListener() {
 				@Override
 				public void searchResultChanged(SearchResultEvent e) {
 					Display.getDefault().asyncExec(() -> fTreeViewer.refresh());
@@ -176,34 +176,41 @@ public class MemorySearchResultsPage extends Page implements ISearchResultPage, 
 
 			@Override
 			public void selectionChanged(final SelectionChangedEvent event) {
-				if (event.getSelection() instanceof StructuredSelection) {
-					IMemoryRenderingContainer containers[] = ((IMemorySearchQuery) fQuery).getMemoryView()
-							.getMemoryRenderingContainers();
-					MemoryMatch match = (MemoryMatch) ((StructuredSelection) event.getSelection()).getFirstElement();
-					if (match != null) {
-						for (int i = 0; i < containers.length; i++) {
-							IMemoryRendering rendering = containers[i].getActiveRendering();
-							if (rendering instanceof IRepositionableMemoryRendering) {
-								try {
-									((IRepositionableMemoryRendering) rendering).goToAddress(match.getStartAddress());
-								} catch (DebugException e) {
-									MemorySearchPlugin.logError(
-											Messages.getString("MemorySearchResultsPage.RepositioningMemoryViewFailed"), //$NON-NLS-1$
-											e);
-								}
-							}
+				IMemoryRenderingContainer containers[] = ((IMemorySearchQuery) fQuery).getMemoryView()
+						.getMemoryRenderingContainers();
+				if (containers == null || containers.length == 0)
+					return;
 
-							if (rendering != null) {
-								// Temporary, until platform accepts/adds new interface for setting the selection
-								try {
-									Method m = rendering.getClass().getMethod("setSelection", //$NON-NLS-1$
-											new Class[] { BigInteger.class, BigInteger.class });
-									if (m != null)
-										m.invoke(rendering, match.getStartAddress(), match.getEndAddress());
-								} catch (Exception e) {
-									// do nothing
-								}
-							}
+				if (event.getSelection() instanceof StructuredSelection sel
+						&& sel.getFirstElement() instanceof MemoryMatch match) {
+
+					for (int i = 0; i < containers.length; i++) {
+						IMemoryRendering rendering = containers[i].getActiveRendering();
+						if (!(rendering instanceof IRepositionableMemoryRendering repositionable))
+							continue;
+
+						try {
+							repositionable.goToAddress(match.getStartAddress());
+						} catch (DebugException e) {
+							MemorySearchPlugin.logError(
+									Messages.getString("MemorySearchResultsPage.RepositioningMemoryViewFailed"), //$NON-NLS-1$
+									e);
+						}
+
+						// Temporary, until platform accepts/adds new interface for setting the selection
+						try {
+							Method m = rendering.getClass().getMethod("setSelection", //$NON-NLS-1$
+									new Class[] { BigInteger.class, BigInteger.class });
+							if (m != null)
+								m.invoke(rendering, match.getStartAddress(), match.getEndAddress());
+						} catch (NoSuchMethodException e) {
+							// Not all renderings have a compatible setSelection
+							// ideally this would be an interface we can check instead
+							// of using reflection!
+						} catch (Exception e) {
+							MemorySearchPlugin.logError(
+									"Exception while invoking the setSelection method for the current rendering", //$NON-NLS-1$
+									e);
 						}
 					}
 				}
@@ -214,8 +221,8 @@ public class MemorySearchResultsPage extends Page implements ISearchResultPage, 
 			@Override
 			public void doubleClick(DoubleClickEvent event) {
 				// Open the memory view to emphasize the effect of the memory address selection in the search
-				if (event.getSelection() instanceof StructuredSelection
-						&& ((StructuredSelection) event.getSelection()).getFirstElement() instanceof MemoryMatch) {
+				if (event.getSelection() instanceof StructuredSelection sel
+						&& sel.getFirstElement() instanceof MemoryMatch) {
 					IWorkbenchPart wb = ((IMemorySearchQuery) fQuery).getMemoryView().getSite().getPart();
 					if (wb == null)
 						return;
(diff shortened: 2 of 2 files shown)


memory/org.eclipse.cdt.debug.ui.memory.search/META-INF/MANIFEST.MF
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.cdt.debug.ui.memory.search;singleton:=true
-Bundle-Version: 1.5.0.qualifier
+Bundle-Version: 1.5.0.100
```
jonahgraham, line 5: The 100 should be in the 3rd position:

```suggestion
Bundle-Version: 1.5.100.qualifier
```

jonahgraham (commented):
This looks reasonable and I will run it up in my IDE and provide any further feedback.

I can squash this PR when I merge it, or if you make further changes you can feel free to squash it so the PR has just one commit.

memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
```diff
@@ -173,34 +174,48 @@ public Object[] getElements(Object inputElement) {
...
+						// Temporary, until platform accepts/adds new interface for setting the selection
+						try {
+							Method m = rendering.getClass().getMethod("setSelection", //$NON-NLS-1$
+									new Class[] { BigInteger.class, BigInteger.class });
+							if (m != null)
+								m.invoke(rendering, match.getStartAddress(), match.getEndAddress());
+						} catch (Exception e) {
+							MemorySearchPlugin.logError(
+									"Exception while invoking the setSelection method for the current rendering", //$NON-NLS-1$
+									e);
```
jonahgraham, line 218: This is now filling my log with errors everytime I move around the selection. I do agree that logging Exception is a good idea, but there should remain an empty `catch (NoSuchMethodException e1)` so that that expected error isn't logged.

PS I'm really not keen on the reflection use at all here and would love to see that fixed as it clearly isn't "temporary"

```suggestion
						} catch (NoSuchMethodException e) {
							// Not all renderings have a compatible setSelection
							// ideally this would be an interface we can check instead
							// of using reflection!
						} catch (Exception e) {
							MemorySearchPlugin.logError(
									"Exception while invoking the setSelection method for the current rendering", //$NON-NLS-1$
									e);
```

memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
```diff
@@ -173,34 +174,48 @@ public Object[] getElements(Object inputElement) {
...
 			public void selectionChanged(final SelectionChangedEvent event) {
-				if (event.getSelection() instanceof StructuredSelection) {
-					IMemoryRenderingContainer containers[] = ((IMemorySearchQuery) fQuery).getMemoryView()
-							.getMemoryRenderingContainers();
+				IMemoryRenderingContainer containers[] = ((IMemorySearchQuery) fQuery).getMemoryView()
+						.getMemoryRenderingContainers();
+				if (containers == null || containers.length == 0)
+					return;
+
+				if (event.getSelection() instanceof StructuredSelection
```
jonahgraham, line 182: If we're rewriting this if anyway, lets use modern Java to simplify it. (I can't use GitHub suggestions as the change crosses unmodified parts of the code:

These two lines:

```java
				if (event.getSelection() instanceof StructuredSelection sel
						&& sel.getFirstElement() instanceof MemoryMatch match) {
```

replace everything including the `if (match == null) return;`

memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
```diff
@@ -173,34 +174,48 @@ public Object[] getElements(Object inputElement) {
...
+					if (match == null)
+						return;
+
+					for (int i = 0; i < containers.length; i++) {
+						IMemoryRendering rendering = containers[i].getActiveRendering();
+						if (rendering == null || !(rendering instanceof IRepositionableMemoryRendering))
+							continue;
+
+						try {
+							((IRepositionableMemoryRendering) rendering).goToAddress(match.getStartAddress());
```
jonahgraham, line 195:
```suggestion
						if (!(rendering instanceof IRepositionableMemoryRendering repositionable))
							continue;

						try {
							repositionable.goToAddress(match.getStartAddress());
```

memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
```diff
@@ -173,34 +174,48 @@ public Object[] getElements(Object inputElement) {
...
+						} catch (DebugException e) {
+							MemorySearchPlugin.logError(
+									Messages.getString("MemorySearchResultsPage.RepositioningMemoryViewFailed"), //$NON-NLS-1$
+									e);
+						}
+
+						// Open the memory view to emphasize the effect of the address selection in the search view
+						IWorkbenchPart wb = containers[i].getMemoryRenderingSite().getSite().getPart();
+						if (wb == null)
```
jonahgraham, line 204: If wb == null the new version of the code won't call setSelection() anymore

This may be a non-issue as perhaps wb can't really be null in practice?

memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
```diff
@@ -297,4 +312,4 @@ public void setActionBars(IActionBars actionBars) {
 	@Override
 	public void setFocus() {
 	}
-}
+}
```
jonahgraham, line 315: (minor) The newline was lost at the end of this file - I don't know for sure if this suggestion from GitHub will fix it.

```suggestion
}
```

memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
```diff
@@ -173,34 +174,48 @@ public Object[] getElements(Object inputElement) {
...
+									Messages.getString("MemorySearchResultsPage.RepositioningMemoryViewFailed"), //$NON-NLS-1$
+									e);
+						}
+
+						// Open the memory view to emphasize the effect of the address selection in the search view
+						IWorkbenchPart wb = containers[i].getMemoryRenderingSite().getSite().getPart();
+						if (wb == null)
+							continue;
+
+						getSite().getPage().activate(wb);
```
jonahgraham, line 207: Can this code be reached multiple times for a single selection. i.e. can there be multiple `containers` that have `getActiveRendering` of type IRepositionableMemoryRendering? 

(later...)

To answer my own question, yes there can if you have a split view memory renderings open. That means activate is called multiple times. I don't know if that is a problem yet though.

jonahgraham (changes requested):
Thanks for the PR @LizzMre - I am reviewing an earlier version as I am just about to hit Submit review I see there are new changes - I'll comment on them in a moment.

Overall this looks great, mostly style changes.

There is one blocking problem though. By *always* activating the memory view on selection change the search view cannot be navigated well with a keyboard. Imagine you have the search view and memory view open in two different tab groups, you cannot press up/down arrow to scroll through all your matches anymore. Can you check if the memory view and the search view are in different tab groups, or perhaps check if the memory view is already visible.

jonahgraham:
The failed test is a known flaky test tracked here - https://github.com/eclipse-cdt/cdt/issues/194

LizzMre:
Thank you for your review @jonahgraham! I will work on the requested changes.

memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
```diff
@@ -178,16 +178,19 @@ public Object[] getElements(Object inputElement) {
 			public void selectionChanged(final SelectionChangedEvent event) {
 				IMemoryRenderingContainer containers[] = ((IMemorySearchQuery) fQuery).getMemoryView()
 						.getMemoryRenderingContainers();
-				if (containers == null || containers.length == 0) return;
+				if (containers == null || containers.length == 0)
```
jonahgraham, line 181: @LizzMre Thanks for applying CDT's coding standards. I am curious as to how you have your editor setup? The CDT projects are configured by default to auto save with this formatting so I was wondering if you are using Eclipse to edit, and if you are if you are purposely bypassing the save actions and formatting rules in the project, or is there something wrong with the CDT developer setup that means the save actions are not working in your use case?

memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
```diff
@@ -173,39 +176,64 @@ public Object[] getElements(Object inputElement) {
...
+						} catch (Exception e) {
+							MemorySearchPlugin.logError(
+									"Exception while invoking the setSelection method for the current rendering", //$NON-NLS-1$
+									e);
 						}
 					}
 				}
 			}
 		});
+		fTreeViewer.addDoubleClickListener(new IDoubleClickListener() {
```
jonahgraham, line 222: Using double-click works well and makes this consistent with other uses of the search view. Thanks!

memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
jonahgraham: You picked up a small part of this change, but not the whole thing, for example the if match == null is never true and there is a warning now on that.

I am going to split the code cleanup commit from the new behaviour (double-click) commit and update this PR.

memory/org.eclipse.cdt.debug.ui.memory.search/src/org/eclipse/cdt/debug/ui/memory/search/MemorySearchResultsPage.java
LizzMre: @jonahgraham I used Eclipse, but I forgot I had a customized formatting added for the code. I also had a few more projects opened. I don't think it's something wrong with the CDT developer setup.

jonahgraham: Great. thanks for letting me know!

LizzMre:
@jonahgraham Thank you again for all you support and very good suggestions! I didn't get to mention that in the end I chose the double click event as it had more sense and would not interfere with the simple selection behavior as you pointed out in one of your earlier comments. 

I am ok with the final version.

jonahgraham:
My pleasure. Once the CI finishes I will merge this.

> I didn't get to mention that in the end  I chose the double click event

As far as I was concerned you did mention it because it was in your commit message! It is a good solution and I appreciate you making the search result page work better.

PR #518, opened 2023-08-18
allow github to index the build directory

jonahgraham:
- https://docs.github.com/en/search-github/searching-on-github/finding-files-on-github

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,8 @@
+# to allow github to index the build directory
+# we need to override the defaults of github here
+# see https://docs.github.com/en/search-github/searching-on-github/finding-files-on-github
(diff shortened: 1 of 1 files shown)


jonahgraham:
Once I can see the search is working properly on my fork I will merge this.

This is to fix the problem identified in https://github.com/eclipse-cdt/cdt/pull/371#issuecomment-1662267107, specifically:

> However it seems that GitHub indexing isn't working because if I type AbstractGnuToolPrefixMacro in the search it says it is not there:
> 
> image
> 
> I have filed a bug report with GitHub on this issue - but I think the link is private https://support.github.com/ticket/personal/0/2276662

PR #522, opened 2023-08-19
Allocate much more ram to the Pod on Jenkins

jonahgraham:
I think with some of the recent changes (Tycho 4, API baselines, maybe even new dependency on Platform M3) it may be that our memory requirements have gone up substantially for the build.

Also, with Sonar in the works that also requires more memory.

Therefore see if the EF's JIPP infra will allocate 10G of ram to our build.

This should fix all the "Killed" messages randomly in the CDT builds. https://wiki.eclipse.org/Jenkins#What_is_killing_my_build.3F_I.27m_using_custom_containers.21

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,7 @@ pipeline {
       steps {
         container('cdt') {
           timeout(activity: true, time: 30) {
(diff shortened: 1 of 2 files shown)


jonahgraham:
PS the failures also started happening immediately after a recent infra failure that was causing all builds to break. So there is a possibility that multiple things are occurring here.

PR #547, opened 2023-09-13
Replace org.freemarker with org.freemarker.freemarker

### LINKED ISSUE #546: org.freemarker version included in p2 repo doesn't match target platform
The p2 repo for CDT includes org.freemarker 2.3.22, the target platform includes 2.3.32

Kummallinen:
Fixes #546

This will have an impact on downstream consumers as they will now need to add their own package imports.

The target platform including the CDT feature was masking this issue earlier


jonahgraham (approved):

Kummallinen:
Apologies @jonahgraham I hadn't noticed the approval notification before I pushed an amendment. The amendment just adds a note to the API Changelog

NewAndNoteworthy/CHANGELOG-API.md
```diff
@@ -7,6 +7,15 @@ This section describes API removals that occurred in past releases, and upcoming
 # API Changes
 
 Below is the detailed descriptions of API changes and mitigation efforts API consumers need to take.
+## API Changes in CDT 11.4.
```
jonahgraham, line 10: I guess this is a breaking API change, so can be listed here. I recommend it goes in the N&N for 11.4 once I create it in the next few days - like https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CDT-11.3.md#api-changes-current-and-planned

If we are doing something that is a breaking change, can we ditch the reexport too?

tools.templates/org.eclipse.tools.templates.freemarker/META-INF/MANIFEST.MF
```diff
@@ -2,12 +2,12 @@ Manifest-Version: 1.0
...
 Bundle-ManifestVersion: 2
 Bundle-SymbolicName: org.eclipse.tools.templates.freemarker
-Bundle-Version: 1.3.0.qualifier
+Bundle-Version: 1.3.100.qualifier
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Require-Bundle: org.eclipse.core.runtime,
  org.eclipse.core.resources,
- org.freemarker;visibility:=reexport,
+ org.freemarker.freemarker;visibility:=reexport,
```
Kummallinen, line 10: @jonahgraham Seems we can either add a filter to suppress the API error here (as both bundles contain the same packages so the API isn't actually broken) or we remove the re-export and bump the major version and make this a breaking API change.

Maybe a point for discussion on the call, do we want to make a breaking API change in the next release?

releng/org.eclipse.cdt.repo/category.xml
```diff
@@ -186,7 +186,7 @@
    <bundle id="org.eclipse.cdt.remote.core.source" version="0.0.0"/>
    <bundle id="com.sun.xml.bind" version="0.0.0"/>
    <bundle id="com.google.gson" version="0.0.0"/>
-   <bundle id="org.freemarker" version="0.0.0"/>
+   <bundle id="org.freemarker.freemarker" version="2.3.32"/>
```
Kummallinen, line 189: For some reason this silently fails. Checking if a different ID is needed

tools.templates/org.eclipse.tools.templates.freemarker/META-INF/MANIFEST.MF
jonahgraham: Decision in the [meeting](https://hackmd.io/@jonahgraham/cdt-call-september-13-2023) was to update freemarker to be package imports and remove the require bundle + reexport. That may lead to a some rare breakages for anyone that was relying on the reexport, but otherwise updating freemarker bundle is identified as API breakage anyway. This needs a N&N/API entry or similar.

Kummallinen:
Current draft updates all the freemarker related dependencies.

Once the main branch is ready for 11.4 changes I will also remove the re-export from the tools.templates.ui plugin currently tricky to do as every bundle shows API errors due to platform related changes.

I'll also add the N&N entry

jonahgraham:
The API errors are tracked in #551

Kummallinen:
@jonahgraham  I think this is everything discussed in the call, but could do with another set of eyes to check I didn't miss anything.

jonahgraham:
@Kummallinen I pushed a commit that updates all the items that depend on tools.templates to 2.0.0, I bumped the ui bundle to 2.0.0 for consistency and added an additional note to the API.

Can you look and see if that looks ok to you?

tools.templates/org.eclipse.tools.templates.core/META-INF/MANIFEST.MF
```diff
@@ -3,7 +3,7 @@ Automatic-Module-Name: org.eclipse.tools.templates.core
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.tools.templates.core
-Bundle-Version: 1.3.0.qualifier
+Bundle-Version: 2.0.0.qualifier
```
Kummallinen, line 6: I didn't bump this as there is no API change on this plugin. But I can see the argument for keeping it consistent with the others

jonahgraham: Thanks @Kummallinen

Issue #551, opened 2023-09-14
CDT 11.3 does not have BREE header in MANIFEST.MF, uses Require-Capability, causes false positive API Compatibility errors


jonahgraham:
Perhaps as a result of updating Tycho upgrade the CDT 11.3 manifests publish with:

```
Require-Capability: osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=17))"
```

automatically converted from BREE:

```
Bundle-RequiredExecutionEnvironment: JavaSE-17
```

This leads to false positive API errors like this:

`
2023-09-13T22:28:04.9985777Z [API ERROR] File MANIFEST.MF at line 5: The minor version should be incremented in version 8.3.0, since execution environments have been changed since version 8.3.0 (location: /home/runner/work/cdt/cdt/core/org.eclipse.cdt.core/META-INF/MANIFEST.MF)
`

jonahgraham:
I suspect the easiest thing short term is to add API filters for this error, but I need to track this back to Tycho and/or PDE to resolve fully.

akurtakov:
This is PDE issue for sure as BREE is deprecated by OSGi for years although PDE never made the move.

Kummallinen:
@jonahgraham Looking at the differences in the CDT pom & my local ones, I think the difference causing this is I have "<deriveHeaderFromSource>false</deriveHeaderFromSource>" in the tycho-compiler-plugin configuration. See https://github.com/eclipse-tycho/tycho/blob/master/RELEASE_NOTES.md#new-parameter-in-tycho-packaging-pluginpackage-plugin

jonahgraham:
The final resolution of this:

- Create an unpublished (so far) CDT 11.3.1 that has the BREE fixed (#553)
- Use that URL for api baseline instead of latest when preparing for 11.4 dev #554 

Last question, should we publish 11.3.1 fully (i.e. add it to https://download.eclipse.org/tools/cdt/releases/11.3/ composite) or just add a note as to why it is excluded. My current preference is the latter, but I am open to suggestions!

Kummallinen:
I'd vote for not publishing 11.3.1 and just adding a note. Future updates will still work so there doesn't seem to be any need to cause an extra update for users.

jonahgraham:
[...]

PR #552, opened 2023-09-14
Swapped arguments

ewaterlander:
Fixed swapped arguments in call of new ExtendedScannerInfo().

### DIFF
--- a/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
+++ b/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
@@ -706,7 +706,7 @@ public abstract class CBuildConfiguration extends PlatformObject implements ICBu
 		for (int i = 0; i < macroFiles.length; ++i) {
 			macroFiles[i] = macroFileEntries[i].getFullMacroFilePath().toOSString();
 		}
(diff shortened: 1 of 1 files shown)


jonahgraham:
Can you explain the effect of this change - i.e. is this critical, should we be releasing an emergency patch, or other?

jonahgraham:
PS As soon as I get #554 resolved we can merge this for the 11.4 branch which release in Dec. My question is if we need to emergency patch the 11.3.0 release that was this week and release a 11.3.1 that has this fix?

ewaterlander:
> My question is if we need to emergency patch the 11.3.0 release that was this week and release a 11.3.1 that has this fix?

Hi Jonah,

This problem was found by a static code analysis tool. I did not notice the problem myself. As far as I'm concerned, it's not necessary to make an emergency patch. And I'm still working with CDT 10.x, so I work with a local patch anyway.

I ran static code analysis on a few CDT source files which I needed to change for my plugin. It found another (small) problem in the same file which I'm going to upload too.

jonahgraham:
@ewaterlander I rebased this on main now that #551 is fixed - once CI checks pass this can be merged.

jonahgraham (approved):

jonahgraham:
Test failiure is known flaky #194

jonahgraham:
Thanks @ewaterlander for the fix.

PR #553, opened 2023-09-14
Restore BREE headers to all MANIFEST.MF

### LINKED ISSUE #551: CDT 11.3 does not have BREE header in MANIFEST.MF, uses Require-Capability, causes false positive API Compatibility errors

jonahgraham:
Fixes #551

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -42,7 +42,6 @@ pipeline {
                       -Dmaven.test.failure.ignore=true \
                       -DexcludedGroups=flakyTest,slowTest \
                       -P baseline-compare-and-replace \
-                      -P api-baseline-check \
                       -Ddsf.gdb.tests.timeout.multiplier=50 \
                       -Dindexer.timeout=300 \
                       -P production \
--- a/build/org.eclipse.cdt.autotools-feature/feature.xml
+++ b/build/org.eclipse.cdt.autotools-feature/feature.xml
@@ -12,7 +12,7 @@
 <feature
       id="org.eclipse.cdt.autotools"
       label="%featureName"
-      version="11.3.0.qualifier"
+      version="11.3.1.qualifier"
       provider-name="%provider"
       plugin="org.eclipse.cdt.autotools.ui"
       license-feature="org.eclipse.license"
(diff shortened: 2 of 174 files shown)


Kummallinen:
Given how long the 11.3.0 artifacts could persist in the baseline is it worth causing the extra install time for users and bumping every plugin version (or at least the ones changed in 11.3) for a 11.3.1 that we just use as a baseline for 11.4? 

So it wouldn't be released, and if we did a patch for 11.3 it would then be 11.3.2.

I really don't like it, but keeping the filters around for potentially years doesn't seem ideal either

jonahgraham:
I'm going to get 11.4 working first, I will bump all the bundles that I need filters on, that way at least those without a BREE won't be in 11.4's release. That way the filters will be removed at beginning of 11.5 cycle.

jonahgraham:
I have come around to the idea that making a (non release) 11.3.1 that has all the BREE fixed and using that for baseline for 11.4 and 11.3.2 (if we do one) is best solution. I am trying that here now.

jonahgraham:
Thanks @Kummallinen  for the support and suggestions - I am not going to be able to finish this tonight. I need to verify that the temp build running [here](https://ci.eclipse.org/cdt/job/cdt-jonahgraham-fork/job/fix-bree-11-3/) indeed has BREE restored everywhere and nothing weird left going on.

Once that is in place I will push the change, built it (in proper place, not just my fork), promote it to builds, and then use that as baseline for 11.4.0

jonahgraham:
Not sure if there is a better way to test for this, but to make sure that all the files have BREE again I did this:

1. download and unzip the [repo from build artifacts](https://ci.eclipse.org/cdt/job/cdt-jonahgraham-fork/job/fix-bree-11-3/lastSuccessfulBuild/artifact/releng/org.eclipse.cdt.repo/target/)
2. unpacked all the jars (in the plugins dir): `for i in *.jar; do ( mkdir ${i}-d; cd ${i}-d; unzip ../${i}; ) done`
3. grep'd for files missing BREE and reviewed the results: `grep -L "Bundle-RequiredExecutionEnvironment" $(find * -name MANIFEST.MF -not -path \*.source_\*)`

The only matches left in 3 are expected to not have BREE as they have no Java code.

Therefore I declare this ready to merge (but I'll repeat above test on merged branch)

PR #556, opened 2023-09-15
Missing synchronization lock on CBuildConfiguration.scannerInfoCache.

ewaterlander:
Added one missing synchronization lock for accessing CBuildConfiguration.scannerInfoCache. This could cause race conditions.

This was detected by static code analysis.

### DIFF
--- a/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
+++ b/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
@@ -678,7 +678,9 @@ public abstract class CBuildConfiguration extends PlatformObject implements ICBu
 	 * @since 6.1
 	 */
 	protected ScannerInfoCache getScannerInfoCache() {
-		return scannerInfoCache;
+		synchronized (scannerInfoLock) {
+			return scannerInfoCache;
+		}
 	}
 
 	private IExtendedScannerInfo getBaseScannerInfo(IResource resource) throws CoreException {


jonahgraham:
The API errors should be resolved soon as I am close to fixing #551

core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/CBuildConfiguration.java
```diff
@@ -678,7 +678,9 @@ protected synchronized void saveScannerInfoCache() {
 	 * @since 6.1
 	 */
 	protected ScannerInfoCache getScannerInfoCache() {
-		return scannerInfoCache;
+		synchronized (scannerInfoLock) {
```
jonahgraham, line 681: This looks fine to me on its own, but I don't understand what problem it solves. I guess the caller still needs to do additional locking if they want to use it in a consistent state, but the lock object itself is inaccessible.

Nothing in current CDT calls this method and it isn't documented.

Is this something you are using? Can you add some javadoc to explain to the next person what this method is for?

I see that even when this code was first written in 6c6901547e50469c64bbae358d686e5f7c94a628 this method was referenced.

ewaterlander: I'm not using it. Perhaps it's better to remove the method. I'm also fine with closing this PR.

jonahgraham: Lets remove it - see #558

jonahgraham:
Replaced with #559

Issue #563, opened 2023-09-17
Improve presentation of debug source and header files in Project Explorer


jld01:
Following discussion on the recent CDT project call, I am creating this issue for the purpose of grouping a number of improvements to the presentation of debug source and header files in Project Explorer.

In cases where a CDT executable project is using a pre-built library that contains debug information, a large number of C/C++ source and header files used within the library may be presented as child nodes of the executable file within _Project Explorer_ (see https://github.com/eclipse-cdt/cdt/issues/443#issuecomment-1618852660). The set of source files provided by the CDT executable project itself can be difficult to locate among all the library source file nodes. In typical usage, the library source files may not be present on the local file system.

The following improvements are proposed for CDT contributions to _Project Explorer_:

- The icons of _external_ source and header files (those files not located within the workspace) should be presented with the external file decoration (`CPluginImages.DESC_OVR_EXTERNAL_FILE`) as used by the _C/C++ Editor_ and other CDT views
- External source and header files should be sorted by file _name_ only (not by file _path_) - this allows for easier visual search by the user
- Absent source and header files should be presented with grey text
- A filter should be provided to allow the user to choose whether to hide absent source and header files
- The open file action should be disabled in the context of an absent source or header file

jld01:
With the latest CDT sources, the presentation of translation unit source files under an ELF executable that has been linked against an unstripped pre-built library is as follows:
![image](https://github.com/eclipse-cdt/cdt/assets/45816183/fa51ccc8-440e-456f-9499-4d6c2232541c)
The final improvement planned is to allow filtering of absent translation unit source files in the _Project Explorer_ view. We have previously debated whether such filtering should be enabled or disabled by default. The distinction between absent and present source files is now much clearer, but I still think it would be preferable to hide the absent files by default.

PR #570, opened 2023-09-24
Adorn images for external translation unit elements

jld01:
Part of #563


jld01:
The test failure is unrelated to this change.

PR #580, opened 2023-10-07
Implement deduction guide template

### LINKED ISSUE #438: intellisense with std::map is broken
**Describe the bug**
It looks like the intellisense does not recognize std::map.  It used to work.

**To Reproduce**
1. create a c++ project and `#include <map> #include <utility>`
2. declarations with `std::pair` works with intellisense, while `std::map` had a red underline.  Click on `#include <map>` goes to the correct header file.

**Expected behavior**
`std::map` should be recognized by the intellisense.

**Screenshots**
NA

 - OS: Ubuntu 22.04.2 LTS

i-garrison:
Add missing implementation for deduction guide template.

Add test and new indexer test strategy which more closely mirror what happens in real life (i.e. index is rebuilt after files are already in the project) which helps reproducing issue #438

Closes #438


jonahgraham (approved):

PR #585, opened 2023-10-09
Sort external translation units by name

jld01:
Part of #563


jonahgraham:
See #586 for the fix to the build failure.

PR #581, opened 2023-10-07
Amend list initialization sequence

i-garrison:
- add more tests for list-initialization of class or aggregate
- amend list-initialization handling of class or aggregate

PR #586, opened 2023-10-09
Update apt before trying to install in GHA workflows

jonahgraham:
Fixes https://github.com/eclipse-cdt/cdt/pull/581#issuecomment-1752771968


jonahgraham:
The actions got past the error we were trying to fix, so I am merging now.

PR #587, opened 2023-10-10
Fix struct declaration introducing name in namespace scope

i-garrison:
When AST is used to resolve binding for class-name and elaborated-type-specifier is found matching `[basic.lookup.elab]` rule introducing the class-name, behavior of `CPPSemantics.resolveAmbiguities()` is different in presence of index.

If there is no index, `CPPVisitor.createBinding()` for `ICPPASTElaboratedTypeSpecifier` creates binding for class-name as introducing the name. When later lookup finds this binding all is good because binding is declared before the use site.

If index is available, lookup for class-name fails in AST too but now matching entry is found in the AST index. When later lookup finds this index binding `CPPSemantics.declaredBefore()` returns false because it does not look in AST index and only checks project index.

To fix this additionally check if `ICPPClassType` object is in AST index, as we already do for `ICPPConstructor`. This way `declaredBefore()` does almost the same thing as `isReachableFromAst()` and lookup succeeds returning the same binding from index.

### DIFF
--- a/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/internal/index/tests/IndexCPPBindingResolutionBugsTest.java
+++ b/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/internal/index/tests/IndexCPPBindingResolutionBugsTest.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2006, 2015 Wind River Systems, Inc. and others.
+ * Copyright (c) 2006, 2023 Wind River Systems, Inc. and others.
  *
  * This program and the accompanying materials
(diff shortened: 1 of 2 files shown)


i-garrison:
Hi @jonahgraham I think this pipeline is now stuck in `Run Bundle Versions Bumped` it looks like something is wrong with runner.

jonahgraham:
> Hi @jonahgraham I think this pipeline is now stuck in `Run Bundle Versions Bumped` it looks like something is wrong with runner.

I have restarted the job - I have never seen it fail quite like that, but other builds ran successfully today so unless it happens again I will put it down to a transient failure.

jonahgraham:

jonahgraham:
Thanks @i-garrison for this change too.

Issue #592, opened 2023-10-11
Managed Build Definitions extension point should prefer "identifier" type of attribute 


ruspl-afed:
**Describe the bug**
Currently fields like `superClass` has type `string` and this is not helpful to specify the right value

**To Reproduce**
1. Try to define new extension for "org.eclipse.cdt.managedbuilder.core.buildDefinitions" extension point
2. Add ToolChain
3. Try to specify superClass value

**Expected behavior**
Fields that expect identifiers suggest them from the list of existing values

ruspl-afed:
Yes, it is easy to implement this for top-level entries, but we have a lot of others, like ToolChain inside Configuration.

So, it seems that I need to introduce id metadata not "by schema", but rather "by entity", to check result after each minor step.

ruspl-afed:
I'm afraid this is not possible to implement as identifiers may come from dynamic extension and we cannot determine the whole list of possible identifiers in PDE static context.

PR #594, opened 2023-10-11
Revert "[#592] Managed Build Definitions : prefer "identifier" type of attribute

ruspl-afed:
This reverts commit 7136606455671b85a6d6eca7c9a8737224275c37.


ruspl-afed:
It seems that initially during the tests I was misled: this works well for IDs defined at the top level, but not for those defined deeper in the tree.

PR #595, opened 2023-10-12
Recognize simple declaration of function name with function type specifier

i-garrison:
Simple declaration of function via function type decl specifier declares the same type as usual function declaration, fix this in the code.


i-garrison:
Hmm fixed test case to be actually valid C++ (use array of pointers to functions, array of functions is not valid construct.)

jonahgraham (approved):

jonahgraham:

jonahgraham:
Thanks @i-garrison for this contribution.

PR #597, opened 2023-10-15
Provide viewer filter for absent C/C++ translation unit files

jld01:
Part of #563

### DIFF
--- a/core/org.eclipse.cdt.ui/plugin.xml
+++ b/core/org.eclipse.cdt.ui/plugin.xml
@@ -4205,6 +4205,12 @@
           name="%HideNonCElements.label"/>
     <commonFilter
           activeByDefault="true"
+          class="org.eclipse.cdt.internal.ui.filters.AbsentTranslationUnitFilter"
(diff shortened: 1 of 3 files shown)


core/org.eclipse.cdt.ui/plugin.xml
```diff
@@ -4203,6 +4203,12 @@
           description="%HideNonCElements.description"
           id="org.eclipse.cdt.ui.navigator.filters.NonCElementFilter"
           name="%HideNonCElements.label"/>
+    <commonFilter
+          activeByDefault="true"
```
jld01, line 4207: @jonahgraham how shall we proceed on a final decision regarding default activation state of this filter?

jonahgraham: If no one objects, I think active by default as the change makes it visible and it is easy enough to restore old behavior. It is hard for users to discover new features, and if we don't activate by default most people won't find it.

If someone objects, then we can have a formal vote on cdt-dev.

PR #605, opened 2023-10-25
Fix disassembly view when view is reset

### LINKED ISSUE #603: Changing Debug contexts breaks Disassembly view
**Describe the bug**

The disassembly view stops working if the debug context changes. This can be exceptions (see the end of report for examples) or simply corrupt displaying in the view

**To Reproduce**

1. Start two debug sessions
2. Open disassembly view
3. Switch between frames in the two sessions
4. Observe exceptions in the error log

**Expected behavior**

Disassembly works with multiple sessions.

**Screenshots**

Example of the view broken:

image

 - OS and OS Version/extra details: All platforms
 - Eclipse Version from Help -> 2023-09 (it works in 2023-06)
 - CDT Version from Help -> About -> Installation Details -> Features tab 11.3.0 (works in 11.2.0)

**Additional context**

See [error log](https://github.com/eclipse-cdt/cdt/files/13169594/disassembly.log) for some exceptions that are observed.

jonahgraham:
The view can be reset in a couple of ways, by changing debug context or manually refreshing. With this change in Platform https://github.com/eclipse-platform/eclipse.platform.ui/pull/963 the CDT violation of the API aronud resetting state of the document was exposed.

Therefore the solution is to instead of trying to reset the state of the existing document, create a new one when the view is reset.

Fixes #603

### DIFF
--- a/dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly/DisassemblyPart.java
+++ b/dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly/DisassemblyPart.java
@@ -2126,8 +2126,10 @@ public abstract class DisassemblyPart extends WorkbenchPart
 		fPCHistory.clear();
 		fPendingPCUpdates.clear();
 		fFile2Storage.clear();
-		fDocument.clear();
+		var oldDocument = fDocument;
+		fDocument = createDocument();
 		fViewer.setDocument(fDocument, new AnnotationModel());
+		oldDocument.dispose();
 		if (fDebugSessionId != null) {
 			attachBreakpointsAnnotationModel();
 			attachExtendedPCAnnotationModel();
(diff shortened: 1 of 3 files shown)


dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly/DisassemblyPart.java
```diff
@@ -2126,8 +2126,10 @@ private void resetViewer() {
 		fPCHistory.clear();
 		fPendingPCUpdates.clear();
 		fFile2Storage.clear();
-		fDocument.clear();
+		var oldDocument = fDocument;
```
jonahgraham, line 2129: I don't know if it is necessary to delay the dispose call until after the setDocument. On simple tests it works either way, but because the old document is passed to listeners in `fViewer.setDocument` call, it seems reasonable to leave it not disposed until the setDocument returns.

jonahgraham (commented):
@betamaxbandit here is a change that I think covers the use case. Please let me know what you think.

dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/disassembly/DisassemblyPart.java
betamaxbandit: Thanks @jonahgraham ,
I've ported your fix to my fork and it appears to work well.

I am also able to make the vanilla CDT Disassembly view fault using the approaches you describe.

Issue #608, opened 2023-11-03
MBS support for grouping libraries on the GNU ld command line


jld01:
This feature request was raised at https://github.com/eclipse-embed-cdt/eclipse-plugins/issues/592 but I think it would be better implemented here in CDT rather than made specific to Embedded CDT.

jantje:
This is the exact same reason I started autoBuild.IMHO one need to hard code stuff or change the extension point schema to implement this issue.
Below is an extract from the CDT dev mailing list

>>Why did you need to modify the CDT managed build system (MBS) extension point schema?

>The current gnuMakeFileGenerator always put input files at the same place. To really support the Arduino framework Sloeber has to put the input files of the type archive on the command line on a different place from the input type object files during link time. (link <bla bla> ${archives} <more bla bla> ${object files} <Another bla bla> ${output files} <last bla bla>)

fanghuaqi:
Hi @jld01 ， will this feature also do group library search for other objects?

![image](https://github.com/eclipse-cdt/cdt/assets/1538922/11c9af59-acc3-47c7-a8df-bc74b3092c45)

jld01:
> ... will this feature also do group library search for other objects?

@fanghuaqi, the object files listed under _Other objects_ are presented to the linker in the same way as object files generated from source files within the project. Does that answer your question?

fanghuaqi:
> @fanghuaqi, the object files listed under _Other objects_ are presented to the linker in the same way as object files generated from source files within the project. Does that answer your question?

@jld01 Thanks, it is clear for me now.

PR #614, opened 2023-11-10
Add GNU C++ linker library grouping option to MBS

jld01:
Part of #608


jld01:
Test failures are not related to this change (#194).

Issue #622, opened 2023-11-16
Launch can connect to serial device rather than REMOTE_TCP under certain conditions


betamaxbandit:
**Describe the bug**
When launching a GDB debug session using the Launch Bar and a Launch Target, the launch can become confused and set the FinalLaunchSequence so that the launch attempts to connect to a serial device rather than a remote TCP connection.

**To Reproduce**
(found during development of a new LaunchConfigurationDelegate so no UI steps)
When the debug launch is of type ITargetedLaunch *and* the ILaunchTarget has been set [1], the logic in GDBControl.completeInitialization(...), introduced in patch [2], can incorrectly override a true value of the Launch Configuration attr IGDBLaunchConfigurationConstants.ATTR_REMOTE_TCP.

I think this is unintentional and a consequence of a number of compounded issues in my scenario:

launch instanceof ITargetedLaunch
target != null
target.getTypeId() != GDBRemoteTCPLaunchTargetProvider.TYPE_ID

When the above are true then the attribute IGDBLaunchConfigurationConstants.ATTR_REMOTE_TCP from the launch config attributes is ignored.

**Expected behavior**
Expected: when IGDBLaunchConfigurationConstants.ATTR_REMOTE_TCP is set true in the launch config, that a REMOTE_TCP launch, and not a serial launch, is started.

**Version Information (please complete the following information):**
 - OS and OS Version: Windows, Linux, macOS
 - Eclipse Version: 4.29.0
 - CDT Version: 11.3.1

1: There is an opportunity to set the ILaunchTarget in the Launch Configuration Delegate, ILaunchConfigurationTargetedDelegate.getLaunch(ILaunchConfiguration, String, ILaunchTarget), method. For example:
```
  class MyLaunchConfigurationDelegate extends LaunchConfigurationDelegate {
  ...
    @Override
    public ITargetedLaunch getLaunch(ILaunchConfiguration configuration, String mode, ILaunchTarget target)
        throws CoreException {
      GdbLaunch launch = (GdbLaunch) super.getLaunch(configuration, mode);
      launch.setLaunchTarget(target);
      return launch;
    }
```
2: GDBControl.completeInitialization(...)
https://git.eclipse.org/r/c/cdt/org.eclipse.cdt/+/143222/1/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/service/command/GDBControl.java#443
[...]

PR #625, opened 2023-11-21
Allows launch config to override launch target connection type

betamaxbandit:
When launching a GDB debug session using the Launch Bar and a Launch Target, now the launch takes into account a REMOTE_TCP value in the launch configuration before considering the type of the Launch Target.

Improves the change submitted in:
Bug 547881 - Allow targets to set ATTR_REMOTE_TCP

Fixes issue #622

### DIFF
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.cdt.dsf.gdb;singleton:=true
-Bundle-Version: 7.1.200.qualifier
+Bundle-Version: 7.1.300.qualifier
 Bundle-Activator: org.eclipse.cdt.dsf.gdb.internal.GdbPlugin
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime,
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/service/command/GDBControl.java
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/service/command/GDBControl.java
@@ -439,16 +439,19 @@ public class GDBControl extends AbstractMIControl implements IGDBControl {
 		}
 
 		// And optionally the target
-		if (launch instanceof ITargetedLaunch) {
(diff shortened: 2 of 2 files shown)


betamaxbandit:
Hi @jonahgraham ,

I'm not sure how to test this!
Either a manual UI test could be done or a JUnit added.
Manual test: I tried a manual test but didn't get past trying to launch it [1]

JUnit: I've investigated this a bit and it looks doable, but not sure whether it's worth it [2]

Do you have any advice please?

[1] Manual test
Using CDT New C/C++ Project wizard, created a CMake Project.

In Preferences > Core Build Toolchains, the auto detected GCC is shown:
Core Build Toolchains
  Available Toolchains
    Type    Name                                         OS     Arch
    GCC     win32 x86_64: C:\msys64\mingw64\bin\gcc.exe  win32  x86_64

In Preferences > CMake, I downloaded an example CMake toolchain file (MSYS2.cmake) from internet and added it to the toolchain above.

CMake
  Toolchain Files
    Toolchain File    Toolchain
      MSYS2.cmake     win32 x86_64: C:\msys64\mingw64\bin\gcc.exe

I created a new Generic Target launch target:
  Name: gen
  Operating System: win32
  CPU Architecture: x86_64

I built the project. An executable was successfully generated.

Click Launch Bar Run (Launch in Run mode) button
Console: No action specified. Skipping.


[2] Junit
I noticed org.eclipse.cdt.tests.dsf.gdb.launching.TestLaunchDelegate. I can extend this class and override 

  GdbLaunchDelegate.getLaunch(ILaunchConfiguration, String)

And then set the launch target using GdbLaunch.setLaunchTarget(ILaunchTarget)
      
  Then setup a launch configuration which includes IGDBLaunchConfigurationConstants.ATTR_REMOTE_TCP attr set to true.
  
Then launch and inspect whether the FinalLaunchSequence.stepRemoteConnection(RequestMonitor) indeed attempts a isTcpConnection rather than a serial connect. [Not sure how to detect this]

jonahgraham (changes requested):
This change lgtm - I can merge it after I branch which is in the new few days.

My feeling for a test here is that a lot of scaffolding is probably needed to automate this test. I suspect the dev effort required to do it for this change doesn't warrant it. But I think we should make sure that at least a manual test runs, even if it is just for the default case where the attribute is not in the launch already. From your [description](https://github.com/eclipse-cdt/cdt/pull/625#issuecomment-1820783394) it sounds like this has exposed other problems with the cmake integration. Some of them may already be [reported in this query](https://github.com/eclipse-cdt/cdt/issues?q=is%3Aissue+is%3Aopen+cmake)

jonahgraham (approved):
@betamaxbandit and I had a call about this and we verified that the behavior wasn't regressing as requested. So this is good to go. I am going to rebase the change to make sure there are no conflicts/issues

jonahgraham:
Thanks @betamaxbandit for this fix.

Issue #646, opened 2023-12-14
Keeping plugin.xml files clean


jld01:
On the CDT monthly call (2023-12-12) we discussed clean up of `plugin.xml` files and https://github.com/eclipse-cdt/cdt/blob/main/build/org.eclipse.cdt.managedbuilder.gnu.ui/plugin.xml in particular.

I have looked at a few command-line tools for XML clean up. None of them seem to provide an exact match to the formatting delivered by the _Plug-in Manifest Editor_, but we can get close using [HTML Tidy](https://www.html-tidy.org/) (which also supports XML). This tool is mature and packaged with most Linux distributions, plus MSYS2 and Homebrew.

I will create a draft PR for comment on the formatting style.

jld01:
Thanks for your comments on #647, @Bananeweizen. We're not committed to any specific approach on XML clean up, yet. In particular, we need to decide which XML files to clean up and whether that clean up will be a one-off activity or become part of CI in some way.

I agree that we do need caution with any reformatting of text content (between opening and closing tags). For many CDT `plugin.xml` files, this is not an issue as text content is not used.

Regarding XML attribute sorting, this was specifically enabled for illustration within my [example reformatted plugin.xml](https://github.com/eclipse-cdt/cdt/blob/0f2e98d916d4276a4732a0b00a9bf42880508adf/build/org.eclipse.cdt.managedbuilder.gnu.ui/plugin.xml). _HTML Tidy_ also supports prioritization of attributes, allowing one particular attribute to be placed first. Assuming we place each attribute on its own line and sort the attributes alphanumerically, my personal view is that the `id` attribute is easy enough to find.

@jonahgraham, perhaps we can discuss next step(s) on the next CDT project call.

jonahgraham:
> @jonahgraham, perhaps we can discuss next step(s) on the next CDT project call.

I added it to the agenda for the [Jan 10 call](https://hackmd.io/@jonahgraham/cdt-call-january-10-2024)

PR #647, opened 2023-12-14
XML whitespace clean up only

jld01:
Using _HTML Tidy_:

`tidy --input-xml yes --output-xml yes --wrap 0 --indent yes --indent-spaces 3 --indent-attributes yes --sort-attributes alpha -modify plugin.xml`

Relates to #646


Bananeweizen:
I'm not a CDT developer, but I would think twice about the sorting of attributes. At least I myself am used to finding the ID attribute as the first one, and that change would make this much harder.
Besides that, have you checked the behavior of the reformatting on multiline text string attributes? I haven't seen an example in the change, but I'm thinking of stuff like feature descriptions. As those are taken _verbatim_ from the feature.xml, reformatting them with line wrapping or indentation always leads to end user visible changes. Similar fields might exist in plugin.xml (e.g. a perspective description).

jld01:
Abandoning this PR in favour of one-off formatting using Eclipse IDE (#801).

PR #648, opened 2023-12-16
[#407] fix CompilerBuiltinsDetector for msys2 on Windows

### LINKED ISSUE #407: CMake project: CompilerBuiltinsDetector does not work for msys2 gcc compiler
When using the gcc/g++ compiler from [MSYS2](https://www.msys2.org/) on Windows 10 in a CMake project, the CompilerBuiltinsDetector cannot fetch the builtins by running the compiler and gathering the results. The line handed to `org.eclipse.cdt.jsoncdb.core.participant.builtins.OutputSniffer.processLine(String)`  is an empty string, because the output buffer is empty. 
The process exit value is 1, but 0 is expected in `org.eclipse.cdt.jsoncdb.core.internal.builtins.CompilerBuiltinsDetector.detectBuiltins(IProject, Path, ICommandLauncher, IConsole, IProgressMonitor)`:
```java
if (state != ICommandLauncher.COMMAND_CANCELED) {
try {
	// check exit status
	final int exitValue = proc.exitValue();
```
Has anybody an idea what the return value of 1 mean? What I've seen so far is, that this value comes from the `Java_org_eclipse_cdt_utils_spawner_Spawner_waitFor` function. This is a C function in `/org.eclipse.cdt.core.native/native_src/win/Win32ProcessEx.c`. The return value comes from this function: [GetExitCodeProcess](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getexitcodeprocess#return-value)
I cannot find a description for 1 as value of  `lpExitCode`.

When I use the `clang` compiler in the CMakeList.txt from MSYS2 it works:
```txt
set(CMAKE_C_COMPILER "clang")
set(CMAKE_CXX_COMPILER "clang++")
```
1. Windows machine with MSYS2 and cmake and gcc installed
2. Add bin folder of MSYS2 installation to PATH user environment variable.
3. Create a new CMake project in CDT
4. Build it
5. See that the system inlcudes cannot be resolved (similar to #376)
image

**Expected behavior**
The CompilerBuiltinsDetector  fetches the system includes from the compiler call: e.g.
```txt
C:/msys2/msys64/mingw64/bin/cc.exe -E -P -dM -Wp,-v C:\TR\workspaces\runtime-cdt-development\Cmk1\build\cmake.debug.win32.x86_64\detect_compiler_builtins.cpp
```

 - OS: Windows 10 Enterprise
 - Version 21H2

ghentschke:
The cc.exe from mingw64 (part of Msys2) on Windows needs the bin folder to be on the PATH to be executed. e.g. 'C:\msys64\mingw64\bin' must be part of the PATH environment variable

fixes #407

### DIFF
--- a/jsoncdb/org.eclipse.cdt.jsoncdb.core/src/org/eclipse/cdt/jsoncdb/core/internal/builtins/CompilerBuiltinsDetector.java
+++ b/jsoncdb/org.eclipse.cdt.jsoncdb.core/src/org/eclipse/cdt/jsoncdb/core/internal/builtins/CompilerBuiltinsDetector.java
@@ -14,6 +14,7 @@ import java.nio.file.Files;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Objects;
 
@@ -183,19 +184,33 @@ public class CompilerBuiltinsDetector {
 	 *         return {@code null}.
 	 */
 	private String[] getEnvp() {
+		var map = new HashMap<String, String>();
+		// The cc.exe from mingw64 (part of Msys2) on Windows needs the bin folder to be on the PATH to be executed.
+		// e.g. 'C:\msys64\mingw64\bin' must be part of the PATH environment variable. That's why we need PATH here:
+		// Fixes CDT #407
+		map.put("PATH", System.getenv("PATH")); //$NON-NLS-1$//$NON-NLS-2$
+
 		// On POSIX (Linux, UNIX) systems reset language variables to default (English)
 		// with UTF-8 encoding since GNU compilers can handle only UTF-8 characters.
 		// Include paths with locale characters will be handled properly regardless
 		// of the language as long as the encoding is set to UTF-8.
 		// English language is set for parser because it relies on English messages
 		// in the output of the 'gcc -v'.
-		String[] strings = {
-				// override for GNU gettext
-				"LANGUAGE" + "=en", //$NON-NLS-1$//$NON-NLS-2$
-				// for other parts of the system libraries
-				"LC_ALL" + "=C.UTF-8" }; //$NON-NLS-1$ //$NON-NLS-2$
 
-		return strings;
+		// override for GNU gettext
+		map.put("LANGUAGE", "en"); //$NON-NLS-1$ //$NON-NLS-2$
+		// for other parts of the system libraries
+		map.put("LC_ALL", "C.UTF-8"); //$NON-NLS-1$ //$NON-NLS-2$
+
+		var envArray = map.entrySet().stream().map(entry -> {
+			var result = entry.getKey() + '=';
+			if (entry.getValue() != null) {
+				result = result + entry.getValue();
+			}
+			return result;
+		}).toArray(String[]::new);
+
+		return envArray;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


jsoncdb/org.eclipse.cdt.jsoncdb.core/src/org/eclipse/cdt/jsoncdb/core/internal/builtins/CompilerBuiltinsDetector.java
```diff
@@ -183,19 +184,33 @@ private List<String> getCompilerArguments() {
 	 *         return {@code null}.
 	 */
 	private String[] getEnvp() {
+		var map = new HashMap<String, String>();
+		// The cc.exe from mingw64 (part of Msys2) on Windows needs the bin folder to be on the PATH to be executed.
+		// e.g. 'C:\msys64\mingw64\bin' must be part of the PATH environment variable. That's why we need PATH here:
+		// Fixes CDT #407
+		map.put("PATH", System.getenv("PATH")); //$NON-NLS-1$//$NON-NLS-2$
```
jld01, line 191: The PATH may be modified at CDT build configuration level. I think we need to use CCorePlugin.getDefault().getBuildEnvironmentManager().getVariables(cfg, true) here. Does that make sense?

jld01 (changes requested):

jsoncdb/org.eclipse.cdt.jsoncdb.core/src/org/eclipse/cdt/jsoncdb/core/internal/builtins/CompilerBuiltinsDetector.java
ghentschke: The environment variables of CDT are not used in this case, since it's a a cmake and not a manged build project. There are no env. variables: 
image
The `CompilerBuiltinsDetector` is used to parse a compile_commands.json file which is produced by cmake.

jld01: Hi Gesa, the build environment is available and relevant for any CDT project (not just managed build projects). In the case of tools such as GCC on Windows hosts, CDT detects the location of the tool installation from the Windows Registry and adds the location to the PATH dynamically. If the tool installation is not detected automatically, users may add the tool location to the PATH themselves for relevant build configurations. Typically, the tool location will not be on the global PATH inherited from the operating system.

jld01: Hi Gesa, I now note from your screenshot that there is no _Environment_ properties page. Is there are reason why this page is supressed? It seems relevant to _all_ CDT projects.

ghentschke: Hi John, I don't know. I think the whole cmake stuff was implemented by @15knots .

jld01 (approved):

ghentschke:
@jld01 Thank you!

jsoncdb/org.eclipse.cdt.jsoncdb.core/src/org/eclipse/cdt/jsoncdb/core/internal/builtins/CompilerBuiltinsDetector.java
15knots: > Hi John, I don't know. I think the whole cmake stuff was implemented by @15knots .

No, it was implemented by Doug Schaefer. I just added the indexer support and a cmake error parser and hooked that in to the existing cmake stuff.

PR #654, opened 2023-12-20
Support for __is_function Clang/GCC built-in

MarkZ3:
I don't remember exactly which code didn't parse for me initially (probably MSVC), but I've found this in some std::is_function (type_traits) implementation:
 #ifdef __clang__
__is_function(_Tp)
...

This is a missing built-in in CDT.

### DIFF
--- a/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/core/dom/parser/cpp/GPPScannerExtensionConfiguration.java
+++ b/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/core/dom/parser/cpp/GPPScannerExtensionConfiguration.java
@@ -44,6 +44,7 @@ public class GPPScannerExtensionConfiguration extends GNUScannerExtensionConfigu
 	private static final int VERSION_8_0 = version(8, 0);
 	private static final int VERSION_10_0 = version(10, 0);
 	private static final int VERSION_11_1 = version(11, 1);
+	private static final int VERSION_14_0 = version(14, 0);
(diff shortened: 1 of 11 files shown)


core/org.eclipse.cdt.core/parser/org/eclipse/cdt/core/dom/parser/cpp/GPPScannerExtensionConfiguration.java
```diff
@@ -246,7 +246,7 @@ public GPPScannerExtensionConfiguration(CompilerType compiler, int version) {
 			addKeyword(GCCKeywords.cp__is_enum, IGCCToken.tTT_is_enum);
 			addKeyword(GCCKeywords.cp__is_final, IGCCToken.tTT_is_final);
 			// __is_floating_point
-			// __is_function
+			addKeyword(GCCKeywords.cp__is_function, IGCCToken.tTT_is_function);
```
i-garrison, line 249: Note recently GCC added `__is_function` too https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=ffc08431f5c89a20dab21f958bb408d6a604cf54
It is probably safe to add this builtin to GCC above.

MarkZ3: Thanks. I assume it's going to be in GCC 14.0 based on the roadmap so I added it there.

jonahgraham:
Thanks @MarkZ3

PR #655, opened 2023-12-21
Fix a case of variable template instance not resolved

MarkZ3:
When the variable template was instantiated through an implicit name (constructor), the current look-up point was used to determine whether or not the variable instance was an explicit specialization but it's not enough. During resolution of implicit name, the look-up was on the constructor call, not on the variable instance. I'm not sure if the current look-up should be changed but we already had the information about the AST node being an explicit specialization down the stack, so we just pass that info now and it seems safer than changing the look-up point.


jonahgraham:
Thanks @MarkZ3 (and nice to see you here!)

PR #658, opened 2023-12-21
clean up maven configuration

Bananeweizen:
* remove invalid tycho resolver config (not valid since many versions)
* fix PMD plugin configuration (config attribute was renamed in 3.18)
* remove duplicate versions already managed in pluginManagement section
* fix indentation
* move all not yet configured versions into pluginManagement of the parent module to avoid further duplication
* consume the target platform as file, not as maven artifact (this simplifies building a single module A LOT)
* remove invalid tycho-source-plugin configuration
* disable tycho consumer POM generation (not needed here, and avoids a warning in each module)
* upgrade tycho to 4.0.4
* remove prerequisites. that's for maven plugins only, and this POM already contains the better suited similar maven enforcer rule
* make Javadoc generation work on Windows by choosing the correct encoding


jonahgraham:
This change looks good, but for some reason it is now failing API checks. Any idea why?

Bananeweizen:
> This change looks good, but for some reason it is now failing API checks. Any idea why?

No. And I have looked at it at 2 different days already for fresh ideas. :) The normal API bumps missing from the first 2 or 3 versions of the PR are clear (and I couldn't test locally because of Tycho behind company proxy issues). But now that all features and bundles have been bumped as told by the previous PR failure messages, I'm lost on why it complains about actual class compatibility changes, as there are absolutely no Java changes in my PR.

However, meanwhile I have fixed the proxy issue in Tycho, so I'll give it one more try for reproducing locally...

Bananeweizen:
Since I was not able to find any issue locally (there are no API changes being reported), I've created a second version which leaves out the Tycho upgrade: #662.

PR #662, opened 2023-12-28
Clean up Maven build

Bananeweizen:
* remove invalid tycho resolver config (not valid since many versions)
* fix PMD plugin configuration (config attribute was renamed in 3.18)
* remove duplicate versions already managed in pluginManagement section
* fix indentation
* move all not yet configured versions into pluginManagement of the parent module to avoid further duplication
* consume the target platform as file, not as maven artifact (this simplifies building a single module A LOT)
* remove invalid tycho-source-plugin configuration
* disable tycho consumer POM generation (not needed here, and avoids a warning in each module)
* remove prerequisites. that's for maven plugins only, and this POM already contains the better suited similar maven enforcer rule
* make Javadoc generation work on Windows by choosing the correct encoding


jonahgraham (approved):
Thanks @Bananeweizen for splitting this change and working on this cleanup.

PR #667, opened 2024-01-05
Partial fix for bug 574196/530103, error involving sizeof...()

MarkZ3:
Also discovered that non-type param is not handled at all.  This case is not covered in this commit but there is a commented-out failing test for it.


jonahgraham (approved):

jonahgraham:
Thank you @MarkZ3

Issue #666, opened 2024-01-02
*.S files are not preprocessed in managed builds


kisvegabor:
**Describe the bug**
`*.S` assembly files are not preprocessed in managed builds

**To Reproduce**

1. Create a new Hello World managed build C project
2. Add `test.S` to the `src` folder
3. Add `nop` to `test.S`
4. Build the project, it works
5. Change the content of `test.S` to
```c
#ifdef MY_ASM_ENABLE
some_invalid_asm_instruction
#endif
```
6. Build the project, now it fails with
```
../src/test.S:2: Error: no such instruction: `some_invalid_asm_instruction'
make: *** [src/subdir.mk:31: src/test.o] Error 1
```
despite of `MY_ASM_ENABLE` is not defined, so `some_invalid_asm_instruction` should be ignored

**Expected behavior**
Preprocess `*.S` files

**Screenshots**
![image](https://github.com/eclipse-cdt/cdt/assets/7599318/0479cb5f-58b7-472f-9872-1ad8951aeb07)

**Version Information (please complete the following information):**
 - OS and OS Version/extra details: Linux Mint (LMDE 5)
 - Eclipse Version from Help -> About: 2023-12 (4.30.0)
 - CDT Version from Help -> About -> Installation Details -> Features tab: 11.4.0

**Additional context**
I could workaround the problem by adding `*.S` as C source file:
![image](https://github.com/eclipse-cdt/cdt/assets/7599318/f80a09d4-cd57-41b7-b71f-16f4e403d455)

jld01:
Hi @kisvegabor, GNU `as` does not invoke the C preprocessor. In order to pre-process your `test.S` file, you will need to use `gcc` as the assembler _Command_ and add `-c` to the _Assembler flags_ option:

![image](https://github.com/eclipse-cdt/cdt/assets/45816183/368f11ec-c927-4816-93cc-484820b2b607)

@jonahgraham and @Torbjorn-Svensson, we should discuss making this the default behaviour.

kisvegabor:
It's cleaner solution, thank you! 

As far as I know `*.s` (lower case) files should not be preprocessed. Will `gcc` make a decision based on the extension?

jld01:
> As far as I know `*.s` (lower case) files should not be preprocessed. Will `gcc` make a decision based on the extension?

Yes.

kisvegabor:
Great, thank you John!

Torbjorn-Svensson:
[...]

PR #671, opened 2024-01-14
Invoke "gcc" to build assembler files

jld01:
Enables pre-processing of *.S and *.sx source files.

Part of #666


jld01:
I am hopeful that this patch will address concerns about the migration of existing projects when invoking `gcc` rather than `as` to build assembly files in managed build projects. My personal opinion is that a switch to using `gcc` is justified here because the current behaviour is not aligned with user expectations for building `*.S` and `*.sx` files.

I have added a command generator for the _Assembler flags_ MBS option that looks for cases where the tool command is `gcc` but the assembler flags option does _not_ include the `-c` flag. In such cases:

- The `-c` flag is added
- Other assembler flags that are not automatically passed to the assembler by the `gcc` wrapper are prefixed with `-Wa,`

For example, a project that was previously building assember files using `as -K -g` would now use `gcc -c -Wa,-K -g`.

ruspl-afed:
Looks good for the simple case. And this covers part of the issues raised during the CDT discussion.

And what about [so nice] MBS inheritance? There are three values changes that can participate to the inheritance chain:
* Tool "command"
* Tool "commandGenerator"
* Option "defaultValue"

How do we ensure that this fix will not break downstream integrations?

jld01:
Thank you the feedback, @ruspl-afed.

> And what about [so nice] MBS inheritance? There are three value changes that can participate to the inheritance chain:
> 
>     * Tool "command"
>     * Tool "commandGenerator"
>     * Option "defaultValue"
> 
> How do we ensure that this fix will not break downstream integrations?

The latest version of this PR accommodates cases where the tool command has been overridden XOR the option default value has been overridden. There is no issue when both have been overridden. There is no issue when neither has been overridden.

This work does _not_ accommodate the possibility that the tool command generator has been overridden while the tool command and option default value have not been overridden. The risk here seems very low.

build/org.eclipse.cdt.managedbuilder.gnu.ui/src/org/eclipse/cdt/managedbuilder/gnu/ui/GnuAsmFlagsCommandGenerator.java
```diff
@@ -0,0 +1,97 @@
...
+	private static final Pattern DO_NOT_LINK_PATTERN = Pattern.compile("(^|\\s)-c($|\\s)"); //$NON-NLS-1$
+	private static final Pattern ASM_FLAG_PATTERN = Pattern.compile("-[aDKLR]\\S*"); //$NON-NLS-1$
+
+	@Override
+	public String generateCommand(IOption option, IVariableSubstitutor macroSubstitutor) {
+		String toolCommand = getToolCommand(option, macroSubstitutor);
+		try {
+			if (null != toolCommand && IOption.STRING == option.getValueType() && option.getCommand().isEmpty()) {
+				String optionValue = option.getStringValue();
+				if (toolCommand.equals("gcc")) { //$NON-NLS-1$
```
Torbjorn-Svensson, line 52: Sorry to be late the party here, but should this line not match `arm-non-eabi-gcc` for example, or `gcc.exe`? Or is it enough to have a strict check for `gcc`?

jld01: We are checking only whether the tool command has been overridden. We don't care if it has been overridden by the user or by an extender. If the tool command has been overridden, then our new default value (`gcc`) will have no effect on the build and we do not need to modify the assembler flags.

PR #676, opened 2024-01-18
Made PreprocessorMacro and TokenList public

max-leuthaeuser:
### Changes

With this PR, `PreprocessorMacro` and its subclasses are no longer package-private but public instead. Same goes for `TokenList`.

### Reasons for the changes

When parsing C/C++ code with CDT as a standalone library I want to track certain macro expansions using the [MacroExpander](https://github.com/eclipse-cdt/cdt/blob/main/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/parser/scanner/MacroExpander.java).

It is not actually part of the public API but declared as `public`. It takes an argument `macroDictionary` of type `CharArrayMap<PreprocessorMacro>` in its constructor. Currently, `PreprocessorMacro` is package-privat and thus, not available without shadowing (i.e., implementing my custom macro tracker in the package `org.eclipse.cdt.internal.core.parser.scanner`).

It questionable whether it is a good design decision to have public and private API mixed together here (rendering the usage of the constructor of `MacroExpander` impossible).

Also, another problem occurs once you take the route of shadowing to have access to `PreprocessorMacro`: the CDT jars are all signed during release (makes sense) but, sadly speaking, that will result in a `SecurityException` when trying to call their API from your own (unsigned) code which lives in the shadowed package (in my case in `org.eclipse.cdt.internal.core.parser.scanner`).

### Problem solved

MacroExpander can now be used from the outside because `PreprocessorMacro` and `TokenList` are no longer package-private. Shadowing is no longer needed. Signed jars are no longer a problem.


jonahgraham (approved):
As this doesn't change the API contract (the change in visibility are in `internal` packages) I am +1 to make this change if it makes your use case simpler.

This can be merged once CI signs off on it.

jonahgraham:
As it turns out this doesn't build as it uses some illegal Java, e.g.:

```
/home/runner/work/cdt/cdt/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/parser/scanner/PreprocessorMacro.java:[115] 
Error:  	public class ObjectStyleMacro extends PreprocessorMacro {
Error:  	             ^^^^^^^^^^^^^^^^
Error:  The public type ObjectStyleMacro must be defined in its own file
```

jonahgraham:
PS As this is your first PR the CI build does not run automatically. You can enable GitHub Actions in your fork and then the CI will run on your fork to test them.

max-leuthaeuser:
Thanks for the help!
Looks good here: https://github.com/max-leuthaeuser/cdt/actions

jonahgraham:
Thank you @max-leuthaeuser for this improvement.

Issue #689, opened 2024-02-02
Add compile-commands.json generation for Managed Build Projects


Kummallinen:
To support the CDT-LSP editor & other uses of language servers or external tools that expect a cmake compilation database we should add support for generating "compile-commands.json" to managed build projects.

There is already enough data in the existing managed build system to generate the file, so long as any extenders are using the command generator interfaces where they needed to customise tool commands.

Should be enabled by default, but have an option to disable on a per-project basis.

Need to consider how multiple configurations, probably just 1 file which matches active project unless we can make the language launcher aware of them.

File should be placed where ClangD expects to find it, so either in the project root or "build" directory. I suggest the "build" directory.

Renesas already has an internal version of this feature, we will contribute this & work with the community to modify as needed to fit CDTs requirements.

jld01:
How can we ensure that `clangd` picks up the correct set of compiler commands based on the active build configuration?

We could simply overwrite `compile_commands.json` when the active build configuration changes. But the active build configuration is persisted in workspace metadata - it is personal to the user. So if we overwrite a project file based on the active build configuration, it seems advisable to mark the file as a derived resource (not shared with other users).

An alternative would be to write multiple compilation database files to the relevant MBS build configuration folder(s). We could then update the `.clangd` file in the root of the project to point to the correct compilation database when the active build configuration is changed. I note that the CDT-LSP project feature already manipulates a per-project `.clangd` file and provides a [setCompilationDatabase](https://github.com/eclipse-cdt/cdt-lsp/blob/2fe1c1dae0fe287eee2f2565c6ffeecac5047807/bundles/org.eclipse.cdt.lsp.clangd/src/org/eclipse/cdt/lsp/clangd/ClangdConfigurationFileManager.java#L183) method. Any thoughts @ghentschke?

I would suggest enabling generation of the compilation database by default only in cases where the CDT-LSP C/C++ editor in installed in the first instance. At present, only a small minority of CDT users will make use of this file.

ghentschke:
> An alternative would be to write multiple compilation database files to the relevant MBS build configuration folder(s). We could then update the .clangd file in the root of the project to point to the correct compilation database when the active build configuration is changed.

[...]

PR #692, opened 2024-02-05
Added feature to generate compile_commands.json file

alicetrifu:
Added a new generator class to provide a compile_commands.json file in the "build" folder. Created a new extension point to provide the opportunity to incorporate additional information into the file.

Part of #689


build/org.eclipse.cdt.managedbuilder.core/META-INF/MANIFEST.MF
```diff
@@ -12,6 +12,7 @@ Export-Package: org.eclipse.cdt.build.core.scannerconfig,
  org.eclipse.cdt.managedbuilder.buildmodel,
  org.eclipse.cdt.managedbuilder.buildproperties,
  org.eclipse.cdt.managedbuilder.core,
+ org.eclipse.cdt.managedbuilder.core.jsoncdb.generator.api,
```
Kummallinen, line 15: Export the API as "org.eclipse.cdt.managedbuilder.core.jsoncdb"

build/org.eclipse.cdt.managedbuilder.core/META-INF/MANIFEST.MF
```diff
@@ -39,7 +40,8 @@ Require-Bundle: org.eclipse.cdt.core;bundle-version="[8.3.0,9.0.0)",
  org.eclipse.ui;bundle-version="[3.2.0,4.0.0)",
  org.eclipse.core.variables;bundle-version="[3.1.100,4.0.0)",
  org.eclipse.cdt.make.core;visibility:=reexport,
- org.eclipse.core.filesystem;bundle-version="1.2.0"
+ org.eclipse.core.filesystem;bundle-version="1.2.0",
+ org.eclipse.jdt.annotation;bundle-version="[2.2.7,3.0.0)";resolution:=optional
```
Kummallinen, line 44: @jonahgraham do we want to use Null annotations in CDT or should this be dropped?

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -0,0 +1,417 @@
...
+	public CompilationDatabaseGenerator() {
+		// Nothing
+	}
+
+	/**
+	 * @param proj
+	 *            Creates and populates compilation database file
+	 * @param config
+	 */
+	public void initialize(IProject proj, IConfiguration config) {
```
Kummallinen, line 76: I think this should be renamed to "generate" & the parameters made parameters of the constructor instead

Kummallinen (changes requested):
Quick initial pass

You also need to get your API baseline set correctly, as there mare API changes here so you need feedback from the API tooling to get the versions set correctly & all appropriate since tags in place

Kummallinen:
See https://github.com/eclipse-cdt/cdt/blob/main/CONTRIBUTING.md#setup-cdt-for-development-manual-setup for API tooling setup guidance

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/CommonBuilder.java
```diff
@@ -505,6 +505,8 @@ private IProject[] build(int kind, IProject project, IBuilder[] builders, boolea
 			}
 
 			for (int i = 0; i < num; i++) {
+				CompilationDatabaseGenerator generator = new CompilationDatabaseGenerator();
```
jld01, line 508: Hi @alicetrifu, would it be reasonable to invoke the compilation database generator whenever changes to the managed build settings are applied, rather than having to perform a build?

Kummallinen: @jld01 this would need to be done on more than just MBS changes. Any new, renamed, moved or deleted file would also need to trigger it too. Potentially the generation could also take a noticeable period of time, especially on toolchains which make heavy use of complex applicability calculators & command generators

Generating on build is close to how it works with CMake projects.

Maybe something to investigate after the initial version is in?

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseContributionManager.java
```diff
@@ -0,0 +1,126 @@
...
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IConfigurationElement;
+import org.eclipse.core.runtime.IExtension;
+import org.eclipse.core.runtime.IExtensionPoint;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.jdt.annotation.NonNull;
+
+public class CompilationDatabaseContributionManager {
```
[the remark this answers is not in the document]
alicetrifu: Hello, could you please provide me with more details about the approach and logic you envision for this implementation? I'm not really clear on the direction I should take to accomplish this.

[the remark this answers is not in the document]
ruspl-afed: Initially suggested non-final singleton looks weird. 

> assuming that Platform is running and all the extensions are delivered before it is accessed

and this is also needs more work. 

Ideally, there should be an interface and an internal implementation, supplied with OSGi component.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/api/CompilationDatabaseInformation.java
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * The compilation database array of “command objects” members to be used for
+ * the extension point
+ * directory: The working directory of the compilation.
+ * command: The compile command. file: The main translation unit source
+ * processed by this compilation step.
+ */
+
+public class CompilationDatabaseInformation {
```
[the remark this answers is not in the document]
jonahgraham: A first use of `record` in CDT. I don't yet have enough experience with records to know if there are downsides here. I assume that the GSon serializer handles the record without issue.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/CommonBuilder.java
```diff
@@ -505,6 +506,8 @@ private IProject[] build(int kind, IProject project, IBuilder[] builders, boolea
 			}
 
 			for (int i = 0; i < num; i++) {
+				CompilationDatabaseGenerator generator = new CompilationDatabaseGenerator(getProject(), activeCfg);
```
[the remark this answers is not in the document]
jld01: @jonahgraham my thinking was based on how the compilation databases are consumed. Ideally, the editing experience that MBS users have grown accustomed to would not regress in any way. It seems that database generation that occurs only on build would be particularly noticeable immediately following project creation or project import. I am OK with hooking into the CDT builder in the first instance with a view to assessing performance and the viability of alternative hooks. I wonder if there is performance to be gained by observing the resource delta.

Regardless, this contribution is good news for CDT. I'm looking forward to seeing the database generator implementation. 👍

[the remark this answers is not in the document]
jonahgraham: > It seems that database generation that occurs only on build would be particularly noticeable immediately following project creation or project import.

Yes, I agree. The CMake flow suffers from this - i.e. immediately on project creation you end up with a project full of red underlines/problem markers.

Thank you for supporting deferring when/where this gets hooked in. Perhaps we can/should fix both the CMake and managed build at the same time by improving the project/import wizard. The other option is improving cdt-lsp to identify when compile_commands.json is missing and prompt to do a build that will then generate it?

@alicetrifu can you add this to a TODO list in #689?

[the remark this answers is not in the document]
Kummallinen: Would be good to look at the CMake timing too. There are also other features that would be useful o have that would benefit from a CMake configure being run in the background rather than only on build.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseContributionManager.java
```diff
@@ -0,0 +1,139 @@
...
+		@Override
+		public @NonNull List<CompilationDatabaseInformation> getAdditionalFiles(@NonNull IConfiguration config) {
+			return Collections.emptyList();
+		}
+
+	}
+
+	private Map<String, IConfigurationElement> factoryExtensions;
+	@NonNull
+	private Map<String, ICompilationDatabaseContributor> loadedInstances;
```
ruspl-afed, line 47: this one can be final

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseContributionManager.java
```diff
@@ -0,0 +1,139 @@
...
+	public class EmptyCompilationDatabaseContributor implements ICompilationDatabaseContributor {
+
+		@Override
+		public @NonNull List<CompilationDatabaseInformation> getAdditionalFiles(@NonNull IConfiguration config) {
+			return Collections.emptyList();
+		}
+
+	}
+
+	private Map<String, IConfigurationElement> factoryExtensions;
```
ruspl-afed, line 45: this one can be final

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseContributionManager.java
```diff
@@ -0,0 +1,139 @@
...
+import org.eclipse.core.runtime.Status;
+import org.eclipse.jdt.annotation.NonNull;
+
+public final class CompilationDatabaseContributionManager {
+	private static final String ATTRIB_RUNNER = "runner"; //$NON-NLS-1$
+	private static final String ATTRIB_TOOLCHAIN_ID = "toolchainID"; //$NON-NLS-1$
+	private static final String ID_COMPILATIONDATABASE = "compilationDatabase"; //$NON-NLS-1$
+	private static final String EXTENSION_ID = "compilationDatabaseContributor"; //$NON-NLS-1$
+
+	public class EmptyCompilationDatabaseContributor implements ICompilationDatabaseContributor {
```
ruspl-afed, line 36: this one can be private

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseContributionManager.java
```diff
@@ -0,0 +1,139 @@
...
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IConfigurationElement;
+import org.eclipse.core.runtime.IExtension;
+import org.eclipse.core.runtime.IExtensionPoint;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.jdt.annotation.NonNull;
+
+public final class CompilationDatabaseContributionManager {
```
ruspl-afed, line 30: if we reduce the visibility of this type to package local, concerns regarding its lifecycle became insignificant, because currently it is accessed relatively late in the lifecycle of the Platform.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -0,0 +1,419 @@
...
+import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Path;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.jdt.annotation.NonNull;
+
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+
+public class CompilationDatabaseGenerator {
```
ruspl-afed, line 57: Is it intended to be overridden? if yes, what are the "hooks" to override? Type, designed  to be overridden, should have a number of protected abstract methods.
If no - please make it final.

jonahgraham (commented):
The code cleanliness failed, but not because of this PR. Once I merge https://github.com/eclipse-cdt/cdt/pull/699 you will need to rebase this change.

build/org.eclipse.cdt.managedbuilder.core/META-INF/MANIFEST.MF
jonahgraham: I think things are working here.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseContributionManager.java
```diff
@@ -0,0 +1,137 @@
...
+import org.eclipse.core.runtime.Assert;
+import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IConfigurationElement;
+import org.eclipse.core.runtime.IExtension;
+import org.eclipse.core.runtime.IExtensionPoint;
+import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.jdt.annotation.NonNull;
+
+final class CompilationDatabaseContributionManager {
```
jonahgraham, line 30: I'm not a fan of package private, but I don't want to override the previous decision. As most cases it was unintentional, the quasi-standard of decorating manually in `/* package */` is a good idea.

```suggestion
/* package */ final class CompilationDatabaseContributionManager {
```

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseContributionManager.java
```diff
@@ -0,0 +1,137 @@
...
+			return Collections.emptyList();
+		}
+	}
+
+	private Map<String, IConfigurationElement> factoryExtensions;
+	private static CompilationDatabaseContributionManager instance;
+
+	private CompilationDatabaseContributionManager() {
+		factoryExtensions = new HashMap<>();
+		loadedInstances = new HashMap<>();
```
jonahgraham, line 51: I marked previous comments resolved around these fields because they weren't appearing on the latest patchset.

These fields lifecycle is unclear, made more complicated by the local variable with a similar name in initialise method.

Can these fields be final, assigned at declaration and have a clearer lifecycle? If my comment is unclear, let me know and I will provide a commit to explain it.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -0,0 +1,419 @@
...
+
+	/**
+	 * @param proj
+	 *            Creates and populates compilation database file
+	 * @param config
+	 */
+	public void generate() {
+
+		IPath buildDirectory = project.getFullPath()
+				.append(IPath.SEPARATOR + CompilationDatabaseGenerator.BUILD_FOLDER);
```
jonahgraham, line 82: Why are we always generating into a new folder called "build" instead of into the config directory?

jonahgraham (commented):
I have done a partial review - there are still some concerns I have on the generator that I need to look at a little closer.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseGenerator.java
alicetrifu: ClangD expects to find the file on a folder called "build", or in the project root. We chose to add it to the "build" folder for Renesas, but I guess this can be discussed more.

Kummallinen: If we're generating the .clangd file into the project root then we could move the json file into the config directory, using the .clangd to point to it. However the ".clangd" file is expected to be checked into source control. If we write a config specific path into .clangd then it would be changed whenever the user changed build configs. That action wouldn't be expected to impact files in source control.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseContributionManager.java
jonahgraham: This is part of what I meant - I will add a commit to this PR to explain explicitly what I mean.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseGenerator.java
jonahgraham: I wonder which way around is better here. We can put `compile_commands.json` in `Debug/compile_commands.json` and `Release/compile_commands.json` and update `.clangd` when the active configuration changes. Or we can put in `build/compile_commands.json`, and require a build to switch between configurations.

My initial thought was to put it in Debug/Release, but in light of the conversation above (https://github.com/eclipse-cdt/cdt/pull/692#discussion_r1479172539) requiring a build all the time to ensure an updated compile_commands.json may be the best for now.

Therefore I will leave this as is and resolved for this PR and add a comment to #689

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseContributionManager.java
jonahgraham: I added b6c8e30b7e7cd2dbeeeb00bd8d7ed56c86f26c3c which shows what I meant. It is fairly minor, but removes some code and gets rid of the extra temporary map.

jonahgraham (changes requested):
Not sure at what step we need to add some tests for this. In the meantime, this change (probably because we generate into "new" directory `build`) causes a test failure in `org.eclipse.cdt.managedbuilder.core.regressions.Bug_303953Test.testBuildAfterSourcefileDelete()` Please see [this section of the testing document](https://github.com/eclipse-cdt/cdt/blob/main/TESTING.md#how-do-i-run-cdt-junit-test-suite) for how to run these tests in the dev environment.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseGenerator.java
Kummallinen: @jonahgraham as my comment above. Updating .clangd on config changes is problematic as .clangd is supposed to be in source control according to clangd's documentation. Changing the active config would not be expected to impact source controlled files as it a user level preference.

jld01: If I have understood correctly, there is one instance of clangd which will look for `.clangd` in ancestor folders of the source file and in `build` folders under each of those ancestor folders. In that case, for MBS projects only, could we update `.clangd` in the project root to point to the configuration databse in the active build configuration folder and mark the `.clangd` file as a _derived resource_?

jonahgraham: Conversation moved to https://github.com/eclipse-cdt/cdt/issues/689#issuecomment-1942583000 and marking this as resolved.

jonahgraham:
Following our call yesterday we decided that a bit of work is needed on enabling the feature. Quoted from [minutes](https://hackmd.io/@jonahgraham/cdt-call-february-14-2024):

> One aspect where a decision agreed (with only minor reservations) by all in attendance is to enable the generator by a workspace preference that is disabled by default, with an API to allow CDT-LSP and other dependent projects (such as Renesas' internal editor) to enable it.

jonahgraham:
> Following our call yesterday we decided that a bit of work is needed on enabling the feature. Quoted from [minutes](https://hackmd.io/@jonahgraham/cdt-call-february-14-2024):
> 
> > One aspect where a decision agreed (with only minor reservations) by all in attendance is to enable the generator by a workspace preference that is disabled by default, with an API to allow CDT-LSP and other dependent projects (such as Renesas' internal editor) to enable it.

This has grown in scope - in light of [our call today](https://hackmd.io/@jonahgraham/cdt-call-february-21-2024) and the comment that @15knots made: https://github.com/eclipse-cdt/cdt/issues/689#issuecomment-1947333602 we do indeed need this to be settable with the default being the workspace setting. The workspace setting default should be off for now.

jonahgraham:
We are going to hold off merging this until after we branch. We plan to branch very soon, a little earlier than normal.

jonahgraham:
The conclusion of the [meeting today](https://hackmd.io/@jonahgraham/cdt-call-february-21-2024) is that the current location for generation (in CommonBuilder) is a reasonable short term solution. There may be better places in the future to keep the `compile_commands.json` more live, but that is deferred and won't hold up the merging of this change.

Similarly the placement of the output in `build/compile_commands.json` is a reasonable place for now, but during the meeting we made note that improving the flexibility in clangd may be best here. See also https://github.com/eclipse-cdt/cdt/issues/689#issuecomment-1948106737

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/core/jsoncdb/generator/CompilationDatabaseContributionManager.java
```diff
@@ -0,0 +1,139 @@
+/********************************************************************************
+ * Copyright (c) 2023, 2024 Renesas Electronics Corp. and others.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License 2.0 which is available at
+ * http://www.eclipse.org/legal/epl-2.0.
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ ********************************************************************************/
+package org.eclipse.cdt.managedbuilder.core.jsoncdb.generator;
```
Kummallinen, line 10: If this isn't being exported probably should name this package as an internal. so "org.eclipse.cdt.managedbuilder.internal.core.jsoncdb.generator"

alicetrifu:
Hello @jonahgraham , 
I've developed a small test suite for functionality validation. I aimed to verify that the functionality applies correctly to CPP projects as well, and to achieve this, I added a new project to the regression workspace and used that in my test. Unfortunately, this test is causing an error during the test execution process along with the build procedure, here on GitHub. 
```
On isCPPFileAllowed (org.eclipse.cdt.managedbuilder.core.tests.CompilationDatabaseGenerationTest) with error
test-results/build/org.eclipse.cdt.managedbuilder.core.tests/target/surefire-reports/TEST-org.eclipse.cdt.managedbuilder.core.tests.CompilationDatabaseGenerationTest.xml [t
I see the following error:
Contains: Errors running builder 'CDT Builder' on project 'helloworldCPP'.
java.lang.NullPointerException: Cannot invoke "org.eclipse.cdt.managedbuilder.core.IOutputType.getOutputExtensions(org.eclipse.cdt.managedbuilder.core.ITool)" because "outType" is null
```
It's worth mentioning that I never encountered this exception in the workspace, the output is always different from null. Do you have any suggestions on how I could address this issue, maybe? Currently, I've marked the test as skipped, but I believe this should be investigated further in the future.

build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/core/tests/CompilationDatabaseGenerationTest.java
```diff
@@ -0,0 +1,189 @@
...
+		}
+
+	}
+
+	/**
+	 * Tests that cpp files are handled by compile_commands.json file generator
+	 * @throws CoreException
+	 */
+	@Test
+	@Ignore("This will be temporary skipped due to builder error")
```
jonahgraham, line 134: The difference between what you see in the workspace and what happens on the build machine is probably down to different sets of enabled bundles. See https://github.com/eclipse-cdt/cdt/blob/main/TESTING.md#using-limited-sets-of-plug-ins-in-tests (point 3)

jonahgraham (commented):
This looks very close to me. I am about to merge in 11.6 version bump now that I have branched for 11.5. I'll try rebasing and see if I can reproduce the difference.

jonahgraham (commented):
The preferences page does not work. The performApply calling itself makes it fail catastrophically, but my attempt at the simple fix did not resolve the issue.

I made comments in the PR for most of the changes I made. I'll mark them as resolved, but wanted you to know why I changed what I did.

build/org.eclipse.cdt.managedbuilder.core.tests/resources/builderTests/regressions/helloworldCPP.zip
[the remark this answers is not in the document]
jonahgraham: @alicetrifu this is the comment that most addressed your test issue. You can compare the before and after zip files.

build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/core/tests/CompilationDatabaseGenerationTest.java
```diff
@@ -0,0 +1,189 @@
...
+		app.build(IncrementalProjectBuilder.FULL_BUILD, null);
+		IFile compilationDatabase = app.getFile("build/compile_commands.json");
+		assertFalse(compilationDatabase.exists());
+	}
+
+	public static boolean isGenerateFileOptionEnabled(boolean value) {
+		try {
+			IPreferenceStore preferenceStore = new ScopedPreferenceStore(InstanceScope.INSTANCE,
+					"org.eclipse.cdt.managedbuilder.ui"); //$NON-NLS-1$
+			preferenceStore.setDefault("generateFile", value);
```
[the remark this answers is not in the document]
jonahgraham: I made this change, but the new code needs to be checked that it is correct.

build/org.eclipse.cdt.managedbuilder.ui/src/org/eclipse/cdt/managedbuilder/internal/ui/compilationdatabase/CompilationDatabaseGeneratorBlock.java
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	@Override
+	public void performDefaults() {
+		preferenceStore.setDefault(ENABLE_FILE_GENERATION, false);
+	}
+
+	@Override
+	public void performApply(IProgressMonitor monitor) throws CoreException {
+		this.performApply(monitor);
```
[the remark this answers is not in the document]
jonahgraham: I made this change, but the new code needs to be checked that it is correct.

build/org.eclipse.cdt.managedbuilder.ui/src/org/eclipse/cdt/managedbuilder/internal/ui/compilationdatabase/CompilationDatabaseGeneratorBlock.java
```diff
@@ -0,0 +1,100 @@
...
+		Group cdbGeneratorOptions = new Group(composite, SWT.NONE);
+		cdbGeneratorOptions.setLayout(new FillLayout(SWT.HORIZONTAL));
+		cdbGeneratorOptions.setText(Messages.JsonCdbGeneratorPreferencePage_description);
+		cdbGeneratorOptions.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, false));
+		generateFileCheckbox = new Button(cdbGeneratorOptions, SWT.CHECK);
+		generateFileCheckbox.setSelection(preferenceStore.getBoolean(ENABLE_FILE_GENERATION));
+		generateFileCheckbox.setText(Messages.JsonCdbGeneratorPreferencePage_generateCompilationdatabase);
+		generateFileCheckbox.addListener(SWT.Selection, e -> {
+			boolean newValue = generateFileCheckbox.getSelection();
+			preferenceStore.setValue(ENABLE_FILE_GENERATION, newValue);
```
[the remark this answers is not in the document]
jonahgraham: I made this change, but the new code needs to be checked that it is correct.

jonahgraham:
Seeing as this implementation the preferences don't work, some additional time is needed here. 

I spent some time trying to replicate how @ghentschke and @ruspl-afed implemented preferences for cdt-lsp (e.g. [EditorOptions](https://github.com/eclipse-cdt/cdt-lsp/blob/156e8d8df2956b62494bc639bae2da93ed78cbd3/bundles/org.eclipse.cdt.lsp/src/org/eclipse/cdt/lsp/editor/EditorOptions.java) and related classes). That implementation is a little complicated, but does allow all the configuration by products needed. (Unfortunately I ran out of time to do this before I went on vacation)

jonahgraham (approved):

jonahgraham:
@ghentschke @Kummallinen @alicetrifu 

I added this additional commit (and squashed the rest)

Generate compile_commands.json into standard build directory. That it is the file ends up in 
 Debug/compile_commands.json instead of build/compile_commands.json. This lines up the behaviour of the generator with that of the code that generates a .clangd file in [ClangdConfigurationFileManager.java](https://github.com/eclipse-cdt/cdt-lsp/blob/ee297ee0ced1747352039e6ee25bcc80c10c7462/bundles/org.eclipse.cdt.lsp.clangd/src/org/eclipse/cdt/lsp/clangd/internal/config/ClangdConfigurationFileManager.java) This code already supports vendors overriding to supply their own (or extended) .clangd contents.

jonahgraham:
Assuming no objections to https://github.com/eclipse-cdt/cdt/pull/692#issuecomment-2113366001 and a clean build, this is ready to merge.

jonahgraham:
> Assuming no objections to https://github.com/eclipse-cdt/cdt/pull/692#issuecomment-2113366001 and a clean build, this is ready to merge.

I discussed this with William this morning and he/Renesas are happy to take advantage of the API that @ghentschke has provided in ClangdConfigurationFileManager to handle updating `.clangd` if and as needed.

Therefore I am merging this now and we can start creating new issues as needed too.

PR #694, opened 2024-02-06
[#693] Fix resource leak warnings for org.eclipse.cdt.core.utils

### LINKED ISSUE #693: Fix resource leak warnings for org.eclipse.cdt.core.utils
We have several "resource leak" warnings for `Dwarf` and `MachO64` types

It would be good to fix them

ruspl-afed:
* use `try-with-resources` for `AutoCloseable`

Fixes #693

### DIFF
--- a/core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/debug/dwarf/Dwarf.java
+++ b/core/org.eclipse.cdt.core/utils/org/eclipse/cdt/utils/debug/dwarf/Dwarf.java
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2000, 2023 QNX Software Systems and others.
+ * Copyright (c) 2000, 2024 QNX Software Systems and others.
  *
  * This program and the accompanying materials
  * are made available under the terms of the Eclipse Public License 2.0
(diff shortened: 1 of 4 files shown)


ruspl-afed:
@jld01 we have more hidden resource leak issues near `Dwarf` and related types when exposing them via `ISymbolReader` adapter: it is really not clear when to close them properly. Idea to override `finalize` is not good for Java, since the usage of this method was deprecated long ago. But these seem to be out of scope here, since most probably the proper the fix will cause breaking API changes.

jld01 (commented):
These changes look positive, @ruspl-afed, but I'm struggling to understand why the `dispose()` methods are necessary in the first place. For example, the `Dwarf` section maps should fall out of scope with each `Dwarf` instance. Do you have any insight into what is preventing the garbage collector from doing its job here?

jonahgraham (approved):
The history of why it needs to be autocloseable and not wait until gc runs is in https://github.com/eclipse-cdt/cdt/pull/132

But the short answer is that all these classes hold onto real files so waiting until gc is too late.

jld01:
Thanks for the explanation, @jonahgraham. 👍

jld01 (approved):

Issue #700, opened 2024-02-08
Platform APIs have been removed breaking CDT build on latest platform dependencies


jonahgraham:
org.eclipse.cdt.make.ui.dialogs.DiscoveredPathContainerPage contains references to lots of previously deprecated and not deleted code in Eclipse Platform.

See https://github.com/eclipse-platform/eclipse.platform.ui/pull/1563 and https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/1748

PR #701, opened 2024-02-08
Remove DiscoveredPathContainerPage

jonahgraham:
The class DiscoveredPathContainerPage used a number of APIs in the Eclipse Platform [removed](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1563) in [Eclipse 4.31 release](https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/blob/81e406456ffae2f82fe0bb244adfdc4121c9e463/eclipse.platform.common/bundles/org.eclipse.platform.doc.isv/porting/removals.html#L595). This class, while API, has not be used, nor usable since the CDT 4 release as it was only for CDT 3.x style projects. The class had been deprecated since 2010.

To mitigate against the possibility that someone may have a dependency on this old class the minor version has been bumped so that version range can have `,8.3)` as their upper version.

Fixes https://github.com/eclipse-cdt/cdt/issues/700


jonahgraham:
@eclipse-cdt/tools-cdt-committers can a committer review this and provide a second approval to removing this class.

15knots:
+2

build/org.eclipse.cdt.make.ui/plugin.xml
```diff
@@ -430,14 +430,6 @@
    </extension>
 
 
-   <extension
-         point="org.eclipse.cdt.ui.PathContainerPage">
-      <PathContainerPage
-            class="org.eclipse.cdt.make.ui.dialogs.DiscoveredPathContainerPage"
-            icon="icons/obj16/lib_obj.gif"
-            id="org.eclipse.cdt.make.core.DISCOVERED_SCANNER_INFO"
-            name="%DiscoveredScannerInfoContainer.name"/>
```
ruspl-afed, line 439: This key needs deletion as well

ruspl-afed (commented):
You are very fast :)

And what should happen with `org.eclipse.cdt.make.internal.core.scannerconfig.DiscoveredPathInitializer`?

build/org.eclipse.cdt.make.ui/plugin.xml
jonahgraham: https://github.com/eclipse-cdt/cdt/pull/702

PR #702, opened 2024-02-08
Remove unreferenced properties key

jonahgraham:
Fixes https://github.com/eclipse-cdt/cdt/pull/701/files#r1483534277


ruspl-afed (approved):
Thanks!

PR #717, opened 2024-02-23
Fix #716: Avoid varying order of elements in project explorer

### LINKED ISSUE #716: Element order in project explorer arbitrarily changes if label decoration is involved
**The observed issue**
The order of elements in the project explorer differs from time to time, for example, if we have a folder under version control (git) and some of the contained files are modified, but not committed. The order should not change arbitrarily.

**Steps to reproduce**
1. Switch to the C++ perspective
2. Take any C++ project under version control (e.g. git) with empty git status (no uncommitted changes)
3. Open any folder in that project in the project explorer (list the children)
4. Modify any file in that folder, leave the other files unchanged
5. Select the folder in the project explorer
6. Hit F5 key multiple times, with only short delays of some ms (triggers refresh operation)
7. See how the modified files sometimes are list first (because of the "> " prefix) and sometimes are listed in their natural order (ordered by file name)

**Expected behavior**
The files should always stay ordered using the file name, even if they are decorated with e.g. a "> " prefix.

**Screenshots**
See an example of differing element order in the project explorer:

Order 1:
image

Order 2:
image

**Version Information**
 - OS: RHEL 7
 - Custom-build Eclipse IDE version 4.30
 - CDT version 11.4.0

**Additional context**
I narrowed the issue down to `org.eclipse.cdt.ui.CElementSorter#compareWithLabelProvider(Viewer, Object, Object)`. Here, for elements from categories resources, resource folders, storage, and others the label provider is used to calculate the file's labels for comparison. The decoration is done asynchronously. If there is no cached decoration available, no decoration is used, but the decoration calculation is triggered. I think, that's why the labels returned by the label provider for the modified files sometimes come with the prefix "> " and sometimes without that prefix. That difference leads to differing order in the project explorer.

I took a look at JDT, i.e. `org.eclipse.jdt.ui.JavaElementComparator#compare(Viewer, Object, Object)`. Here, instead of using the label providers an label is calculated without decorating the `IResource`'s name, see `org.eclipse.cdt.ui.CElementSorter#getNonJavaElementLabel(Viewer, Object)`. I gues, it's a good idea to do the same in `CElementSorter`.

travkin79:
Fixes #716.


jonahgraham:
Build failed with "Error: Exception in thread "pool-131-thread-1" java.lang.OutOfMemoryError: Java heap space" - retriggering.

jonahgraham (approved):
Approved - subject to code cleanliness completing successfully.

jonahgraham:
@travkin79 I see that this commit is against an old commit - I'll rebase it it this time, but if you can submit against a more recent commit next time that would be great.

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/ui/CElementSorter.java
```diff
@@ -427,20 +432,28 @@ private ISourceRoot getSourceRoot(Object element) {
...
+	// implementation taken from org.eclipse.jdt.ui.JavaElementComparator
+	private String getNonCElementLabel(Viewer viewer, Object element) {
+		// try to use the workbench adapter for non - C++ resources or if not available, use the viewers label provider
+		if (element instanceof IResource) {
+			return ((IResource) element).getName();
+		}
+		if (element instanceof IStorage) {
+			return ((IStorage) element).getName();
+		}
+		if (element instanceof IAdaptable) {
```
ruspl-afed, line 444: Here it is better to use `Adapters#adapt`, because a lot of elements adapted "externally", with `IAdapterFactory`

jonahgraham: I prefer consistency with JDT if possible. See also https://github.com/eclipse-cdt/cdt/issues/716#issuecomment-1961552164

ruspl-afed: > Please let me know if matching JDT is a problem here.

The problem _may_ appear for element contributed by 3rd party, but currently I'm not aware of any.

ruspl-afed: Anyway, this was a generic recommendation regarding Platform usage, should not be considered as an obstacle to merge this one.

jonahgraham: +1 I agree - new code should definitely use Adapters#adapt.

Issue #719, opened 2024-02-28
Add required root files


ruspl-afed:
- [x] NOTICE.md
- [x] SECURITY.md

ruspl-afed:
Done with #720

jonahgraham:
I think changing the security policy for the project should get some approval from the committer community first.

ruspl-afed:
> I think changing the security policy for the project should get some approval from the committer community first.

Understood. Let me remove it and propose a PR with SECURITY.md

PR #720, opened 2024-02-28
[#719] Add required root files

ruspl-afed:
* add NOTICE.md
* add SECURITY.md

with dedicated commits to process independently if needed

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,20 @@
+# Security Policy
+
+This project implements the Eclipse Foundation Security Policy
+
+* https://www.eclipse.org/security
+
+## Supported Versions
+
+These versions of C/C++ Development Tools (CDT) are currently being supported
+with security updates.
+
+| Version           | Released   | Supported | 
+| ----------------- | ---------- | --------- | 
+| 11.4.0            | 2023-12-06 | Yes       | 
+| < 11.3.0          | 2023-09-13 | No        | 
+
+## Reporting a Vulnerability
+
+Please report vulnerabilities to the Eclipse Foundation Security Team at
+security@eclipse.org
--- a/NOTICE.md
+++ b/NOTICE.md
@@ -1,0 +1,42 @@
+# Notices for Eclipse C/C++ Development Tools
+
+This content is produced and maintained by the Eclipse C/C++ Development Tools
+project.
+
+* Project home: https://projects.eclipse.org/projects/tools.cdt
+
+## Trademarks
+
+CDT™ is a trademark of the Eclipse Foundation.
+
+## Copyright
+
+All content is the property of the respective authors or their employers. For
+more information regarding authorship of content, please consult the listed
+source code repository logs.
+
+## Declared Project Licenses
+
+This program and the accompanying materials are made available under the terms
+of the Eclipse Public License v. 2.0 which is available at
+https://www.eclipse.org/legal/epl-2.0.
+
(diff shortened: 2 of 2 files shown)


ruspl-afed:
> Error assembling JAR: Problem creating jar: Execution exception: Java heap space

Are we really that big?

SECURITY.md
```diff
@@ -0,0 +1,20 @@
...
+* https://www.eclipse.org/security
+
+## Supported Versions
+
+These versions of C/C++ Development Tools (CDT) are currently being supported
+with security updates.
+
+| Version           | Released   | Supported | 
+| ----------------- | ---------- | --------- | 
+| 11.4.0            | 2023-12-06 | Yes       |
```
jonahgraham, line 14: I don't like this :-( I think such a broad commitment as supporting security updates for old releases needs broader committer acceptance. Other projects, in particular eclipse platform, don't do this.

ruspl-afed: Yes, there is a lot of room for different interpretations. Let's have a look at [EF Project Handbook](https://www.eclipse.org/projects/handbook/#legaldoc-security)

> The SECURITY file describes products and product versions that for which security updates may be applied and the means by which vulnerabilities are reported.

Personally, I do not see "broad commitment as supporting security updates for old releases" here. It **may be applied**.

Of cause, I am not a native speaker and may miss some layers of the message, but I prefer to use the meaning of MAY from [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119#section-5):

> [5](https://datatracker.ietf.org/doc/html/rfc2119#section-5). MAY   This word, or the adjective "OPTIONAL", mean that an item is
   truly optional.

However, I may be wrong. So, let's try to ask the author. Dear @waynebeaton , please help us to make it clear.

And I agree, we miss the EOL policy. Perhaps, this is something that you can discuss on Planning Council @jonahgraham

jonahgraham: To me the EF handbook says that we _may_ choose which versions to support - but saying in SECURITY.MD that 11.4.0 is supported says that we are supporting security releases on it, and I don't think that is accurate statement. In fact I think we should be explicit that we do *not* support security releases on released versions and that security fixes will come out with the next scheduled release.

I think the CDT security policy should say something along the lines of:

`Security updates are applied to the HEAD of the main branch.`  (wording adapted from https://github.com/eclipse/dash-licenses/blob/master/SECURITY.md)

If we do decide to support released versions for security updates (like [Jetty does](https://github.com/jetty/jetty.project/blob/jetty-12.0.x/SECURITY.md)) then I think the security policy should not have the explicit version numbers in the file. e.g. Jetty says:

> All [stable versions](https://eclipse.dev/jetty/download.php) of jetty are actively supported for security issues. [Deprecated versions](https://eclipse.dev/jetty/download.php) may be supported for serious security issues or on a commercial support basis.

(PS Probably should move this conversation out of a closed PR for visibility)

ruspl-afed: > (PS Probably should move this conversation out of a closed PR for visibility)

Yes, please take this to the place that is best suited for this discussion @jonahgraham.

waynebeaton: The EMO currently regards this as a "should do". In our current climate of increased focus on supply chain integrity, the "should to" is evolving into a "must do" from a community expectations perspective. Thank you for getting ahead of this an making the project's policy explicit.

Issue #634, opened 2023-12-05
Closing a CDT project somethimes reopens it on the spot.


jantje:
While I was testing the persistency of my autobuild plugin by opening and closing the project I noticed that sometimes the project gets loaded at (or around) closing time.
When this happened the project looked closed but when I open it eclipse did not trigger my loading code as once should expect during opening of a project.
The stacktrace is pretty big with lots of repititions (I still thought it was best to dump the whole thing)
I'm not sure this is a CDT issue but adding a "is the project open" test to the isNewStyleProject implementations or even to the getProjectDescription may be a good workaround/solution.

```
org.eclipse.core.internal.resources.ResourceException(/001_all_gnu_cross_shared_lib_cdt_cross_gnu)[372]: java.lang.Exception: Project '001_all_gnu_cross_shared_lib_cdt_cross_gnu' is not open.
	at org.eclipse.core.internal.resources.ResourceException.provideStackTrace(ResourceException.java:42)
	at org.eclipse.core.internal.resources.ResourceException.<init>(ResourceException.java:38)
	at org.eclipse.core.internal.resources.Project.checkAccessible(Project.java:186)
	at org.eclipse.core.internal.resources.Project.hasNature(Project.java:564)
	at io.sloeber.schema.internal.Tool.isEnabled(Tool.java:143)
	at io.sloeber.autoBuild.integration.AutoBuildConfigurationDescription.options_updateDefault(AutoBuildConfigurationDescription.java:557)
	at io.sloeber.autoBuild.integration.AutoBuildConfigurationDescription.<init>(AutoBuildConfigurationDescription.java:429)
	at io.sloeber.autoBuild.integration.AutoBuildConfigurationDescriptionProvider.loadConfiguration(AutoBuildConfigurationDescriptionProvider.java:98)
	at org.eclipse.cdt.internal.core.settings.model.CConfigurationDescriptionCache.loadData(CConfigurationDescriptionCache.java:139)
	at org.eclipse.cdt.internal.core.settings.model.CProjectDescription.loadDatas(CProjectDescription.java:199)
	at org.eclipse.cdt.internal.core.settings.model.xml.XmlProjectDescriptionStorage.loadProjectDescription(XmlProjectDescriptionStorage.java:509)
	at org.eclipse.cdt.internal.core.settings.model.xml.XmlProjectDescriptionStorage.getProjectDescription(XmlProjectDescriptionStorage.java:239)
[...]

PR #725, opened 2024-03-04
fix for issue #634 closing project automatically loads the CDT project

jantje:
isNewStyleProject called by gui loads closed CDT projects

### DIFF
--- a/core/org.eclipse.cdt.core/model/org/eclipse/cdt/internal/core/settings/model/CProjectDescriptionManager.java
+++ b/core/org.eclipse.cdt.core/model/org/eclipse/cdt/internal/core/settings/model/CProjectDescriptionManager.java
@@ -2461,7 +2461,8 @@ public class CProjectDescriptionManager implements ICProjectDescriptionManager {
 
 	@Override
 	public boolean isNewStyleProject(IProject project) {
-		return isNewStyleProject(getProjectDescription(project, false));
+		System.out.println("calling isNewStyleProject for project " + project.getName());
+		return isNewStyleProject(getProjectDescription(project, false, false));
 	}
 
 	@Override


ruspl-afed:
The proposed change may have much more impact than desired, because for other use cases client code already expects the current behavior of `ICProjectDescriptionManager`.

Please consider the fix near `org.eclipse.cdt.internal.ui.workingsets.HasManagedCdtProjectSelection.hasManagedCdtProjectSelection(HasManagedCdtProjectSelection.java:43)` where context may be enough to understand that we don't need to load closed projects.

Also, at least ECA check should became happy, PR cannot be processed further in the current state.

jonahgraham:
This comment in #634:

> Because getProjectDescription(project, false) translates to getProjectDescription(project, true, false) which means "load the project when it is not yet loaded" all closed CDT projects are loaded when they are selected in the project explorer.

This is a (at least sometimes) incorrect conclusion. If a project is closed, then `getProjectDescription(project, true, false)` eventually calls `getProjectDescriptionStorage` which checks if the project is accessible, and because the project is closed it is not accessible, so the project is not loaded and a core exception is thrown:

https://github.com/eclipse-cdt/cdt/blob/c7169b3186d2fef20f97467c3e2ad78e2943ed1b/core/org.eclipse.cdt.core/model/org/eclipse/cdt/internal/core/settings/model/CProjectDescriptionManager.java#L886-L890

that exception is caught and a null is returned for project description:

https://github.com/eclipse-cdt/cdt/blob/c7169b3186d2fef20f97467c3e2ad78e2943ed1b/core/org.eclipse.cdt.core/model/org/eclipse/cdt/internal/core/settings/model/CProjectDescriptionManager.java#L409-L414

---

I'll add a comment to the issue too in case I have any useful insight.

jonahgraham:
I have converted this to draft until it is ready for a new review.

jonahgraham:
I am closing this old PR. Please feel free to re-open/comment if further work is planned here.

PR #729, opened 2024-03-07
Prep repo for 11.6.0

jonahgraham:


jonahgraham:
A while bunch of `org.eclipse.cdt.ui.tests.text.MarkOccurrenceTest` tests failed. It may be the update to the newest platform, so I will defer that to a separate PR.

jonahgraham:
> A while bunch of `org.eclipse.cdt.ui.tests.text.MarkOccurrenceTest` tests failed. It may be the update to the newest platform, so I will defer that to a separate PR.

This looks like a regression that is very similar to this reported regression:

- https://github.com/eclipse-jdt/eclipse.jdt.ui/issues/1236

There is already a fix for this merged to platform, so we'll see if tonights platform I-build resolves it. The original issue causing the regression is:

- https://github.com/eclipse-platform/eclipse.platform.ui/pull/1703

PR #748, opened 2024-03-29
Add GNU assembler debug level option to N&N

### LINKED ISSUE #727: MBS should generate debug info when building GNU assembly language files
When debugging MBS projects containing GNU assembly language files, it is not possible to single-step within assembly language source files by default. The GNU toolchain description should be modified to invoke the GNU assembler with the `-g` flag to generate debug info within debug build configurations (as for the GNU C/C++ complier).

jld01:
Resolves #727


jld01:
Checks are failing because linuxtools now depends on Java 21: https://github.com/eclipse-linuxtools/org.eclipse.linuxtools/pull/345

Cc: @jonahgraham

ruspl-afed:
> linuxtools now depends on Java 21

Good item for CDT meeting

PR #760, opened 2024-04-19
Update Java version to 21

ghentschke:
to fix the build errors

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -21,10 +21,10 @@ jobs:
             - 'dsf/**'
             - 'debug/**'
             - 'jtag/**'
-    - name: Set up JDK 17
+    - name: Set up JDK 21
       uses: actions/setup-java@v3
       with:
-        java-version: '17'
+        java-version: '21'
         distribution: 'temurin'
         cache: maven
     - name: Set up Maven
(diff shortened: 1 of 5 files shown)


jonahgraham:
If/when the build passes this is good for me to go.

ghentschke:
Hm, it does not build. Does anyone has an idea to fix this: 
`Error:  Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.1.0:run (generate-configuration) on project org.eclipse.cdt.debug.application: An Ant BuildException has occured: The following error occurred while executing this line:
Error:  /home/runner/work/cdt/cdt/debug/org.eclipse.cdt.debug.application/configGenerator.xml:4: java.lang.UnsupportedOperationException: The Security Manager is deprecated and will be removed in a future release
Error:  around Ant part ...<ant antfile="configGenerator.xml" target="generate.configuration" />... @ 6:74 in /home/runner/work/cdt/cdt/debug/org.eclipse.cdt.debug.application/target/antrun/build-main.xml
Error:  -> [Help 1]`

ghentschke:
I think we have to get rid of the security manager since its an error in Java 21.

ruspl-afed:
in this case we need to rework from ant call to something better

ruspl-afed:
> in this case we need to rework from ant call to something better

Unfortunately it is possible to do only with 17-> 21 upgrade and some other changes, see https://github.com/eclipse-cdt/cdt/pull/761

jonahgraham:
With #761 these improvements are all covered and the build works on Java 21. Thanks @ghentschke for your efforts here.

There is a part of this change that can/should be applied if done fully, that is removing use of security manager.

PR #768, opened 2024-04-27
Make CodanRunner progress monitor work a bit better

MarkZ3:
The progress monitor didn't show which file was being analyzed properly.

Also, the work units didn't seem to be counted in a good way because the progress bar would stay stuck at like 1%. The fix in this commit for that is not perfect as it gives equal weight to all resources on a same level until further split into sub monitors. It still works quite a bit better without requiring more heavy changes.

Also fix usage of deprecated SubProgressMonitor.


jonahgraham:
Thanks @MarkZ3

PR #770, opened 2024-04-28
Parse empty __attribute__(())

MarkZ3:
This is accepted by both GCC and Clang and encountered in real code. It looks a bit strange but basically a condtionally defined macro was used inside the (()) depending on a configurable feature macro of the project.


jonahgraham:
Thanks @MarkZ3

PR #779, opened 2024-05-10
Fix missing mylyn.wikitext dependency

MarkZ3:
lsp4e depends on mylyn.wikitext but it could not be found. For example: https://github.com/eclipse-cdt/cdt/actions/runs/9026888723/job/24804874235?pr=778

### DIFF
--- a/releng/org.eclipse.cdt.target/cdt.target
+++ b/releng/org.eclipse.cdt.target/cdt.target
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <?pde version="3.8"?>
-<target name="cdt" sequenceNumber="157">
+<target name="cdt" sequenceNumber="158">
 	<locations>
 		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
 			<repository location="https://download.eclipse.org/cbi/updates/license/" />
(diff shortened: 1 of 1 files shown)


releng/org.eclipse.cdt.target/cdt.target
```diff
@@ -14,6 +14,9 @@
 			<unit id="org.eclipse.test.feature.group" version="0.0.0" />
 			<unit id="org.eclipse.unittest.ui" version="0.0.0" />
 		</location>
+		<location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+			<repository location="https://download.eclipse.org/mylyn/updates/release/latest/" />
```
ruspl-afed, line 18: @MarkZ3 SimRel version of Mylyn WikiText
Since LSP4E is full of innovations I wonder why they still don't upgrade Mylyn WikiText.

akurtakov: That's an easy answer - people doing and caring for releng are so few and spread over so many projects that such things are easily missed. Best would be if PRs are provided for such changes wherever needed.
I have to admit that with lsp4e being with very few deps I manage to find time for updating its deps only few times per year (and usually this is driven by LSP4J ).

jonahgraham:
I assume build is failing because of https://github.com/eclipse-cdt/cdt/issues/780 - I'll try to solve the immediate issue soon.

releng/org.eclipse.cdt.target/cdt.target
MarkZ3: > @MarkZ3 SimRel version of Mylyn WikiText

Sorry is this a question or suggestion?

jonahgraham:
I added this commit to #782 to get a clean build.

PR #784, opened 2024-05-13
Update direct-from-maven dependencies to latest available minor versions

merks:
Use orbit-aggregation/2024-06.


merks:
As a reminder, there is this report:

https://github.com/eclipse-orbit/orbit-simrel/blob/main/report/maven-osgi/cdt/REPORT.md

And this commit is effectively just a copy-and-paste (other than the aggregation update) of this:

https://raw.githubusercontent.com/eclipse-orbit/orbit-simrel/main/report/maven-osgi/cdt/updated.target

PR #798, opened 2024-05-21
Update gson to 2.11.0

merks:


jonahgraham:
I'm just working on the M3 contribution - I'll see if I can get this in. However the reason that some dependencies are tight on version range is because in the past gson have changed API in a breaking way without a major version bump. Have you  run something like japicmp on 2.10 and 2.11? If not, I will do that as part of this release.

However as LSP4J is much more sensitive to gson versions, I will do it there first - see https://github.com/eclipse-lsp4j/lsp4j/issues/836

merks:
I don’t have the capacity to test orbit bundles. 

I just enhanced the orbit maven generator so that it updates the file in a local git repository so I can more easily create a PR. 

Other projects like the Platform have already updated.

jonahgraham:
> I don’t have the capacity to test orbit bundles.

:-) Didn't expect you to - just making sure I wasn't repeating work

> I just enhanced the orbit maven generator so that it updates the file in a local git repository so I can more easily create a PR.

Awesome - next step will be a bot :-)

> Other projects like the Platform have already updated.

I think CDT will be fine, but I don't think newer LSP4J will end up in SimRel without https://github.com/eclipse-lsp4j/lsp4j/issues/836

PR #803, opened 2024-05-29
LLVM optimization remarks.

### LINKED ISSUE #752: Error parser for LLVM Clang
As far as I know there is no error parser for LLVM clang. I could not find it. Using the default GNU gcc/g++ error parser results in to false errors during the build step. For instance a clang "remark:" message is categorized as an error by the GNU gcc/g++ error parser.

To reproduce compile this small C++ program, using clang, with CFLAGS = -O2 -Rpass=inline

```
int foo(int a, int b) {
	return a + b;
}

int bar(int j) {
	return foo(j, j - 2);
}

int main(int argc, char **argv) {
	return 0;
}
```

Results in:

```
Building in: hello/build/make.run.linux.x86_64
make -f ../../Makefile
clang++ -c -O2 -Rpass=inline   -o hello.o hello/hello.cpp
hello/hello.cpp:6:9: remark: _Z3fooii inlined into _Z3bari with cost=-35 (threshold=337) [-Rpass=inline]
        return foo(j, j - 2);
               ^
clang++  -o hello hello.o
Build complete (1 errors, 0 warnings): hello/build/make.run.linux.x86_64

```

I would like to have an error parser for LLVM clang.

I would also like to know what alternative I have programmatically. Would it require add a complete new error parser plugin? Or is there an easy option to insert a regex via a plugin to the GNU gcc/g++ error parser?

The context is a CDT Core Build Makefile project based IDE, using LLVM clang based toolchain.

ewaterlander:
Added support for LLVM optimization remarks to the GCC c/c++ error parser. They were falsly treated as errors (OtherError). Now they are treated as generic infos. GCC has no diagnostic messages of 'remark' type. Since GCC and LLVM messages are very similar, no separate LLVM error parser was created.

Fixes #752

### DIFF
--- a/core/org.eclipse.cdt.core/META-INF/MANIFEST.MF
+++ b/core/org.eclipse.cdt.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.cdt.core; singleton:=true
-Bundle-Version: 8.4.100.qualifier
+Bundle-Version: 8.4.200.qualifier
 Bundle-Activator: org.eclipse.cdt.core.CCorePlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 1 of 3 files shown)


jonahgraham (approved):

jonahgraham:
Thanks @ewaterlander

PR #804, opened 2024-05-29
Explicitly set application id for standalone debugger

jonahgraham:
Follow up for #782 to always explicitly set application

This wasn't needed in my development environment, but in the EPP version it was needed. I think it is because the normal config.ini specifies the application:

```sh
$ grep application eclipse/configuration/config.ini 
eclipse.application=org.eclipse.ui.ide.workbench
```

now that we don't define our own config.ini (see #782) we need to explicitly override the one set in the config.ini (if present)


ruspl-afed (approved):
LGTM

jonahgraham:
The test fails don't have to do with this change - I will review them on main if they continue to fail after merging.

PR #805, opened 2024-05-30
Remove 'Experimental' label from Core Build Makefile projects.

ewaterlander:
Removed the 'Experimental' label in project creation wizard for Core Build Makefile projects. Seven years after the inception it's time to remove it. Core Build Makefile is now in a usable state. The 'Experimental' label scares users away from it, making them fall back to the old Managed Build Makefile projects.

### DIFF
--- a/build/org.eclipse.cdt.make.ui/src/org/eclipse/cdt/make/internal/ui/wizards/NewMakefileProjectWizard.java
+++ b/build/org.eclipse.cdt.make.ui/src/org/eclipse/cdt/make/internal/ui/wizards/NewMakefileProjectWizard.java
@@ -50,7 +50,7 @@ public class NewMakefileProjectWizard extends TemplateWizard {
 
 				Button genSourceButton = new Button(buttonComp, SWT.CHECK);
 				genSourceButton.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
-				genSourceButton.setText("Generate Source and Makefile");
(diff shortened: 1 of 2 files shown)


ewaterlander:
I can add that at Intel we built a custom flow on top of Core Build Makefile which is working fine.

build/org.eclipse.cdt.make.ui/src/org/eclipse/cdt/make/internal/ui/wizards/NewMakefileProjectWizard.java
```diff
@@ -50,7 +50,7 @@ public void createControl(Composite parent) {
 
 				Button genSourceButton = new Button(buttonComp, SWT.CHECK);
 				genSourceButton.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
-				genSourceButton.setText("Generate Source and Makefile");
+				genSourceButton.setText("Create Hello World Source and Makefile example"); //$NON-NLS-1$
```
jonahgraham, line 53: This string should be translatable, so until it is lets leave off the non-translated string comment.

```suggestion
				genSourceButton.setText("Create Hello World Source and Makefile example");
```

jonahgraham (commented):
Other than minor fix, lets go for this. Thank you for moving this forward!

jonahgraham (approved):

PR #809, opened 2024-05-31
[#808] Support overriden default-editor-association in EditorUtility

### LINKED ISSUE #808: Support overriden default editor association in EditorUtility.getEditorID
Due to this [change #1665](https://github.com/eclipse-platform/eclipse.platform.ui/pull/1665) in Eclipse platform, the editor association for external C/C++ files has been broken when the new LSP based C/C++ editor is used (see this [cdt-lsp issue #338](https://github.com/eclipse-cdt/cdt-lsp/issues/338)) and it's set as default editor.

Note: the association is only broken for external files which resides outside of the workspace.

**Describe the solution you'd like**
This can be fixed by supporting `ExternalEditorInput` in `org.eclipse.cdt.internal.ui.util.EditorUtility`

This fixes the [cdt-lsp issue #338](https://github.com/eclipse-cdt/cdt-lsp/issues/338)

ghentschke:
fixes #808 
fixes https://github.com/eclipse-cdt/cdt-lsp/issues/338

### DIFF
--- a/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/internal/ui/util/EditorUtility.java
+++ b/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/internal/ui/util/EditorUtility.java
@@ -97,6 +97,7 @@ import org.eclipse.ui.IEditorInput;
 import org.eclipse.ui.IEditorPart;
 import org.eclipse.ui.IEditorRegistry;
 import org.eclipse.ui.IFileEditorInput;
+import org.eclipse.ui.IURIEditorInput;
 import org.eclipse.ui.IWorkbench;
 import org.eclipse.ui.IWorkbenchPage;
 import org.eclipse.ui.IWorkbenchWindow;
@@ -677,9 +678,9 @@ public class EditorUtility {
 			}
 		}
 		IEditorDescriptor desc = null;
-		if (input instanceof ExternalEditorInput externalInput) {
+		if (input instanceof IURIEditorInput uriEditorInput) {
 			try {
-				IFileStore fileStore = EFS.getFileSystem("file").getStore(externalInput.getPath()); //$NON-NLS-1$
+				IFileStore fileStore = EFS.getStore(uriEditorInput.getURI());
 				// get editor by considering overridden default editor association via IEditorAssociationOverride:
 				desc = IDE.getEditorDescriptorForFileStore(fileStore, false);
 			} catch (CoreException e) {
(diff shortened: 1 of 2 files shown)


ghentschke:
> If you need it in this release, let me know and I will investigate more to see if we can squeeze it in.

I know that this is pretty late but I found this issue in the morning. IMO it's a bigger bug when users want to work with old an new editor in the same workspace (as we do in out product) and we want to update our IDE after this release. So it would be great if you have the time to check this PR.

jonahgraham:
OK - I will have a look in detail. If I don't see too much concern I will merge this before RC2 build.

core/org.eclipse.cdt.ui/src/org/eclipse/cdt/internal/ui/util/EditorUtility.java
```diff
@@ -676,8 +676,20 @@ public static String getEditorID(IEditorInput input, Object inputObject) {
 				contentType = Platform.getContentTypeManager().getContentType(CCorePlugin.CONTENT_TYPE_BINARYFILE);
 			}
 		}
-		IEditorRegistry registry = PlatformUI.getWorkbench().getEditorRegistry();
-		IEditorDescriptor desc = registry.getDefaultEditor(input.getName(), contentType);
+		IEditorDescriptor desc = null;
+		if (input instanceof ExternalEditorInput externalInput) {
```
ghentschke, line 680: IMO the risk here is moderate, because this code will be reached only if the input is not  an instance of `IFileEditorInput` and it must be an instanceof `ExternalEditorInput` which limits the call on non workspace files as well. However, when this code should throw an exception, the previous behavior is ensured (desc is still null).

jonahgraham: Can you have a look at https://github.com/eclipse-cdt/cdt/blob/46d7fb74da53129d42296ef3f86730ddd811202e/debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/CDebugModelPresentation.java#L232-L240 and let me know if that pattern (which is also used in LSP4E) is more general and applicable here? 

Besides that question - I agree this change is low risk and worth getting in.

ghentschke: yes, it works as well. Good point!

jonahgraham (approved):

jonahgraham:
Thanks @ghentschke - I have merged now.

PR #810, opened 2024-06-03
Sets the CMake CMAKE_BUILD_TYPE define Fixes #807

### LINKED ISSUE #807: CMake projects are ignoring the Launch mode Run/Debug
CMake projects are not setting cmake -DCMAKE_BUILD_TYPE=... based on the Launch mode in the LaunchBar.
If you switch Launch mode to Debug in the LaunchBar and build you get a binary without debug information, because CFLAG -g was missing. Hitting a breakpoint will show you no code.

**To Reproduce**
1. Create a CMake project with hello world example.
2. Set Launch mode to Debug.
3. Build
4. Start debug

The program will break at main, but you see no code. 

**Expected behavior**
You see the code.

**Screenshots**
image

 - OS and OS Version/extra details: Linux
 - Eclipse Version from Help -> 2024-06 (4.32)
 - CDT Version from Help -> 11.6

**Additional context**
I did some debug.
In CommandDescriptorBuilder.makeCMakeCommandline() a call is done to cmakeProperties.getBuildType() which always returns null.
And this is caused by the fact that the build type was never set. CMakePropertiesBean.setBuildType() is never called.

betamaxbandit:
The CMAKE_BUILD_TYPE is added to CMake configuration build, based on the active launch mode; DEBUG_MODE sets a buildType of "Debug", everything else sets a buildType of "Release".
Note, the CMake CMAKE_BUILD_TYPE definition is not directly equivalent to the Launch Bar Launch Mode; there might be a scenario where the user wants to launch a debug session using the CMake Release build and not run it. However, for most cases this is probably sufficient.

For Issue CMake projects are ignoring the Launch mode Run/Debug #807.

### DIFF
--- a/cmake/org.eclipse.cdt.cmake.core/META-INF/MANIFEST.MF
+++ b/cmake/org.eclipse.cdt.cmake.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.cdt.cmake.core;singleton:=true
-Bundle-Version: 1.5.500.qualifier
+Bundle-Version: 1.5.600.qualifier
 Bundle-Activator: org.eclipse.cdt.cmake.core.internal.Activator
 Bundle-Vendor: %providerName
 Require-Bundle: org.eclipse.core.runtime,
--- a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeBuildConfiguration.java
+++ b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeBuildConfiguration.java
@@ -68,6 +68,7 @@ import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.jobs.Job;
+import org.eclipse.debug.core.ILaunchManager;
 import org.osgi.service.prefs.Preferences;
 
 public class CMakeBuildConfiguration extends CBuildConfiguration {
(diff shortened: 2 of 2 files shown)


betamaxbandit:
Manual Test instructions:


Setup host and Eclipse according to:
https://github.com/eclipse-cdt/cdt/blob/main/cmake/org.eclipse.cdt.cmake.ui.tests/manualTests/Bug579242_manual_tests.md#setup--prerequisites

Create a CMake project
* In Eclipse, choose File > C/C++ Project.
*  In the New C/C++ Project wizard, choose CMake in the left hand side sash and then CMake Project and click Next.
*  Enter a project name, for example helloworld and click Finish.

In the Launch Bar, select the Local Launch Target.

Test 1: Launch Mode "Run" causes a CMake "Release" build:
  In the Launch Bar Launch Mode, select Run.
  In the Launch Bar, click Build.
  The Console shows build activity 

```
  Configuring in: C:\Users\betamax\cdt-only2\runtime-New_configuration_CDT2\cmake_hw\build\default
  cmake -G MinGW Makefiles -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXPORT_COMPILE_COMMANDS=ON C:\Users\betamax\cdt-only2\runtime-New_configuration_CDT2\cmake_hw
  ...
```
Expected result: the presence of "-DCMAKE_BUILD_TYPE=Release"

Test 2: Launch Mode "Debug" causes a CMake "Debug" build:
  In the Launch Bar Launch Mode, select Debug.
  In the Launch Bar, click Build.
  The Console shows build activity 
  ```
Configuring in: C:\Users\betamax\cdt-only2\runtime-New_configuration_CDT2\cmake_hw\build\cmake.debug.win32.x86_64
  cmake -G MinGW Makefiles -DCMAKE_BUILD_TYPE=Debug -DCMAKE_EXPORT_COMPILE_COMMANDS=ON C:\Users\betamax\cdt-only2\runtime-New_configuration_CDT2\cmake_hw
  ...
```
Expected result: the presence of "-DCMAKE_BUILD_TYPE=Debug"

ewaterlander:
> Note, the CMake CMAKE_BUILD_TYPE definition is not directly equivalent to the Launch Bar Launch Mode; there might be a scenario where the user wants to launch a debug session using the CMake Release build and not run it.

In the launch configuration the user should be able to override the build type with the additional CMake arguments. But I see an issue that the additional arguments are not added at the end, but before CMAKE_BUILD_TYPE is set based on the launch mode.

PR #814, opened 2024-06-05
Add user additional CMake arugments at the end.

ewaterlander:
The additional CMake arguments added by the user in the launch configuration UI are added at the end of the CMake command, such that the user is able to override arguments set by the system.

### DIFF
--- a/cmake/org.eclipse.cdt.cmake.core/META-INF/MANIFEST.MF
+++ b/cmake/org.eclipse.cdt.cmake.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.cdt.cmake.core;singleton:=true
-Bundle-Version: 1.5.500.qualifier
+Bundle-Version: 1.5.600.qualifier
 Bundle-Activator: org.eclipse.cdt.cmake.core.internal.Activator
 Bundle-Vendor: %providerName
 Require-Bundle: org.eclipse.core.runtime,
--- a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeBuildConfiguration.java
+++ b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeBuildConfiguration.java
@@ -33,9 +33,9 @@ import org.eclipse.cdt.cmake.core.ICMakeToolChainFile;
 import org.eclipse.cdt.cmake.core.ICMakeToolChainManager;
 import org.eclipse.cdt.cmake.core.ParsingConsoleOutputStream;
 import org.eclipse.cdt.cmake.core.internal.CommandDescriptorBuilder.CommandDescriptor;
-import org.eclipse.cdt.cmake.core.internal.properties.CMakePropertiesBean;
 import org.eclipse.cdt.cmake.core.properties.CMakeGenerator;
 import org.eclipse.cdt.cmake.core.properties.ICMakeProperties;
+import org.eclipse.cdt.cmake.core.properties.ICMakePropertiesController;
 import org.eclipse.cdt.cmake.core.properties.IOsOverrides;
 import org.eclipse.cdt.core.CommandLauncherManager;
 import org.eclipse.cdt.core.ConsoleOutputStream;
@@ -83,6 +83,9 @@ public class CMakeBuildConfiguration extends CBuildConfiguration {
 	public static final String CLEAN_COMMAND_DEFAULT = "clean"; //$NON-NLS-1$
 
 	private ICMakeToolChainFile toolChainFile;
+
+	// lazily instantiated..
+	private CMakePropertiesController pc;
 
 	private Map<IResource, IScannerInfo> infoPerResource;
 	/**
@@ -155,7 +158,7 @@ public class CMakeBuildConfiguration extends CBuildConfiguration {
 				runCMake = true;
 			}
 
-			ICMakeProperties cmakeProperties = new CMakePropertiesBean();
+			ICMakeProperties cmakeProperties = getPropertiesController().load();
 			runCMake |= !Files.exists(buildDir.resolve("CMakeCache.txt")); //$NON-NLS-1$
 
 			final SimpleOsOverridesSelector overridesSelector = new SimpleOsOverridesSelector();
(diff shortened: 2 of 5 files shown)


ewaterlander:
Hi @betamaxbandit 
See this change.

betamaxbandit:
Hi @ewaterlander ,
Seems like a good idea.
Just to confirm the intention here is that the last define value on the CMake command line is used in preference to the same define earlier in the line?

For example, the following CMake invocation has 2 CMAKE_BUILD_TYPE defines and it is the last, Debug, that CMake uses:

cmake -G MinGW Makefiles -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXPORT_COMPILE_COMMANDS=ON C:\Users\betamax\cdt-only2\runtime-New_configuration_CDT2\cmake_hw -DCMAKE_BUILD_TYPE=Debug 

Is this correct?

ewaterlander:
> Hi @ewaterlander , Seems like a good idea. Just to confirm the intention here is that the last define value on the CMake command line is used in preference to the same define earlier in the line?
> 
> For example, the following CMake invocation has 2 CMAKE_BUILD_TYPE defines and it is the last, Debug, that CMake uses:
> 
> cmake -G MinGW Makefiles -DCMAKE_BUILD_TYPE=Release -DCMAKE_EXPORT_COMPILE_COMMANDS=ON C:\Users\betamax\cdt-only2\runtime-New_configuration_CDT2\cmake_hw -DCMAKE_BUILD_TYPE=Debug
> 
> Is this correct?

Yes, that is correct. The last -DCMAKE_BUILD_TYPE= argument prevails.

ewaterlander:
I looked a bit further at the code. Above my change, on line 80, there is also a call CommandDescriptorBuilder.appendCMakeArguments(args, cmakeProperties.getExtraArguments());
Here cmakeProperties.getExtraArguments() returns an empty list.

cmakeProperties.getExtraArguments() is getting settings that were stored in <project-dir>/.settings/CDT-cmake.yaml
This file was created in CMakeBuildConfiguration.getPropertiesController().

In CMakeBuildConfiguration.build() the settings from this file are loaded at line 162:
ICMakeProperties cmakeProperties = getPropertiesController().load();

There is no place in the code where extra arguments are stored in this file.

The launch configuration UI stores the additional CMake arguments in the project's preference store, and I think that is the correct place. The method I moved to the end also gets the settings from there.

I think the PropertiesController and the YAML file can go. It is actually not used and it only make the code confusing. There is no need for an extra YAML file to store CMake settings. This can be done in a separate PR.

betamaxbandit (approved):
LGTM

betamaxbandit:
> I looked a bit further at the code. Above my change, on line 80, there is also a call CommandDescriptorBuilder.appendCMakeArguments(args, cmakeProperties.getExtraArguments()); Here cmakeProperties.getExtraArguments() returns an empty list.
> 
> cmakeProperties.getExtraArguments() is getting settings that were stored in /.settings/CDT-cmake.yaml This file was created in CMakeBuildConfiguration.getPropertiesController().
> 
> In CMakeBuildConfiguration.build() the settings from this file are loaded at line 162: ICMakeProperties cmakeProperties = getPropertiesController().load();
> 
> There is no place in the code where extra arguments are stored in this file, so you will always get an empty list.
> 
> The launch configuration UI stores the additional CMake arguments in the project's preference store, and I think that is the correct place. The method I moved to the end also gets the settings from there.
> 
> I think the the YAML file can go. It is actually not used and it only makes the code confusing. There is no need for an extra YAML file to store CMake settings. I will make a separate PR.

I would urge caution.
1) 
I override com.renesas.cdt.epl.cmake.core.CMakeBuildConfiguration.build(int, Map<String, String>, IConsole, IProgressMonitor).
And in it I I use the ICMakePropertiesController to set my own extra arguments, similar to this: 

		ICMakePropertiesController propertiesController = getAdapter(ICMakePropertiesController.class);
		ICMakeProperties cmakeProps = propertiesController.load();
		cmakeProps.setExtraArguments(extraArguments);
		propertiesController.save(cmakeProps);

So I suggest not removing:
 		CommandDescriptorBuilder.appendCMakeArguments(args, cmakeProperties.getExtraArguments());


2) Regarding the Yaml file, I've often wondered what it's purpose is.
Could it be related to the CMake GUI?
Perhaps Martin Weber knows - I've forgotten his github handle.

ewaterlander:
> I would urge caution. 1) I override com.renesas.cdt.epl.cmake.core.CMakeBuildConfiguration.build(int, Map<String, String>, IConsole, IProgressMonitor). And in it I I use the ICMakePropertiesController to set my own extra arguments, similar to this:
> 
> ```
> 	ICMakePropertiesController propertiesController = getAdapter(ICMakePropertiesController.class);
> 	ICMakeProperties cmakeProps = propertiesController.load();
> 	cmakeProps.setExtraArguments(extraArguments);
> 	propertiesController.save(cmakeProps);
> ```
> 
> So I suggest not removing: CommandDescriptorBuilder.appendCMakeArguments(args, cmakeProperties.getExtraArguments());
> 
> 2. Regarding the Yaml file, I've often wondered what it's purpose is.
>    Could it be related to the CMake GUI?
>    Perhaps Martin Weber knows - I've forgotten his github handle.

If it would work correctly then the line
CommandDescriptorBuilder.appendCMakeArguments(args, cmakeProperties.getExtraArguments());
would add the user additional arguments (stored in YAML) and later they would be added again, because additional arguments are stored in the project preferences.
The reason we don't get double user arguments on the command line is that the YAML file is never saved. So the load() always returns an empty CMakeProperties.

Secondly there is a single YAML file for both Run and Debug. This is wrong.

I locally saved the YAML file. The user arguments are not stored correctly in the YAML file. So it is totally broken.

I will not remove it, but deprecate it.

This is how the YAML file looks like. Instead of arguments you get some identifier.

`buildType: Release
cacheFile: ''
clearCache: false
debugOutput: false
debugTryCompile: false
extraArguments: &id001 []
linuxOverrides:
  command: cmake
  extraArguments: *id001
  generator: UnixMakefiles
  useDefaultCommand: true
trace: false
warnNoDev: false
warnUnitialized: false
warnUnused: false
windowsOverrides:
  command: cmake
  extraArguments: *id001
  generator: MinGWMakefiles
  useDefaultCommand: true`

betamaxbandit:
Hi @15knots ,
Can you say what the purpose of the YAML file is please?

ewaterlander:
The only usage of the CMakePropertiesController is creating an empty new CMakeProperties by loading a non-existent YAML file. So my conclusion is that we can deprectate the CMakePropertiesController completely and remove all its usage.

I guess this is some old code that was not removed after somebody stored the CMake preferences in the project preferences, as it should be.

ewaterlander:
I deprecated CMakePropertiesController and removed its usage in this PR.

15knots:
Am Freitag, 7. Juni 2024, 12:27:09 CEST schrieb betamax:
> Hi @15knots ,
> Can you say what the purpose of the YAML file is please?

I planned to allowto persist option values for the cmake command-line on a 
per-projecz level in the yaml file. 
But when I found out that some of the cmake options are passed through totally 
undocumented properties from a project not being part of CDT (launchbar 
plugin) I gave up.
You may safely delete the stuff dealing with that or find a way to combine both 
the YAML stettings and the launchbar settings.

Martin 

-- 
Cd wrttn wtht vwls s mch trsr.

ewaterlander:
> Code review LGTM.

Hi @jonahgraham @betamaxbandit 

John approved commit 1. Commit 3 reverted commit 2. I think this PR is OK for merging.

regards,

Erwin

PR #820, opened 2024-06-11
Start building against Eclipse Platform 4.33 I-builds

jonahgraham:


jonahgraham:
As numerous dependencies of CDT are also starting to build against 4.33 we should too.

Issue #822, opened 2024-06-12
Compiler path in compile_commands.json must be absolute


alicetrifu:
Currently, the path to the compiler in the compile_commands.json file is not absolute; it is just the name of the executable. For better functionality, it is necessary to specify the exact location of the compiler in the file.
eg:
![image](https://github.com/eclipse-cdt/cdt/assets/61590135/6fabd962-06f8-466d-90e9-609fecd8e43a)

jonahgraham:
Thanks @alicetrifu for the fix.

PR #824, opened 2024-06-13
Compiler path in compile_commands.json must be absolute

alicetrifu:
Added code to create the absolute path of the compiler on compile_commands.json file
#822

### DIFF
--- a/build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/core/tests/CompilationDatabaseGenerationTest.java
+++ b/build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/core/tests/CompilationDatabaseGenerationTest.java
@@ -12,6 +12,7 @@ package org.eclipse.cdt.managedbuilder.core.tests;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.FileReader;
@@ -20,9 +21,12 @@ import org.eclipse.cdt.managedbuilder.core.jsoncdb.CompilationDatabaseInformatio
 import org.eclipse.cdt.managedbuilder.internal.core.CommonBuilder;
 import org.eclipse.cdt.managedbuilder.testplugin.AbstractBuilderTest;
 import org.eclipse.cdt.managedbuilder.testplugin.ManagedBuildTestHelper;
+import org.eclipse.cdt.utils.CommandLineUtil;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IncrementalProjectBuilder;
+import org.eclipse.core.runtime.IPath;
+import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.jface.preference.IPreferenceStore;
 import org.eclipse.ui.preferences.ScopedPreferenceStore;
@@ -166,5 +170,33 @@ public class CompilationDatabaseGenerationTest extends AbstractBuilderTest {
 		IPreferenceStore preferenceStore = new ScopedPreferenceStore(InstanceScope.INSTANCE,
 				"org.eclipse.cdt.managedbuilder.ui");
 		preferenceStore.setToDefault(CommonBuilder.COMPILATION_DATABASE_ENABLEMENT);
+	}
+
+	@Test
+	public void testCompilerPath() throws Exception {
+		setWorkspace("regressions");
+		final IProject app = loadProject("helloworldC");
+		setGenerateFileOptionEnabled(true);
+		app.build(IncrementalProjectBuilder.FULL_BUILD, null);
+		IFile commandsFile = app.getFile("Debug/compile_commands.json");
+
+		if (commandsFile.exists()) {
+			try (FileReader reader = new FileReader(commandsFile.getLocation().toFile())) {
+				Gson gson = new Gson();
+				JsonArray jsonArray = gson.fromJson(reader, JsonArray.class);
+				for (JsonElement element : jsonArray) {
+					CompilationDatabaseInformation compileCommand = gson.fromJson(element,
+							CompilationDatabaseInformation.class);
+					String command = compileCommand.command();
+					String[] commandParts = CommandLineUtil.argumentsToArray(command);
+					String compilerPath = commandParts[0];
+					assertNotNull("Compiler path should not be null", compilerPath);
+					assertFalse(compilerPath.isEmpty(), "Compiler path should not be empty");
+					IPath path = new Path(compilerPath);
+					boolean isAbsolute = path.isAbsolute();
+					assertTrue(isAbsolute, "Path should be absolute: " + path);
+				}
+			}
+		}
 	}
 }
--- a/build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
+++ b/build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
@@ -10,13 +10,19 @@
 package org.eclipse.cdt.managedbuilder.internal.core.jsoncdb.generator;
 
 import java.io.ByteArrayInputStream;
+import java.io.File;
 import java.io.InputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
+import java.util.Map;
 
+import org.eclipse.cdt.core.CCorePlugin;
+import org.eclipse.cdt.core.envvar.IEnvironmentVariable;
+import org.eclipse.cdt.core.settings.model.ICConfigurationDescription;
 import org.eclipse.cdt.core.settings.model.ICSourceEntry;
 import org.eclipse.cdt.core.settings.model.util.CDataUtil;
 import org.eclipse.cdt.managedbuilder.core.BuildException;
@@ -34,6 +40,8 @@ import org.eclipse.cdt.managedbuilder.core.jsoncdb.ICompilationDatabaseContribut
 import org.eclipse.cdt.managedbuilder.internal.macros.FileContextData;
 import org.eclipse.cdt.managedbuilder.macros.BuildMacroException;
 import org.eclipse.cdt.managedbuilder.macros.IBuildMacroProvider;
+import org.eclipse.cdt.utils.CommandLineUtil;
+import org.eclipse.cdt.utils.PathUtil;
 import org.eclipse.core.resources.IFile;
 import org.eclipse.core.resources.IFolder;
 import org.eclipse.core.resources.IProject;
@@ -58,7 +66,11 @@ public final class CompilationDatabaseGenerator {
 
 	private static final String CDB_FILENAME = "compile_commands.json"; //$NON-NLS-1$
 	private static final String ERROR_MESSAGE = "Can not set contents to compile_commands.json file"; //$NON-NLS-1$
-
+	/**
+	 * Checked on each build
+	 * Used before we look up the environment
+	 */
+	private Map<ITool, String> toolMap = new HashMap<>();
 	private IProject project;
 	private IConfiguration configuration;
 	private ICSourceEntry[] srcEntries;
@@ -217,10 +229,22 @@ public final class CompilationDatabaseGenerator {
 						outputLocation + "", inputStrings, sourceLocation, outputLocation); //$NON-NLS-1$
 
 				IBuildMacroProvider provider = ManagedBuildManager.getBuildMacroProvider();
-				String resolvedOptionFileContents = provider.resolveValueToMakefileFormat(cmdLInfo.getCommandLine(), "", //$NON-NLS-1$
-						" ", IBuildMacroProvider.CONTEXT_FILE, //$NON-NLS-1$
-						new FileContextData(sourceLocation, outputLocation, null, tool));
+				String compilerName = CompilationDatabaseGenerator.getCompilerName(tool);
+				String commandLine = cmdLInfo.getCommandLine();
+				commandLine = commandLine.replace(compilerName, "").trim(); //$NON-NLS-1$
+				String compilerPath = findCompilerInPath(tool, config);
+				String resolvedOptionFileContents;
+				if (compilerPath != null && !compilerPath.isEmpty()) {
+					resolvedOptionFileContents = provider.resolveValueToMakefileFormat(compilerPath + " " + commandLine, //$NON-NLS-1$
+							"", " ", //$NON-NLS-1$//$NON-NLS-2$
+							IBuildMacroProvider.CONTEXT_FILE,
+							new FileContextData(sourceLocation, outputLocation, null, tool));
+				} else {
+					resolvedOptionFileContents = provider.resolveValueToMakefileFormat(commandLine, "", " ", //$NON-NLS-1$//$NON-NLS-2$
+							IBuildMacroProvider.CONTEXT_FILE,
+							new FileContextData(sourceLocation, outputLocation, null, tool));
 
+				}
 				objList.add(new CompilationDatabaseInformation(project.getLocation().toString(),
 						resolvedOptionFileContents, resource.getLocation().toString()));
 			}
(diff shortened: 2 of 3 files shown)


build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -217,8 +225,12 @@ private List<CompilationDatabaseInformation> populateObjList(IProject project, I
 						outputLocation + "", inputStrings, sourceLocation, outputLocation); //$NON-NLS-1$
 
 				IBuildMacroProvider provider = ManagedBuildManager.getBuildMacroProvider();
-				String resolvedOptionFileContents = provider.resolveValueToMakefileFormat(cmdLInfo.getCommandLine(), "", //$NON-NLS-1$
-						" ", IBuildMacroProvider.CONTEXT_FILE, //$NON-NLS-1$
+				String compilerName = CompilationDatabaseGenerator.getCompilerName(tool);
+				String commandLine = cmdLInfo.getCommandLine();
+				commandLine = commandLine.replace(compilerName, "").trim(); //$NON-NLS-1$
+				commandLine = commandLine.replaceAll("\\s+", " ").trim(); //$NON-NLS-1$ //$NON-NLS-2$
```
jonahgraham, line 231: This seems unexpected to me - at least for corner cases. What if there is spaces in existing file names? Maybe if you can explain what you are trying to achieve with this line I will understand better.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -414,4 +426,41 @@ public boolean visit(IResourceProxy proxy) throws CoreException {
...
+		}
+		ICConfigurationDescription cfg = ManagedBuildManager.getDescriptionForConfiguration(config);
+		IEnvironmentVariable[] variables = CCorePlugin.getDefault().getBuildEnvironmentManager().getVariables(cfg,
+				true);
+
+		for (IEnvironmentVariable variable : variables) {
+			if ("PATH".equals(variable.getName())) { //$NON-NLS-1$
+				IPath resolvedPath = PathUtil.findProgramLocation(compilerName, variable.getValue());
+				if (resolvedPath != null) {
+					String path = resolvedPath.toString().replaceAll("\\s+", " ").trim(); //$NON-NLS-1$//$NON-NLS-2$
```
jonahgraham, line 448: Similar question here as to line 231

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -414,4 +426,41 @@ public boolean visit(IResourceProxy proxy) throws CoreException {
...
+		if (pathToCompiler.isAbsolute()) {
+			toolMap.put(tool, compilerName);
+			return "\"" + compilerName + "\""; //$NON-NLS-1$ //$NON-NLS-2$
+		}
+		ICConfigurationDescription cfg = ManagedBuildManager.getDescriptionForConfiguration(config);
+		IEnvironmentVariable[] variables = CCorePlugin.getDefault().getBuildEnvironmentManager().getVariables(cfg,
+				true);
+
+		for (IEnvironmentVariable variable : variables) {
+			if ("PATH".equals(variable.getName())) { //$NON-NLS-1$
```
jonahgraham, line 445: Is PATH guaranteed to be uppercase in return from getVariables? Normally on Windows we have to handle the case of path being case insensitive.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -414,4 +426,41 @@ public boolean visit(IResourceProxy proxy) throws CoreException {
 
 	}
 
+	private String findCompilerInPath(ITool tool, IConfiguration config) {
+		if (toolMap.containsKey(tool)) {
```
jonahgraham, line 430: Can `CCorePlugin.getDefault().getBuildEnvironmentManager().getVariables(cfg, true)` can return different values over the lifetime of the toolMap cache?

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -217,8 +225,12 @@ private List<CompilationDatabaseInformation> populateObjList(IProject project, I
...
 
 				IBuildMacroProvider provider = ManagedBuildManager.getBuildMacroProvider();
-				String resolvedOptionFileContents = provider.resolveValueToMakefileFormat(cmdLInfo.getCommandLine(), "", //$NON-NLS-1$
-						" ", IBuildMacroProvider.CONTEXT_FILE, //$NON-NLS-1$
+				String compilerName = CompilationDatabaseGenerator.getCompilerName(tool);
+				String commandLine = cmdLInfo.getCommandLine();
+				commandLine = commandLine.replace(compilerName, "").trim(); //$NON-NLS-1$
+				commandLine = commandLine.replaceAll("\\s+", " ").trim(); //$NON-NLS-1$ //$NON-NLS-2$
+				String resolvedOptionFileContents = provider.resolveValueToMakefileFormat(
+						findCompilerInPath(tool, config) + " " + commandLine, "", " ", IBuildMacroProvider.CONTEXT_FILE, //$NON-NLS-1$//$NON-NLS-2$ //$NON-NLS-3$
```
jonahgraham, line 233: findCompilerInPath can return `null` - so I think we have to handle that either by having it not return null, or if it does return null than don't modify the command line? 

With the null AFAICT you can end up with a string like `null -c a.c -o a.o` in the compile_commands.json

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -217,8 +225,12 @@ private List<CompilationDatabaseInformation> populateObjList(IProject project, I
 						outputLocation + "", inputStrings, sourceLocation, outputLocation); //$NON-NLS-1$
 
 				IBuildMacroProvider provider = ManagedBuildManager.getBuildMacroProvider();
-				String resolvedOptionFileContents = provider.resolveValueToMakefileFormat(cmdLInfo.getCommandLine(), "", //$NON-NLS-1$
-						" ", IBuildMacroProvider.CONTEXT_FILE, //$NON-NLS-1$
+				String compilerName = CompilationDatabaseGenerator.getCompilerName(tool);
+				String commandLine = cmdLInfo.getCommandLine();
+				commandLine = commandLine.replace(compilerName, "").trim(); //$NON-NLS-1$
+				commandLine = commandLine.replaceAll("\\s+", " ").trim(); //$NON-NLS-1$ //$NON-NLS-2$
+				String resolvedOptionFileContents = provider.resolveValueToMakefileFormat(
```
jonahgraham, line 232: Since resolveValueToMakefileFormat turns variables into values can `cmdLInfo.getCommandLine()` .`CompilationDatabaseGenerator.getCompilerName(tool)` return a variable for the compiler name? If so, how is that dealt with (I suspect the null comment on the next line is related)

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -414,4 +426,41 @@ public boolean visit(IResourceProxy proxy) throws CoreException {
...
+			}
+		}
+
+		return null;
+	}
+
+	private static String getCompilerName(ITool tool) {
+		String compilerCommand = tool.getToolCommand();
+		String[] arguments = CommandLineUtil.argumentsToArray(compilerCommand);
+		return arguments[0];
```
jonahgraham, line 463: Possible index out of bounds exception here. Generally that would only happen if there is a problem in the tool, but we should avoid the runtime exception here.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
alicetrifu: No, it can't. That's why the check is done before we look up the environment. I've addded a comment on the variable.

jonahgraham: Great, thanks for confirming and documenting.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -414,4 +438,44 @@ public boolean visit(IResourceProxy proxy) throws CoreException {
...
+		}
+
+		return null;
+	}
+
+	private static String getCompilerName(ITool tool) {
+		String compilerCommand = tool.getToolCommand();
+		String[] arguments = CommandLineUtil.argumentsToArray(compilerCommand);
+		if (arguments.length == 0) {
+			throw new IllegalArgumentException("No arguments found in the compiler command: " + compilerCommand); //$NON-NLS-1$
```
jonahgraham, line 476: This is still a runtime exception. Can you return an empty string here please.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -217,10 +229,22 @@ private List<CompilationDatabaseInformation> populateObjList(IProject project, I
...
-						" ", IBuildMacroProvider.CONTEXT_FILE, //$NON-NLS-1$
-						new FileContextData(sourceLocation, outputLocation, null, tool));
+				String compilerName = CompilationDatabaseGenerator.getCompilerName(tool);
+				String commandLine = cmdLInfo.getCommandLine();
+				commandLine = commandLine.replace(compilerName, "").trim(); //$NON-NLS-1$
+				String compilerPath = findCompilerInPath(tool, config);
+				String resolvedOptionFileContents;
+				if (compilerPath != null && !compilerPath.isEmpty()) {
+					resolvedOptionFileContents = provider.resolveValueToMakefileFormat(
+							findCompilerInPath(tool, config) + " " + commandLine, "", " ", //$NON-NLS-1$//$NON-NLS-2$//$NON-NLS-3$
```
jonahgraham, line 239:
```suggestion
							compilerPath + " " + commandLine, "", " ", //$NON-NLS-1$//$NON-NLS-2$//$NON-NLS-3$
```

(you may need to reapply formatting too because of line length change

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
jonahgraham: Code changed in latest commit, resolving.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
jonahgraham: Fixed in latest version with equalsIgnoreCase

jonahgraham (changes requested):
This looks good. Please add some tests, can you cover cases when compiler is on the system PATH, but the project PATH points to a different one? I think that case is the most likely to regress as it is non-obvious to most developers.

jonahgraham (changes requested):
Thanks @alicetrifu for the updates - the code all looks ok now. Just the tests as we discussed from the last review.

build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/core/tests/CompilationDatabaseGenerationTest.java
```diff
@@ -167,4 +170,32 @@ public void restoreDefaultForGenerateFile() {
...
+				JsonArray jsonArray = gson.fromJson(reader, JsonArray.class);
+				for (JsonElement element : jsonArray) {
+					CompilationDatabaseInformation compileCommand = gson.fromJson(element,
+							CompilationDatabaseInformation.class);
+					String command = compileCommand.command();
+					String[] commandParts = command.split(" ");
+					String compilerPath = commandParts[0]; // assuming the compiler path is the first part of the command
+					assertNotNull("Compiler path should not be null", compilerPath);
+					IPath path = new Path(compilerPath);
+					assertTrue(path.isAbsolute());
```
jonahgraham, line 194: This line of the test is failing:

```
expected: <true> but was: <false>
org.opentest4j.AssertionFailedError: expected: <true> but was: <false>
	at org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
	at org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
	at org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
	at org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:31)
	at org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:183)
	at org.eclipse.cdt.managedbuilder.core.tests.CompilationDatabaseGenerationTest.testCompilerPath(CompilationDatabaseGenerationTest.java:194)
```

jonahgraham (changes requested):

build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/core/tests/CompilationDatabaseGenerationTest.java
```diff
@@ -167,4 +170,43 @@ public void restoreDefaultForGenerateFile() {
...
+							CompilationDatabaseInformation.class);
+
+					String command = compileCommand.command();
+					String compilerPath;
+					if (command.startsWith("\"")) {
+						// Linux format with quotes around paths
+						int endIndex = command.indexOf("\"", 1); // find closing quote
+						compilerPath = command.substring(1, endIndex);
+					} else {
+						// Windows format with direct path
```
jonahgraham, line 197: I don't understand this else branch. In the code change you seem to always surround with `"`.

For this test can you use CommandLineUtil.argumentsToArray and then verify `result[0]` is an absolute path?

build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/core/tests/CompilationDatabaseGenerationTest.java
```diff
@@ -167,4 +171,34 @@ public void restoreDefaultForGenerateFile() {
...
+				for (JsonElement element : jsonArray) {
+					CompilationDatabaseInformation compileCommand = gson.fromJson(element,
+							CompilationDatabaseInformation.class);
+					String command = compileCommand.command();
+					String[] commandParts = CommandLineUtil.argumentsToArray(command);
+					String compilerPath = commandParts[0];
+					assertNotNull("Compiler path should not be null", compilerPath);
+					assertFalse("Compiler path should not be empty", compilerPath.isEmpty());
+					IPath path = new Path(compilerPath);
+					//Added this to see the source of the error and try to fix
```
jonahgraham, line 196: This code looks correct - I hope it is enough to diagnose the build machine error. 

Please tag me once you need my input again.

jonahgraham (changes requested):

build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/core/tests/CompilationDatabaseGenerationTest.java
alicetrifu: Hello @jonahgraham , your suggestion to use CommandLineUtil.argumentsToArray fixed the issue because the method will correctly treat the quoted string as a single argument, no matter the OS. The way I was splitting the command before was the problem with the Linux environment. 

Do you think I need to add anything else?

build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/core/tests/CompilationDatabaseGenerationTest.java
```diff
@@ -167,4 +171,33 @@ public void restoreDefaultForGenerateFile() {
...
+							CompilationDatabaseInformation.class);
+					String command = compileCommand.command();
+					String[] commandParts = CommandLineUtil.argumentsToArray(command);
+					String compilerPath = commandParts[0];
+					assertNotNull("Compiler path should not be null", compilerPath);
+					assertFalse("Compiler path should not be empty", compilerPath.isEmpty());
+					IPath path = new Path(compilerPath);
+					boolean isAbsolute = path.isAbsolute();
+					assertTrue("Path should be absolute: " + path, isAbsolute);
+					System.out.println("Compiler Path: " + path);
```
jonahgraham, line 198:
```suggestion
```

jonahgraham (changes requested):

build/org.eclipse.cdt.managedbuilder.core.tests/tests/org/eclipse/cdt/managedbuilder/core/tests/CompilationDatabaseGenerationTest.java
jonahgraham: Just to remove the added logging/debugging code. I didn't look thoroughly, but did mark one instance in latest review.

jonahgraham (changes requested):
This is ready to go, just noticed a minor nit about usage of JUnit. The test failure seems unrelated and I retriggered the build to verify that.

jonahgraham (approved):
Once build reports a clean run this is good to go.

PR #817, opened 2024-06-08
On createBuildConfiguration, reset project's ScannerInfoProvider

betamaxbandit:
When the project's active IBuildConfiguration has the default name and the chosen ICBuildConfigurationProvider.getCBuildConfiguration does not support the IBuildConfiguration.DEFAULT_CONFIG_NAME and returns null, this can cause the project's ScannerInfoProvider to become "stuck" (https://bugs.eclipse.org/bugs/show_bug.cgi?id=413357) on the wrong setting (eg LanguageSettingsScannerInfoProvider instead of ICBuildConfiguration) until Eclipse is restarted or the project is closed and reopened. When this happens, the indexer does not function.

This problem may arise if an ISV contributes a
ICBuildConfigurationProvider which has very specific naming conventions for it's build configurations.

The solution uses the API (resetCachedScannerInfoProvider(project)), introduced by 413357, to reset the project's ScannerInfoProvider when a new ICBuildConfiguration is created.

PR #826, opened 2024-06-13
On createBuildConfiguration, reset project's ScannerInfoProvider

betamaxbandit:
When the project's active IBuildConfiguration has the default name and the chosen ICBuildConfigurationProvider.getCBuildConfiguration does not support the IBuildConfiguration.DEFAULT_CONFIG_NAME and returns null, this can cause the project's ScannerInfoProvider to become "stuck" (https://bugs.eclipse.org/bugs/show_bug.cgi?id=413357) on the wrong setting (eg LanguageSettingsScannerInfoProvider instead of ICBuildConfiguration) until Eclipse is restarted or the project is closed and reopened. When this happens, the indexer does not function.

This problem may arise if an ISV contributes a
ICBuildConfigurationProvider which has very specific naming conventions for it's build configurations.

The solution uses the API (resetCachedScannerInfoProvider(project)), introduced by 413357, to reset the project's ScannerInfoProvider when a new ICBuildConfiguration is created.

### DIFF
--- a/core/org.eclipse.cdt.core.tests/META-INF/MANIFEST.MF
+++ b/core/org.eclipse.cdt.core.tests/META-INF/MANIFEST.MF
@@ -45,7 +45,10 @@ Require-Bundle: org.eclipse.core.resources,
  org.eclipse.ltk.core.refactoring;bundle-version="3.4.0",
  org.hamcrest.core,
  org.hamcrest.library,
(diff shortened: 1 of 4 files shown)


jonahgraham:
> as requested I've added this PR to backport my https://github.com/eclipse-cdt/cdt/pull/817 to the cdt_11_6 branch.

This PR is against the main branch, please create a PR against the cdt_11_6 branch.

image

PR #827, opened 2024-06-14
On createBuildConfiguration, reset project's ScannerInfoProvider

betamaxbandit:
When the project's active IBuildConfiguration has the default name and the chosen ICBuildConfigurationProvider.getCBuildConfiguration does not support the IBuildConfiguration.DEFAULT_CONFIG_NAME and returns null, this can cause the project's ScannerInfoProvider to become "stuck" (https://bugs.eclipse.org/bugs/show_bug.cgi?id=413357) on the wrong setting (eg LanguageSettingsScannerInfoProvider instead of ICBuildConfiguration) until Eclipse is restarted or the project is closed and reopened. When this happens, the indexer does not function.

This problem may arise if an ISV contributes a
ICBuildConfigurationProvider which has very specific naming conventions for it's build configurations.

The solution uses the API (resetCachedScannerInfoProvider(project)), introduced by 413357, to reset the project's ScannerInfoProvider when a new ICBuildConfiguration is created.

### DIFF
--- a/core/org.eclipse.cdt.core.tests/META-INF/MANIFEST.MF
+++ b/core/org.eclipse.cdt.core.tests/META-INF/MANIFEST.MF
@@ -45,7 +45,10 @@ Require-Bundle: org.eclipse.core.resources,
  org.eclipse.ltk.core.refactoring;bundle-version="3.4.0",
  org.hamcrest.core,
  org.hamcrest.library,
- com.google.gson;bundle-version="[2.8.6,3.0.0)"
+ com.google.gson;bundle-version="[2.8.6,3.0.0)",
+ org.eclipse.cdt.debug.core,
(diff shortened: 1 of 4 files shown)


jonahgraham:
I need to do some housekeeping on this branch before it will build cleanly. I think it points at the RC2 target platform which has probably been deleted.

jonahgraham:
Indeed:

```
Failed to load p2 metadata repository from location https://download.eclipse.org/eclipse/updates/4.32-I-builds/: No repository found at https://download.eclipse.org/eclipse/updates/4.32-I-builds.
```

This needs to be updated. I will do so soon.

Thanks for the backport PR

jonahgraham:
I am closing this and recreating it with the new backport action.

PR #829, opened 2024-06-16
Extend GNU toolchain optimization options for MBS

### LINKED ISSUE #828: Extend optimization options within MBS GNU toolchain description
The user should be able to select `-Og` (optimize for debug) and `-Oz` (optimize aggressively for size) from the list of available compiler optimization options within managed build projects using the GNU toolchain.

jld01:
Resolves #828


jld01:
There appears to be a repeatable issue with the license vetting status check. Can you advise please, @jonahgraham?

Ref: https://github.com/eclipse-cdt/cdt/actions/runs/9535221869/job/26285663622?pr=829#step:11:573

jonahgraham:
This is the error:

```
[INFO] License information could not be automatically verified for the following content:
[INFO] 
[INFO] p2/orbit/p2.eclipse.plugin/org.hamcrest.core/2.2.0.v20230809-1000
```

It is now successful on a rerun. The licensing depends on a live database whose scores sometimes change for existing content, and in this case it was fixed since the weekend.

PR #830, opened 2024-06-18
XML self-closing element consistency clean up only

jld01:


jld01:
Relates to https://github.com/eclipse-cdt/cdt/issues/646

Issue #842, opened 2024-06-23
Prepare 11.6 branch for developement


jonahgraham:
With the roadmap we recently decided to have an 11.6.x release as needed (see this announcement https://github.com/eclipse-cdt/cdt/discussions/837)

There are a number of important fixes (see items on the [11.6.1 milestone](https://github.com/eclipse-cdt/cdt/milestone/16)) that we plan to roll out.

This issue is to cover bringing the 11.6 branch "to life" as it's current target platform contains some I-build dependencies and we need to lock them down to the versions released in 2024-06 SimRel.

jonahgraham:
An additional item is (also as discussed in #837) if we can use some tooling to do the backporting like Eclipse Tycho does.

jonahgraham:
The branch seems to be ready to go, and the backport tool is doing its job.

Issue #816, opened 2024-06-05
Some CDT DSF GDB tests failing on GitHub Actions


jonahgraham:
At some point recently CDT's DSF GDB tests started having some failures on GitHub actions, specifically `8 errors, 60 fail`. See for example this run https://github.com/eclipse-cdt/cdt/runs/25845154142

jonahgraham:
The "8 errors, 60 fail" is the known and expected failures at the moment.

PR #844, opened 2024-06-23
Use final p2 repos for all CDT 11.6 dependencies

jonahgraham:
It is important we don't point at "latest" URLs for dependencies as we want to ensure compatibility with 2024-06 and if some project does a newer release we want to explicitly opt-in to it on this CDT 11.6 branch

Part of #842


jonahgraham:
This PR includes the other prep work for CDT 11.6.1 - see individual commits for additional details.

jonahgraham:
test failures are the known ones in #816

PR #880, opened 2024-08-06
Allow multiple parserPage XML elements

jld01:
Eliminates warning on `org.eclipse.cdt.ui/plugin.xml`


jld01:
The failing check is not related to this work.

PR #885, opened 2024-08-14
Add to string methods for testing.

raghucssit:
**This PR is not be merged. Only for testing purpse.**


raghucssit:
@Torbjorn-Svensson Please trigger build for this.
I will compare test results with my branch here. This gives me the comparable state of build machine to me.

PR #896, opened 2024-08-29
__DATE__ should have day prefixed with space

Torbjorn-Svensson:
The documentation for GCC, and other compilers, stipulates that `__DATE__` with a day of month less than 10 should be prefixed by a space, not a zero.

### DIFF
--- a/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/parser/scanner/DateMacro.java
+++ b/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/parser/scanner/DateMacro.java
@@ -16,6 +16,7 @@ package org.eclipse.cdt.internal.core.parser.scanner;
 
 import java.text.DateFormatSymbols;
 import java.util.Calendar;
+import java.util.Locale;
 
 import org.eclipse.cdt.core.parser.IToken;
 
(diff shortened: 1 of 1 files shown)


core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/parser/scanner/DateMacro.java
```diff
@@ -36,7 +36,11 @@ private char[] createDate() {
 		DateFormatSymbols dfs = new DateFormatSymbols();
 		buffer.append(dfs.getShortMonths()[cal.get(Calendar.MONTH)]);
```
Torbjorn-Svensson, line 37: I'm also a bit puzzled about this field as the compiler documentations appears to imply that the string is not translated, but the `java.text.DateFormatSymbols` is a way to retrieve the localized name of the month.

Maybe we should do as GCC does and just have an array of 3 letter English names of the months and pick from that?

jld01: @Torbjorn-Svensson I agree that the `__DATE__` macro should provide a non-localized string according to the ANSI standard. So my suggestion would be to specify an appropriate locale in the `DateFormatSymbols` constructor:

`new DateFormatSymbols(Locale.ENGLISH).getShortMonths()`

core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/parser/scanner/DateMacro.java
```diff
@@ -36,7 +36,11 @@ private char[] createDate() {
...
 		buffer.append(" "); //$NON-NLS-1$
-		append(buffer, cal.get(Calendar.DAY_OF_MONTH));
+		int dom = cal.get(Calendar.DAY_OF_MONTH);
+		if (dom < 10) {
+			buffer.append(" "); //$NON-NLS-1$
```
jld01, line 41: Looks good to me.

core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/parser/scanner/DateMacro.java
Torbjorn-Svensson: Thanks for the suggestion @jld01!

PR #901, opened 2024-09-13
Support environment variables in Core Build projects for Local

ewaterlander:
Added an Environment tab to the Core Build launch configurations for the Local target. Similar as the standard C/C++ Application launch configurations for Managed Build.


ewaterlander:
There are some tests failing due to a timeout in a gdb server request. I see no relations with this patch.

jonahgraham (approved):

Issue #908, opened 2024-10-07
CMakeToolChainManager removeToolChainFile not working


alirana01:
**Describe the bug**
CMakeToolChainManager is not removing the cmake toolchains. I can still see them even after calling the remove function.

**To Reproduce**
We have cmake toolchains that are initialized like and added using the CMakeToolChainManager. The issue is that when I am trying to remove them by calling the removeToolChainFile the tcId calculated there is at times different than the one that is used while adding them. Because of this the toolchains are not removed and I am still able to see them in the Preferences.

**Expected behavior**
Calling the toolchain manager's remove method should remove the toolchains from the eclipse preferences.

**Code**
My function to remove all CMake ToolChains.
```
	/**
	 * Removes all the cmake toolchains in the current environment
	 */
	public void removeCmakeToolChains()
	{
		ICMakeToolChainManager cmakeTcManager = CCorePlugin.getService(ICMakeToolChainManager.class);
		List<ICMakeToolChainFile> cIcMakeToolChainFiles = new ArrayList<>(cmakeTcManager.getToolChainFiles());
		for (ICMakeToolChainFile cmakeToolchain : cIcMakeToolChainFiles)
		{
			try
			{
				cmakeTcManager.removeToolChainFile(cmakeToolchain);
			}
			catch (Exception e)
			{
				Logger.log(e);
			}
		}
	}
```
	

The addToolChainFile(ICMakeToolChainFile file) function in CMakeToolChainManager is calculating tcId using 
`makeToolChainId(toolchain);`

but the removeToolChainFile(ICMakeToolChainFile file) is using this
```
String tcId = makeToolChainId(file.getProperty(CMakeBuildConfiguration.TOOLCHAIN_TYPE),
				file.getProperty(CMakeBuildConfiguration.TOOLCHAIN_ID));
```

I am not sure why these two are different while debugging calculating the tcId using the way in add I am able to successfully remove these.

**Version Information (please complete the following information):**
 - OS and OS Version/extra details: [e.g. Windows, Linux, macOS] (all)
 - Eclipse Version: v4.33.0 2024-09
 - CDT Version: 11.6.1

alirana01:
I also wanted to become a contributor to eclipse and have a branch and some code ready to push but seems like I need to have some access before doing that I have followed the CONTRIBUTING.MD guide and waiting for the approval to get it

jonahgraham:
[...]

PR #909, opened 2024-10-07
fix(908): fix for the cmake tool chain removal function.

alirana01:
The calculated ID in the removal function is different than the one in the add

[Original reported issue](https://github.com/eclipse-cdt/cdt/issues/908)

### DIFF
--- a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeToolChainManager.java
+++ b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeToolChainManager.java
@@ -166,10 +166,8 @@ public class CMakeToolChainManager implements ICMakeToolChainManager {
 				tcNode.removeNode();
 				prefs.flush();
 			}
-		} catch (CoreException ce) {
-			Activator.log(ce);
-		} catch (BackingStoreException be) {
-			Activator.log(be);
+		} catch (CoreException | BackingStoreException e) {
+			Activator.log(e);
 		}
 	}
 
--- a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeToolChainManager.java
+++ b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeToolChainManager.java
@@ -156,20 +156,18 @@ public class CMakeToolChainManager implements ICMakeToolChainManager {
 	public void removeToolChainFile(ICMakeToolChainFile file) {
 		init();
 		fireEvent(new CMakeToolChainEvent(CMakeToolChainEvent.REMOVED, file));
-		String tcId = makeToolChainId(file.getProperty(CMakeBuildConfiguration.TOOLCHAIN_TYPE),
-				file.getProperty(CMakeBuildConfiguration.TOOLCHAIN_ID));
-		filesByToolChain.remove(tcId);
-
-		String n = ((CMakeToolChainFile) file).n;
-		if (n != null) {
-			Preferences prefs = getPreferences();
-			Preferences tcNode = prefs.node(n);
-			try {
+		try {
+			String tcId = makeToolChainId(file.getToolChain());
+			filesByToolChain.remove(tcId);
+			String n = ((CMakeToolChainFile) file).n;
+			if (n != null) {
+				Preferences prefs = getPreferences();
+				Preferences tcNode = prefs.node(n);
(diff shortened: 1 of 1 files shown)


jonahgraham:
@betamaxbandit Can you review this please. It looks correct to me, but I think this is in your area.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeToolChainManager.java
```diff
@@ -156,20 +156,18 @@ public void addToolChainFile(ICMakeToolChainFile file) {
...
-				file.getProperty(CMakeBuildConfiguration.TOOLCHAIN_ID));
-		filesByToolChain.remove(tcId);
-
-		String n = ((CMakeToolChainFile) file).n;
-		if (n != null) {
-			Preferences prefs = getPreferences();
-			Preferences tcNode = prefs.node(n);
-			try {
+		try {
+			String tcId = makeToolChainId(file.getToolChain());
```
betamaxbandit, line 160: Yes, this looks like a definite improvement.

I couldn't figure out why the original code was written the way it was.

I think in the past I've worked around this by assigning the TOOLCHAIN_TYPE and TOOLCHAIN_ID properties in my code that adds the toolchain file, just after calling org.eclipse.cdt.cmake.core.ICMakeToolChainManager.newToolChainFile(Path), like this:

			// toolchainFile.setProperty(ICBuildConfiguration.TOOLCHAIN_TYPE, tc.getTypeId());
			// toolchainFile.setProperty(ICBuildConfiguration.TOOLCHAIN_ID, tc.getId());

But this new code improves that situation so it's no longer necessary.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeToolChainManager.java
```diff
@@ -156,20 +156,18 @@ public void addToolChainFile(ICMakeToolChainFile file) {
...
+			String n = ((CMakeToolChainFile) file).n;
+			if (n != null) {
+				Preferences prefs = getPreferences();
+				Preferences tcNode = prefs.node(n);
 				tcNode.removeNode();
 				prefs.flush();
-			} catch (BackingStoreException e) {
-				Activator.log(e);
 			}
+		} catch (Exception e) {
```
betamaxbandit, line 169: I would consider handling the CoreException, from org.eclipse.cdt.cmake.core.ICMakeToolChainFile.getToolChain(), separately. I think it's generally considered bad practice to catch the general Exception.

betamaxbandit (changes requested):
LGTM.
Just an outstanding issue with the exception handling.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeToolChainManager.java
alirana01: Thanks for review @betamaxbandit updated the exception block for CoreException and BackingStoreException separately.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeToolChainManager.java
jonahgraham: Thanks @betamaxbandit for having a look. Marking this thread as resolved.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeToolChainManager.java
```diff
@@ -156,20 +156,20 @@ public void addToolChainFile(ICMakeToolChainFile file) {
...
+				Preferences tcNode = prefs.node(n);
 				tcNode.removeNode();
 				prefs.flush();
-			} catch (BackingStoreException e) {
-				Activator.log(e);
 			}
+		} catch (CoreException ce) {
+			Activator.log(ce);
+		} catch (BackingStoreException be) {
+			Activator.log(be);
```
jonahgraham, line 172:
```suggestion
		} catch (CoreException | BackingStoreException e) {
			Activator.log(e);
```

jonahgraham (approved):
This LGTM. I am going to apply my minor suggested change and merge once CI approves too.

PR #915, opened 2024-10-11
Classify unrecognized command line options as an error

Torbjorn-Svensson:


core/org.eclipse.cdt.core/plugin.properties
```diff
@@ -67,6 +67,7 @@ CDTGNUCErrorParser.regex.ParseErrorBefore=(.*?):(\\d+):(\\d+:)? (parse error bef
 CDTGNUCErrorParser.regex.ErrorUndeclared=(.*?):(\\d+):(\\d+:)? [Ee]rror: ([`'"](.*)['"] undeclared .*)
 CDTGNUCErrorParser.regex.ErrorConflictingTypesFor=(.*?):(\\d+):(\\d+:)? [Ee]rror: (conflicting types for .*[`'"](.*)['"].*)
 CDTGNUCErrorParser.regex.GenericError=(.*?):(\\d+):(\\d+:)?\\s*(([Ee]rror)|(ERROR)): (.*)
+CDTGNUCErrorParser.regex.GenericError2=(.*?): (([Ee]rror)|(ERROR)): (.*)
```
jld01, line 70: It seems that `GenericError` catches an error on a specific line of code while `GenericError2` catches errors that relate to invocation of the source file as a whole. Could we make the new identifier more descriptive? InvocationError? SourceFileError? It's a minor point. I am approving this change anyway.

jld01 (approved):

core/org.eclipse.cdt.core/plugin.properties
Torbjorn-Svensson: I tried to combined them into one, but then it would never extract the line number properly.
I don't really care what it's named, so would you be satisfied with `InvocationError`?

jld01: `InvocationError` works for me.

jld01 (approved):

Issue #758, opened 2024-04-12
Improve CMake project behaviours for feature parity with Makefile projects


jld01:
There are a number of CDT behaviours available within _Makefile_ projects that are not available within _CMake_ projects:

- Presentation of the _Includes_ virtual resource containing include file paths within the _Project Explorer_
- Ability to define _Build Variables_ and _Environment_ which may be used to customize the behaviour of build tools
- Ability to select _Container Settings_, _Binary Parsers_ and _Error Parsers_ of relevance to the toolchain

This issue has been created for discussion on how we can work towards feature parity of _CMake_ projects with _Makefile_ projects.

ewaterlander:
I would say: make Core Build (CMake and Makefile) on par with Managed Build (classic Makefile).

ewaterlander:
A fundamental difference between Managed and Core build project is that Managed projects have a "C/C++ Build" section in the project properties and Core Build projects don't. Core build projects have the build settings in the launch configuration, and these settings are limited to selecting a toolchain and setting a build command. There is one launch configuration for Running and one for Debug, but this is hidden for the user. They are selected based on the Launch Mode in the LaunchBar. The Core Build launch configurations are not visible in the Run- and Debug Configurations windows, but only in the LaunchBar.

To get on par with Managed build the Core Build projects could also get a C/C++ Build section in the project properties, but without the Makefile generation stuff (Settings >Tool Settings).
Or the Core build launch configurations could be extended w.r.t. build settings. But I find a launch configuration a strange place to store build settings. I don't know why this was chosen in the past. Perhaps to make it simpler for the user.

Also w.r.t. launching the Core Build launch configurations are much more limited than the standard C/C++ launch configurations. I typically use the Core Build launch configuration only for building, and then a normal launch configuration for running and debug.

ewaterlander:
> * Ability to define _Build Variables_ and _Environment_ which may be used to customize the behaviour of build tools

W.r.t. environment variables, I think this should not be added. The Core Build flow was made for environments where command line is king. See [here](https://bugs.eclipse.org/bugs//show_bug.cgi?id=321443#c6). CDT is not the master of the build information. One user uses command line, another user uses Eclipse/CDT. And you have CI regressions. The environment should be the environment in which you start Eclipse.

[...]

PR #940, opened 2024-11-18
Update Main Tab for Core Build local launch configurations.

ewaterlander:
A new Main Tab was created for Core Build local projects based on the Main Tab used for the classic Managed Build projects. It adds these features:

* Option to select a different binary.
* Option to control launch pre-builds.

The default value for the binary is empty string. With empty string the behaviour for binary selection stays the same as it was.

The project name is fixed and cannot be changed. A Core Build launch configuration is created with the project and tied to it.

This change relates to #758. It affects all Core Build projects, including CMake projects.

### DIFF
--- a/debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchConfigDelegate.java
+++ b/debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchConfigDelegate.java
@@ -10,6 +10,7 @@
  *******************************************************************************/
 package org.eclipse.cdt.debug.core.launch;
 
+import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.LinkedHashSet;
@@ -23,14 +24,18 @@ import org.eclipse.cdt.core.build.IToolChainManager;
 import org.eclipse.cdt.core.model.CoreModel;
 import org.eclipse.cdt.core.model.IBinary;
 import org.eclipse.cdt.debug.core.CDebugCorePlugin;
+import org.eclipse.cdt.debug.core.ICDTLaunchConfigurationConstants;
 import org.eclipse.cdt.debug.internal.core.InternalDebugCoreMessages;
 import org.eclipse.core.resources.IBuildConfiguration;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.resources.IProjectDescription;
 import org.eclipse.core.runtime.CoreException;
+import org.eclipse.core.runtime.IPath;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
+import org.eclipse.core.runtime.Path;
 import org.eclipse.core.runtime.Status;
+import org.eclipse.core.variables.VariablesPlugin;
 import org.eclipse.debug.core.ILaunchConfiguration;
 import org.eclipse.launchbar.core.target.ILaunchTarget;
 import org.eclipse.launchbar.core.target.launch.LaunchConfigurationTargetedDelegate;
@@ -121,6 +126,28 @@ public abstract class CoreBuildLaunchConfigDelegate extends LaunchConfigurationT
 		return exeFile;
 	}
 
+	/**
+	 * @since 8.8
+	 */
+	protected String getProgramPath(ILaunchConfiguration configuration, IBinary exeFile) throws CoreException {
+		String programName = configuration.getAttribute(ICDTLaunchConfigurationConstants.ATTR_PROGRAM_NAME, ""); //$NON-NLS-1$
+
+		if (programName.isBlank()) {
+			return Paths.get(exeFile.getLocationURI()).toString();
+		} else {
+			IPath path = new Path(
+					VariablesPlugin.getDefault().getStringVariableManager().performStringSubstitution(programName));
+			String fullPath;
+			if (path.isAbsolute()) {
+				fullPath = path.toOSString();
+			} else {
+				IProject project = getProject(configuration);
+				fullPath = project.getFile(path).getLocation().toOSString();
+			}
+			return fullPath;
+		}
+	}
+
 	@Override
 	protected IProject[] getBuildOrder(ILaunchConfiguration configuration, String mode) throws CoreException {
 		// 1. Extract project from configuration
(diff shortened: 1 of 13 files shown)


ewaterlander:
A picture of the new main tab:

image

betamaxbandit:
Hi @ewaterlander ,
thanks for creating the change.

TBH, I'm not sure about this. Everything I've done recently in this area has gone away from having the main tab in the launch configuration delegate. However, I can see how it might be useful if you have projects that contain multiple binaries.

For projects that build a single binary, can you say what the usecase for this is please? After all, if there is only a single binary then it can be selected automatically.

Can you explain what you mean by "Option to control launch pre-builds.".

Cheers John

ewaterlander:
> For projects that build a single binary, can you say what the usecase for this is please?

In our case a binary is built that doesn't run on the local system. The launch starts a local simulator. The simulator will load the binary and run it. So we put the path to simulator in the "C/C++ application" text field, and based on arguments in the Arguments tab the simulator knows what to load.

For plain single binary local programs you just leave the "C/C++ application" text field empty and it will be selected automatically.
And indeed it is helpful for multiple binaries, because currently the automatic selection will pick the first one found.

> Can you explain what you mean by "Option to control launch pre-builds.".

You can disable the automatic build before launching, and only trigger a build by manually pushing the build button. This can save you time if you want to do multiple runs and you are only changing input data.

The classic local C/C++ launch configuration also has these features. I think it is good to keep these.

ewaterlander:
Update: Instead of adding a new class CoreBuildMainTab2 I updated old class CoreBuildMainTab.

betamaxbandit:
Hi @ewaterlander ,
thanks for the extra details.
For your use case I guess this makes sense.
I'll review the code, hopefully early next week. I need to check how/if this impacts the changes I've made in my client code.
Cheers John

launch/org.eclipse.cdt.launch/src/org/eclipse/cdt/launch/ui/CMainTab2.java
```diff
@@ -93,7 +93,7 @@ public class CMainTab2 extends CAbstractMainTab {
 	 */
 	protected Combo fCoreTypeCombo;
 
-	private final boolean fDontCheckProgram;
+	private boolean fDontCheckProgram;
```
betamaxbandit, line 96: Q: since removing the "final" specifier, I think this variable should be initialised here.

launch/org.eclipse.cdt.launch/src/org/eclipse/cdt/launch/ui/corebuild/CoreBuildMainTab.java
```diff
@@ -10,73 +10,51 @@
...
-
-		label = new Label(comp, SWT.NONE);
-		label.setText("Project:");
-
-		projectName = new Text(comp, SWT.READ_ONLY | SWT.BORDER);
-		projectName.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false));
-
-		setControl(comp);
-	}
+public class CoreBuildMainTab extends CMainTab2 {
```
betamaxbandit, line 21: O: I'm not sure about this change.
CMainTab2 says "This class is not intended to be subclassed."

launch/org.eclipse.cdt.launch/src/org/eclipse/cdt/launch/ui/corebuild/CoreBuildMainTab.java
```diff
@@ -10,73 +10,51 @@
...
 
+	/*
+	 * A Core Build launch configuration is created immediately upon the Core Build project creation.
+	 * It cannot be created by hand and it is not duplicatable and can't be renamed.
+	 * The launch configuration is tied to the project. The project name may not be changed.
+	 */
 	@Override
-	public void setDefaults(ILaunchConfigurationWorkingCopy configuration) {
-		// none
+	protected void createProjectGroup(Composite parent, int colSpan) {
```
betamaxbandit, line 29: O: I'd prefer your previous patch where you used CoreBuildMainTab2 because this change will change the behaviour of the CoreBuildMainTab that I use.

betamaxbandit (changes requested):
Look good apart from my concern with CoreBuildMainTab

debug/org.eclipse.cdt.debug.core/src/org/eclipse/cdt/debug/core/launch/CoreBuildLaunchConfigDelegate.java
```diff
@@ -121,6 +126,28 @@ protected IBinary getBinary(ICBuildConfiguration buildConfig) throws CoreExcepti
...
+			return Paths.get(exeFile.getLocationURI()).toString();
+		} else {
+			IPath path = new Path(
+					VariablesPlugin.getDefault().getStringVariableManager().performStringSubstitution(programName));
+			String fullPath;
+			if (path.isAbsolute()) {
+				fullPath = path.toOSString();
+			} else {
+				IProject project = getProject(configuration);
+				fullPath = project.getFile(path).getLocation().toOSString();
```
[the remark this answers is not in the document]
ewaterlander: I think getProject() will only return null when the resource is the workspace, which will never be the case here.

launch/org.eclipse.cdt.launch/src/org/eclipse/cdt/launch/ui/CMainTab2.java
ewaterlander: I agree. I will add it.

launch/org.eclipse.cdt.launch/src/org/eclipse/cdt/launch/ui/corebuild/CoreBuildMainTab.java
ewaterlander: That line was written in 2008, before Core Build was active. I thought it would be best to reuse the class. Otherwise I would have to duplicate a lot of code.

launch/org.eclipse.cdt.launch/src/org/eclipse/cdt/launch/ui/corebuild/CoreBuildMainTab.java
ewaterlander: OK, no problem. Will change it back.

ewaterlander: Should I amend the commit and force push, or add commits to this PR?

betamaxbandit (approved):
LGTM

ewaterlander:
> LGTM

Hi @betamaxbandit , thanks for the review!
Hi @jonahgraham , would you like to merge it?

jonahgraham (approved):

jonahgraham:
Thanks @ewaterlander for the improvement. I have added the noteworthy tag because we should highlight this change. It may be best to add an entry to the N&N document for CDT 12 in the API section: https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CDT-12.0.md#api-changes-current-and-planned

jonahgraham:
Thanks @betamaxbandit for the review.

ewaterlander:
> I have added the noteworthy tag because we should highlight this change. It may be best to add an entry to the N&N document for CDT 12 in the API section: https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CDT-12.0.md#api-changes-current-and-planned

I'm working on one other PR that I will upload soon, that will add the Debugger, Source, and Common tabs to the Core Build launch config. When that is done I think the local Core Build launch configuration is finally mature. It will make CMake (and Makefile) local projects usable out of the box.
After that I will add an entry to N&N and also contribute to documentation for Core Build projects.

PR #941, opened 2024-11-19
Use launch configuration environment attributes in Core Build local d…

ewaterlander:
…ebug.

The Core Build local debug launch uses the environment attributes from the new environment tab that was added in #901.


ewaterlander:
This change is not needed. It is already taken care of by GdbLaunch.

PR #959, opened 2024-12-11
Core Build project without binaries.

ewaterlander:
Do not check for binaries in the project directory if the user defined
a binary in the Main tab. This supports the case where the user
disables auto build and selects a binary outside the project directory.
Otherwise the binaries check would stop the launch with the message
that no binaries are found in the project.


ewaterlander:
would you like to review this change?

jonahgraham:
Thanks @ewaterlander

For the record the failing tests are unrelated to this change (https://github.com/eclipse-cdt/cdt/issues/816)

PR #960, opened 2024-12-12
Description of the Core Build System.

ewaterlander:
This is an initial description of the Core Build System for the user manual.
I think it should be placed just before the explanation of creating CMake and Makefile projects.

### DIFF
--- a/doc/org.eclipse.cdt.doc.user/topics_Getting_Started.xml
+++ b/doc/org.eclipse.cdt.doc.user/topics_Getting_Started.xml
@@ -4,6 +4,7 @@
           <topic label="Preparing the Workbench" href="getting_started/cdt_w_prepare_workbench.htm"/>
           <topic label="Creating a simple application" href="getting_started/cdt_w_basic.htm"/>
           <topic label="Creating a Makefile project" href="getting_started/cdt_w_newproj.htm"/>
+          <topic label="Core Build System" href="getting_started/core_build_system.htm"/>
           <topic label="Creating a CMake project" href="getting_started/new_cmake_proj.htm"/>
           <topic label="Importing an existing project" href="getting_started/cdt_w_import.htm"/>
           <topic href="getting_started/cdt_w_newcpp.htm" label="Creating a C++ file"/>
--- a/doc/org.eclipse.cdt.doc.user/src/getting_started/cdt_w_newproj.adoc
+++ b/doc/org.eclipse.cdt.doc.user/src/getting_started/cdt_w_newproj.adoc
@@ -142,8 +142,7 @@ at any time by right-clicking on the project in the *Project Explorer*
 view and clicking Properties.
 
 xref:cdt_w_newcpp.adoc[image:ngnext.gif[Next
-Icon,width=16,height=16]] xref:cdt_w_newcpp.adoc[*Next: Creating your {cpp}
-file*]
+Icon,width=16,height=16]] xref:core_build_system.adoc[*Next: Core Build System*]
 
 For more information about:
 
--- a/doc/org.eclipse.cdt.doc.user/src/getting_started/core_build_system.adoc
+++ b/doc/org.eclipse.cdt.doc.user/src/getting_started/core_build_system.adoc
@@ -1,0 +1,77 @@
+////
+Copyright (c) 2000, 2025 Contributors to the Eclipse Foundation
+This program and the accompanying materials
+are made available under the terms of the Eclipse Public License 2.0
+which accompanies this distribution, and is available at
+https://www.eclipse.org/legal/epl-2.0/
+
+SPDX-License-Identifier: EPL-2.0
+////
+
+// pull in shared headers, footers, etc
+:docinfo: shared
+
+// support image rendering and table of contents within GitHub
+ifdef::env-github[]
+:imagesdir: ../../images
+:toc:
+:toc-placement!:
+endif::[]
+
+// enable support for button, menu and keyboard macros
+:experimental:
+
+// Until ENDOFHEADER the content must match adoc-headers.txt for consistency,
+// this is checked by the build in do_generate_asciidoc.sh, which also ensures
+// that the checked in html is up to date.
+// do_generate_asciidoc.sh can also be used to apply this header to all the
+// adoc files.
+// ENDOFHEADER
+
+= Core Build System
+
+The Core Build System is the successor of the legacy Managed Build
+System.
+
+The classical Makefile projects are based on the Managed Build System.
+Managed Build projects can generate a Makefile based on GUI controlled
+build configurations in the project properties. The user can
+manage the separate build configurations for Run and Debug. There is a
+notion of "active" build configuration, which is the one that gets
+built when the build button is pressed.
+
+The Core Build System is the basis for CMake projects and Core Build
+Makefile projects. The CDT LaunchBar plays a central role and is used
+for building and launching the program. The philosophy is "Build for
+launch". It simplifies the flow a lot.
+
+These are the main differences compared with the Managed Build System:
+
+* No build configurations. No build file (Makefile or
+  CMakeLists.txt) generation. The user writes it by hand. A
+  "helloworld" example build file can be created for new projects.
+* Each new project is created with a launch configuration tied to it.
+* Build settings are presented in the launch configuration Build Settings tab.
+* The LaunchBar determines Run or Debug mode for building. There is no
+  concept of an active build configuration as there is in Managed
+  Build projects.
+* Auto discovery. The project's macro definitions and include paths,
+  needed for indexing, are extracted from the build step. For CMake
+  projects, data is obtained from the generated JSON file. For Makefile
+  projects, data is extracted from the build step console output.
+* Simpler wizards. Creation of a CMake project is via a simple one page
+  wizard. Using existing code is simply done by creating a new project
+  on top of the existing code. It is not necessary to import the project
+  (menu:File[Import...]).
+
+Core Build connects CDT projects to industry standard external build
+systems like CMake, Makefile, and Meson. The removal of the CDT
+specific build configurations allow Core Build projects to be
+easily be shared with other IDEs, command-line use, and Continuous
+Integration (CI) flows.
+
+Thanks to the auto discovery, the indexing of the C/C++ code has
+improved, because all compilation flags are known per file. This is especially
+beneficial for projects that use multiple toolchains in one project.
+
+icon:arrow-circle-right[] link:pass:[../html/new_cmake_proj.html][Next: Creating a CMake project]


ewaterlander:
Hi @jonahgraham , @betamaxbandit , and @jld01 ,
here is my brief description of the Core Build System.

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
```diff
@@ -0,0 +1,57 @@
...
+:toc:
+:toc-placement!:
+endif::[]
+
+// enable support for button, menu and keyboard macros
+:experimental:
+
+= Core Build System
+
+The Core Build System is the successor of the Managed Build System.
```
jld01, line 24: Does everyone agree with this messaging? Personally, I view Core Build more as an alternative rather than a successor at present. But that could change over time.

Cc: @jonahgraham

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
```diff
@@ -0,0 +1,57 @@
...
+manage the separate build configurations for Run and Debug. There is a
+notion of "active" build configuration, which is the one that gets
+built when the build button is pressed.
+
+The Core Build System is the basis for CMake projects and Core Build
+Makefile projects. The CDT LaunchBar plays a central role and is used
+for building and launching the program. The philosophy is "Build for
+launch". It simplified the flow a lot.
+
+These are the main differences with the old Managed Build System:
```
jld01, line 38: I would be inclined to drop "old" here.

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
ruspl-afed: Since I totally missed the moment when MBS became a "legacy" and Core Build became a 
"favorite" here is the discussion to collect opinions https://github.com/eclipse-cdt/cdt/discussions/961

ewaterlander: I got this idea when reading old discussions with Dough Schaefer. You are right. We keep Managed Build. I will change it.

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
ewaterlander: I will do that.

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
```diff
@@ -0,0 +1,57 @@
...
+
+* No build configurations. No build file (Makefile or
+  CMakeLists.txt) generation. The user writes it by hand. A
```
ruspl-afed, line 41: > The user writes it by hand.

Accented like a benefit. However, for an IDE it may sound like a degradation in a path to a "just colored editor"

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
jonahgraham: CDT "deprecated" MBS many many years ago and @ewaterlander's messaging is in line with what was decided then - see this release review from 2017 as an example https://projects.eclipse.org/projects/tools.cdt/releases/9.3.0/review - quoted here:

> We have a few systems that are duplicated legacy and new. We have the new Core Build system that should be a long term replacement for the Managed Build system. Until we can do everything with the new one, we have both systems remaining.

jonahgraham: And the follow on release review quoted here https://projects.eclipse.org/projects/tools.cdt/releases/9.4.0/review:

> As Core Build matures, and with Managed Build largely unmaintained, we have started planning for Managed Build's deprecation. Integrators adding in new build integrations should look at Core Build.

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
```diff
@@ -0,0 +1,57 @@
...
+* Build settings are in the launch configuration Build Settings tab.
+* The LaunchBar determines Run or Debug mode for building. There is no
+  more concept of an active build configuration like in the Managed
+  Build projects.
+* Auto discovery. The project's macro definitions and include paths,
+  needed for indexing, are extracted from the build step. For CMake
+  projects this is done via the generated JSON file. For Makefile
+  projects it is extracted from the build step console output.
+* Using existing code is simply done by creating a new project on top
+  of the existing code. No more project import.
```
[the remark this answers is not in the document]
jonahgraham: File -> Open projects was done by Mat Booth a while ago:  https://github.com/eclipse-cdt/cdt/commit/a957e8121df1937ae27243389bb7286fcd933c6b

ruspl-afed:
>  The Managed Build projects are proprietary to Eclipse CDT and requires Eclipse CDT to build

I would say it looks more like Eclipse CDT "vendor lock". Yes, this is a problem, and many people don't like this.

ewaterlander:
> > The Managed Build projects are proprietary to Eclipse CDT and requires Eclipse CDT to build
> 
> I would say it looks more like Eclipse CDT "vendor lock". Yes, this is a problem, and many people don't like this.

For us a big advantage is that with Core Build you can use the same (c)make flow for IDE users, command-line users, and nightly regressions. And the real showstopper for MBS was that it is not usable for multi-core systems. Building a GUI to set compiler flags for multiple binaries was not doable, or at least hard to do. People can easily live without makefile generation. The essential parts of the IDE are code navigation and debugging.

ruspl-afed:
> And the real showstopper for MBS was that it is not usable for multi-core systems.

Why so? One can use multiple projects (or multiple configuration that is less usable). 
How Core Build helps to work with multi-core systems?

ewaterlander:
> > And the real showstopper for MBS was that it is not usable for multi-core systems.
> 
> Why so? One can use multiple projects (or multiple configuration that is less usable). How Core Build helps to work with multi-core systems?

It is possible to split into multiple MBS projects. I actually tried that, before I discovered Core Build. But that adds a lot of overhead. I work in a Linux-only environment, where command-line is still king. If I had to ask programmers to split their single Makefile projects, they would not use the IDE.

ruspl-afed:
>  If I had to ask programmers to split their single Makefile projects, they would not use the IDE.

Ah, this is to bring the lost back into the fold of the IDE. This is a good argument, I hope it will have an effect.

ewaterlander:
> Ah, this is to bring the lost back into the fold of the IDE. This is a good argument, I hope it will have an effect.

Like Jonah already said, being able to share the project is a key difference. Not only with other IDEs, but also with command-line. We encourage people to use the IDE, but they are free to use command line, or both. If you're able to easily switch between IDE and command-line you can use the best of both worlds.

jonahgraham (approved):
LGTM - @jld01 if you are happy with the updated language (especially around removal of "old" and use of "legacy") then we are good to go here.

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
```diff
@@ -0,0 +1,66 @@
...
+launch". It simplified the flow a lot.
+
+These are the main differences with the Managed Build System:
+
+* No build configurations. No build file (Makefile or
+  CMakeLists.txt) generation. The user writes it by hand. A
+  "helloworld" example build file can be created for new projects.
+* A project gets immediately a launch configuration tied to it.
+* Build settings are in the launch configuration Build Settings tab.
+* The LaunchBar determines Run or Debug mode for building. There is no
```
ruspl-afed, line 46: > The LaunchBar determines Run or Debug mode for building.

Does it influence any tool arguments?

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
ruspl-afed: I think this is answered in comments. Since we cannot invent something better than MBS, the only way is to let user write build file by hand.

ruspl-afed (approved):
LGTM

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
ruspl-afed: Thank you for the answer @jonahgraham

jld01:
@ewaterlander thanks for the document. I have added some suggestions for consistency and flow. Let me know what you think.

We also need to introduce the HTML file that will be generated at `html/core_build_system.html` to the Eclipse help system by declaring a topic such as:
```
<topic label="Core Build System" href="getting_started/cdt_w_newproj.htm"/>
```
in https://github.com/eclipse-cdt/cdt/blob/main/doc/org.eclipse.cdt.doc.user/topics_Getting_Started.xml

ewaterlander:
> @ewaterlander thanks for the document. I have added some suggestions for consistency and flow. Let me know what you think.

I think your suggestions are addressed in the second commit. Did I miss something?

> We also need to introduce the HTML file that will be generated at `html/core_build_system.html` to the Eclipse help system by declaring a topic such

I will add it.

jonahgraham (approved):

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
jonahgraham: AFAICT this is the only remaining comment. @jld01 any final thoughts, we can discuss on Wednesday call if you want. I added this to agenda in  https://github.com/eclipse-cdt/cdt/discussions/967

ewaterlander:
> I don't see where these comments are addressed. At present, the text seems to be in a conversational style. My suggested edits are intended to give the text more of a documentation style in keeping with the other help pages. I would prefer that these changes (or similar) be applied before merging.

Let's talk about this in the meeting, because I don't see what I missed. And also about the style.

ewaterlander:
I have included all John's suggestions.
I squashed it into a single commit and rebased on main.

doc/org.eclipse.cdt.doc.user/getting_started/cdt_w_newproj.htm
jld01, line 1: I think this file will be replaced by a generated version as part of #995 so some liaison will be required concerning which PR is merged first.

doc/org.eclipse.cdt.doc.user/src/asciidoc/core_build_system.adoc
jld01, line 1: This file will need to be placed in the `getting_started` folder is merged after #995.

doc/org.eclipse.cdt.doc.user/topics_Getting_Started.xml
```diff
@@ -4,6 +4,7 @@
           <topic label="Preparing the Workbench" href="getting_started/cdt_w_prepare_workbench.htm"/>
           <topic label="Creating a simple application" href="getting_started/cdt_w_basic.htm"/>
           <topic label="Creating a Makefile project" href="getting_started/cdt_w_newproj.htm"/>
+          <topic label="Core Build System" href="html/core_build_system.html"/>
```
jld01, line 7: Precise name/location of generated HTML will be different if merged after #995.

jld01 (approved):
Thanks for the updates, @ewaterlander.

jonahgraham:
I am very close to finishing #995 - I will rebase this change once I merge #995 into the new location of the file. I think #995 should be merged first to maintain a "nicer" history on this file.

doc/org.eclipse.cdt.doc.user/getting_started/cdt_w_newproj.htm
jonahgraham: I have rebased this on #995.

doc/org.eclipse.cdt.doc.user/topics_Getting_Started.xml
jonahgraham: I have rebased this on #995.

jonahgraham:
Thanks @ewaterlander for this valuable write-up.

ewaterlander:
> I have rebased and will merge assuming acceptance by CI.
> 
> @ewaterlander at the CDT call you mentioned that loading each page was very slow, when you get a chance can you test it again with my recent changes to see if I have resolved the issue. If you still have the problem, can you create a new issue please?

Hi @jonahgraham ,

Thanks for doing the rebase!

I noticed today that Eclipse for committers requires the new glibc, as was announced. Unfortunately at the moment I'm working on machines that don't meet the requirement. Upgrade is in progress, but that may take some time. Perhaps I can get access to an upgraded machine sooner. I noticed the delay also on WSL, so I can test that right now.

ewaterlander:
> I noticed today that Eclipse for committers requires the new glibc, as was announced. Unfortunately at the moment I'm working on machines that don't meet the requirement. Upgrade is in progress, but that may take some time. Perhaps I can get access to an upgraded machine sooner. I noticed the delay also on WSL, so I can test that right now.

On WSL the pages now show without delay. It has improved a lot. Thanks!

I did have a fight with proxy settings. For Eclipse Committers to do it's updates I need to have the proxies defined. But when I run CDT from Eclipse Committers I can't see any help page when proxies are defined. In the launch configuration I can set empty string values to http_proxy and https_proxy, but that doesn't work. They need to be undefined. It is not possible to undefine variables in the Environment tab. So to see the help I have to restart Eclipse Committers without proxies defined.

jonahgraham:
Thanks for the feedback @ewaterlander 

Its been a while since I worked with proxy settings, but it sounds like the proxy is getting involved when destination is localhost (as it is when using help inside the IDE). I don't know how to resolve this, but just to confirm, does it look like:

image

> I noticed today that Eclipse for committers requires the new glibc, as was announced.

You can stick with the older version of Eclipse for committers, it should be just fine.

Issue #972, opened 2024-12-16
Preferences and UI elements incorrectly remain hidden when using Core Build to debug


jonahgraham:
**Describe the bug**

There are a number of preferences and other UI elements that are only enabled once that flow started. For example, Preferences -> C/C++ -> Debug -> GDB is only visible after launching a GDB debug session:

![image](https://github.com/user-attachments/assets/9dba9418-dd10-4c0d-85d2-55d0d318724e)

With core build and launch, those preferences are not being displayed.

**To Reproduce**
1. Create a new workspace
2. Observe no Preferences -> C/C++ -> Debug -> GDB  visible
3. Create a C/C++ CMake project, build and debug it
4. Still no GDB preference page visible
5. Create a C/C++ Application Launch Configuration and launch it
6. Preferences -> C/C++ -> Debug -> GDB  visible

**Expected behavior**
Once anything with GDB is touched, make the UI visible.

jonahgraham:
The key line of code that needs to run is https://github.com/eclipse-cdt/cdt/blob/481f08beb43d41de50f3465004b29c0c2fd9ace5/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/GdbLaunchDelegate.java#L92

jonahgraham:
I have added my preferred fix in #974

PR #973, opened 2024-12-16
Ensure DSF GDB preference pages and UI elements are available in Core Build

jonahgraham:
When using Core Build, the GdbLaunchDelegate is not used to create the launch. Therefore in the core launch we need to enable the activity to show all the DSF GDB specific UI in another place.

This PR adds one of those places, but it may very well be there are other places that this should be added.

Fixes https://github.com/eclipse-cdt/cdt/issues/972


jonahgraham:
I prefer the fix in #974

PR #978, opened 2024-12-18
Synchronize the access to LRUCache

ramele1907:
- a wizard that is creating multiple cdt projects encounters ConcurrentModificationException

### DIFF
--- a/build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/language/settings/providers/AbstractLanguageSettingsOutputScanner.java
+++ b/build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/language/settings/providers/AbstractLanguageSettingsOutputScanner.java
@@ -19,13 +19,14 @@ import java.net.URI;
 import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.function.Function;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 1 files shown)


jonahgraham:
Hi @ramele1907 - thanks for the PR. Unfortunately the code as provided doesn't compile. See [build ouput](https://github.com/eclipse-cdt/cdt/actions/runs/12388745382/job/34945566946?pr=978#step:7:3465) for more details.

`findPathInProjectCache` has some more complicated semantics as it is a map of maps, so the inner map probably needs synchronizing too. e.g. add additional `Collections.synchronizedMap` around the `new LRUCache` in findPathInProject.

Because # of reads far outweigh number of writes and you indicated that multiple projects are being created simultaneously, did you consider concurrency instead of synchronizing all accesses to the map? Those caches were added in 9e7b5beaa96f6741df5b25336b750279cb58458b to dramatically speed up the code, and I think that adding synchronized around each read access may have an unintentional slowdown. If you can demonstrate that `synchronizedMap` doesn't (seriously) negatively affect performance (see the lined commit for some stats) then we don't need to consider concurrency. 

It would help if there is a way to reproduce the problem so that we can see the problem in action.

jonahgraham:
@ramele1907 Would you like to have a look at this and see if you can resolve the compile errors and other comments I made in my preliminary review https://github.com/eclipse-cdt/cdt/pull/978#issuecomment-2564481329

ramele1907:
@jonahgraham I resolved the compile errors. I will try to use concurrency instead of synchronization to avoid performance issues. 

Regarding the reproduction topic I will try to create a mock plugin with a wizard that creates multiple projects.

jonahgraham:
Also did you see this comment from my original review https://github.com/eclipse-cdt/cdt/pull/978#issuecomment-2564481329?

> `findPathInProjectCache` has some more complicated semantics as it is a map of maps, so the inner map probably needs synchronizing too. e.g. add additional `Collections.synchronizedMap` around the `new LRUCache` in findPathInProject.

I am relying on you to verify that my statement is true (that the inner maps need addressing)

ramele1907:
#

> `findPathInProjectCache` has some more complicated semantics as it is a map of maps, so the inner map probably needs synchronizing too. e.g. add additional `Collections.synchronizedMap` around the `new LRUCache` in findPathInProject.
> 
I agree, the access to `findPathInProject` might also lead to a `ConcurrentModificationException`.
> 
> I'm pleased you found concurrency easy enough to do. Did you calculate the performance issue? Was it significant?
> 
I saw no performance gain in my product after switching from synchronization to concurrency. I think the computation of the values was the bottleneck addressed in  [9e7b5be](https://github.com/eclipse-cdt/cdt/commit/9e7b5beaa96f6741df5b25336b750279cb58458b)
>
>Wonderful! I recently did something similar by creating an example CMake project wizard that isn't shipped to customers, but demonstrates how to make custom wizards. See https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CDT-12.0.md#cmake-support-easier-for-extenders-and-isvs
>
Thank you for your suggestion! I will come back with updates on this topic.

jonahgraham:
@ramele1907 I see that you did a merge, but that didn't ended up getting pushed as the latest commit in this PR. I have merged and rebased your change onto the latest HEAD commit so that the workflows can run.

jonahgraham:
> Thank you for your suggestion! I will come back with updates on this topic.

We are nearly out of time for CDT 12.0. Are there more updates coming on this topic?

jonahgraham:
> > Thank you for your suggestion! I will come back with updates on this topic.
> 
> We are nearly out of time for CDT 12.0. Are there more updates coming on this topic?

@ramele1907 Please re-open or comment if you plan to work on this more.

PR #979, opened 2024-12-18
Add improved CBS launch configuration for local applications to N&N

ewaterlander:


jonahgraham (approved):

jonahgraham:
I think the exact location within the file of this N&N entry may not be correct, but I'll resolve that closer to release date when we have more content.

PR #986, opened 2024-12-28
Update to latest versions of build tools

jonahgraham:
Beta versions of maven plugins are omitted from this upgrade.

A new version of Tycho brings in an update ECJ which means some of the class files are different, hence the need to bump versions of bundles too.

A new calendar year brings an update to Copyrights


jonahgraham:
This is good to go now, the test failures are known https://github.com/eclipse-cdt/cdt/labels/flakytests

PR #989, opened 2025-01-04
[#942] GDBControl discards ILaunchConfiguration in its constructor

### LINKED ISSUE #942: `GDBControl` discards `ILaunchConfiguration` parameter in its constructor
**Describe the bug**
There is a deep [and terrible] hierarchy of `GDBControl` where every successor carefully passes the  `ILaunchConfiguration` parameter. But, in the end, no one uses it.

**To Reproduce**
https://github.com/eclipse-cdt/cdt/blob/9b6bb0711f8210a5fa519cef941a9435c31deb40/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/service/command/GDBControl.java#L242

**Expected behavior**
It would be great to have `ILaunchConfiguration` available at least for successors

**Screenshots**
```java
	/**
	 * @since 4.1
	 */
	protected GDBControl(DsfSession session, boolean useThreadAndFrameOptions, ILaunchConfiguration config,
			CommandFactory factory) {
		super(session, useThreadAndFrameOptions, factory);
	}

```

**Version Information:**

* main branch

ruspl-afed:
add protected final field to be available for successors 
adding final accessor method may break Embedded CDT

Fixes #942

### DIFF
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.cdt.dsf.gdb;singleton:=true
-Bundle-Version: 7.1.500.qualifier
+Bundle-Version: 7.2.0.qualifier
 Bundle-Activator: org.eclipse.cdt.dsf.gdb.internal.GdbPlugin
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime,
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/service/command/GDBControl.java
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/service/command/GDBControl.java
@@ -113,6 +113,11 @@ public class GDBControl extends AbstractMIControl implements IGDBControl {
 	private static final int STATUS_CODE_COMMAND_TIMED_OUT = 20100;
 
 	/**
+	 * @since 7.2
+	 */
+	protected final ILaunchConfiguration config;
+
+	/**
 	 * Event indicating that the back end process has started.
 	 */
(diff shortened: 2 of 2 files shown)


ruspl-afed:
Interesting, why adding a field to a class considered as a major change
```
Error:  [API ERROR] File GDBControl.java at line 117: The field org.eclipse.cdt.dsf.gdb.service.command.GDBControl.config has been added to a class (location: /work/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/service/command/GDBControl.java)
Error:  [API ERROR] File MANIFEST.MF at line 6: The major version should be incremented in version 7.2.0, since API breakage occurred since version 7.1.300 (location: /work/dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF)
```

jonahgraham (commented):
> Interesting, why adding a field to a class considered as a major change
> 
> ```
> Error:  [API ERROR] File GDBControl.java at line 117: The field org.eclipse.cdt.dsf.gdb.service.command.GDBControl.config has been added to a class (location: /work/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/service/command/GDBControl.java)
> Error:  [API ERROR] File MANIFEST.MF at line 6: The major version should be incremented in version 7.2.0, since API breakage occurred since version 7.1.300 (location: /work/dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF)
> ```

Please see https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/Evolving-Java-based-APIs-2.md#:~:text=Adding%20an%20API%20field%20to%20an%20API%20class%20that%20is%20extended%20by%20Clients%20is%20dangerous%20in%20two%20respects%3A

So this change requires a major version bump as it is possibly breaking change. That is ok in principle for CDT 12 as that is a major version. But you have to decide whether CDT API consumers are likely to be affected negatively more or less than how you benefit from this change.

I am +0 on this change as on a quick examination of the open source and closed source extensions of GDBControl I don't actually see anyone who has their own field called `config`

ruspl-afed:
> So this change requires a major version bump as it is possibly breaking change

Thank you for review @jonahgraham 
Now I see. I don't think this small improvement is worth the breaking changes. FYI @jld01 
Let's close this PR

PR #990, opened 2025-01-04
[#943] Provide API for "org.eclipse.cdt.debug.gdbjtag.core.jtagDevice"

### LINKED ISSUE #943: Provide API to reference "org.eclipse.cdt.debug.gdbjtag.core.jtagDevice" launch attribute
Currently we have `IGDBJtagConstants#ATTR_JTAG_DEVICE_ID` for "org.eclipse.cdt.debug.gdbjtag.core.jtagDeviceId" but nothing similar for "org.eclipse.cdt.debug.gdbjtag.core.jtagDevice"

This leads to a duplicated declarations even for Eclipse CDT code base

1. https://github.com/eclipse-cdt/cdt/blob/9b6bb0711f8210a5fa519cef941a9435c31deb40/jtag/org.eclipse.cdt.debug.gdbjtag.core/src/org/eclipse/cdt/debug/gdbjtag/core/GDBJtagDSFFinalLaunchSequence.java#L111
2. https://github.com/eclipse-cdt/cdt/blob/9b6bb0711f8210a5fa519cef941a9435c31deb40/jtag/org.eclipse.cdt.debug.gdbjtag.ui/src/org/eclipse/cdt/debug/gdbjtag/ui/GDBJtagDSFDebuggerTab.java#L83

It would be great to add API for this launch attribute

ruspl-afed:
* add `IGDBJtagConstants.ATTR_JTAG_DEVICE` and remove duplicated definitions

Fixes #943

### DIFF
--- a/jtag/org.eclipse.cdt.debug.gdbjtag.core/META-INF/MANIFEST.MF
+++ b/jtag/org.eclipse.cdt.debug.gdbjtag.core/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.cdt.debug.gdbjtag.core;singleton:=true
-Bundle-Version: 10.8.200.qualifier
+Bundle-Version: 10.9.0.qualifier
 Bundle-Activator: org.eclipse.cdt.debug.gdbjtag.core.Activator
 Bundle-Localization: plugin
 Require-Bundle: org.eclipse.core.runtime,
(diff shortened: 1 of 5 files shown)


ruspl-afed:
@jonahgraham after reading both https://github.com/eclipse-cdt/cdt/blob/main/NewAndNoteworthy/CHANGELOG-API.md and https://bugs.eclipse.org/bugs/show_bug.cgi?id=566462 I still miss the rationale why this was done. 
Perhaps, because we want to switch to `jtagDeviceId` and stop spreading the `jtagDevice`. But in this case we need to first stop spreading the `jtagDevice` through our own codebase. Perhaps we should provide a common way to resolve `GDBJtagDeviceContribution` instead of duplicating the same code for `core` and `UI`

@jld01 could you please provide your opinion here?

jonahgraham:
> Perhaps we should provide a common way to resolve `GDBJtagDeviceContribution` instead of duplicating the same code for `core` and `UI`

This exists already in `GDBJtagDeviceContributionFactory.getInstance().findByDeviceId` - the duplication around the call sites of findByDeviceId are because the attributes are stored in a different location in UI and core parts (ILaunchConfiguration.getAttribute vs a map called attributes - which is populated by a combination of the launch configuration and other attributes). The other duplication is handling of the deprecated attribute, but that code could eventually be deleted and is only needed for compatibility with old launch configurations.

Issue #952, opened 2024-12-03
Provide accessor for launch attributes map field in `StartOrRestartProcessSequence_7_0`


ruspl-afed:
Currently `StartOrRestartProcessSequence_7_0` uses the `Map<String, Object> fAttributes` assigned in its constructor but does not allow successors to access this map.

It would be good to have protected accessor for `Map<String, Object> fAttributes` and remove duplications in successors.

ruspl-afed:
> this style (of having duplicate private fields with same information) is prevalent throughout CDT DSF

Yes, true. do you know the rationale behind this?
My particular case is to decrease the amount of code one need to write while extending `StartOrRestartProcessSequence*` chain, nothing extra special. I thought that since `attributes` field is final it could be accessible for successors, may be in immutable form. Of cause I can follow the CDT DSF style and copy&paste more code.

jonahgraham:
> Yes, true. do you know the rationale behind this?

No, I don't know the rationale behind this design choice.

ruspl-afed:
Closed, see discussion at #991

PR #991, opened 2025-01-04
[#952] Provide access for launch attributes map field

ruspl-afed:
* make `fAttributes` protected and remove code duplications

### DIFF
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/META-INF/MANIFEST.MF
@@ -3,7 +3,7 @@ Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-Vendor: %providerName
 Bundle-SymbolicName: org.eclipse.cdt.dsf.gdb;singleton:=true
-Bundle-Version: 7.1.500.qualifier
+Bundle-Version: 7.2.0.qualifier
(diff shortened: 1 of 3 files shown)


jonahgraham (commented):
I am +0 on this too (see #989) even though I see some extenders do have their own field.  This basically causes new warnings in extenders who have their own `fAttributes` already.

ruspl-afed:
Thank you for review @jonahgraham 
It looks like it is better for all to leave the current implementation [dying] as it is, because any change that looks like improvement (for example, because it reduces the code duplication) causes potential issues. 
I tried to integrate back changes I did for debug scenario on our product that allowed me to "heal" DSF code via removing duplications (thousands of LOC) for both CDT and EmbedCDT. I tried to select some of them (the most laconic and simple) and had no luck. Perhaps, this is clear sign that CDT DSF is mature enough to change anything and it is easier to write an alternative [with less cumbersome and more maintainable code] than gradually improve the existing one.

I'm fine to close this PR as well. But before doing so, I would like to learn how do you see the future of DSF @jonahgraham @jld01

jonahgraham:
> I would like to learn how do you see the future of DSF

I see it similar to many other parts of CDT + Eclipse ecosystem. An absolutely critical part of the CDT offerings, under-maintained but generally very good.

ruspl-afed:
Thank you for the guidance @jonahgraham, I've learned that I need to be much more careful before suggesting changes to the CDT codebase.
I will close this PR.

Issue #1000, opened 2025-01-14
CDT CMake Improvements


betamaxbandit:
The document* describes a number of bug fixes and enhancements to CMake support.

[20250226CDT+CMake+Improvements.pdf](https://github.com/user-attachments/files/18985707/20250226CDT%2BCMake%2BImprovements.pdf)

Items in the "API Breaking" column marked Yes are planned to be added to CDT 12.

*: formatting is affected by exporting from internal document management system.

betamaxbandit:
In summary, the following pull requests were completed as part of this activity (CDT CMake Improvements #1000):

 - #1010 addresses IDE-82683-REQ-017
 - #1046 addresses IDE-82683-REQ-012
 - #1052 related to #1046
 - #1076 addresses IDE-82683-REQ-016
 - #1084 addresses IDE-82683-REQ-024
 - #1090 addresses IDE-82683-REQ-013
 - #1095 addresses IDE-82683-REQ-011
 
Not done:
 - #1067 addresses IDE-82683-REQ-004 & IDE-82683-REQ-005. Paused because specification needs more work.

PR #1010, opened 2025-01-15
Make CMakeBuildConfiguration API

betamaxbandit:
For ISV integration/extension to CDT CMake, it is necessary to extend several non-API classes (eg: CMakeBuildConfiguration, CMakeBuildConfigurationProvider). This would cause "Discouraged access: The type XXX is not API" warnings.

The classes have now been made API so can be extended without warnings.

Addresses Issue: CDT CMake Improvements #1000, IDE-82683-REQ-017 Extending CMakeBuildConfiguration

### DIFF
--- a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/Messages.java
+++ b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/Messages.java
@@ -8,12 +8,18 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  *******************************************************************************/
-package org.eclipse.cdt.cmake.core.internal;
+package org.eclipse.cdt.cmake.core;
 
 import org.eclipse.osgi.util.NLS;
 
+/**
+ * @noextend This class is not intended to be subclassed by clients.
+ * @noinstantiate This class is not intended to be instantiated by clients.
+ * @noreference This class is not intended to be referenced by clients.
+ * @since 1.6
+ */
 public class Messages extends NLS {
-	private static final String BUNDLE_NAME = "org.eclipse.cdt.cmake.core.internal.messages"; //$NON-NLS-1$
+	private static final String BUNDLE_NAME = "org.eclipse.cdt.cmake.core.messages"; //$NON-NLS-1$
 	public static String CMakeBuildConfiguration_Building;
 	public static String CMakeBuildConfiguration_BuildingIn;
 	public static String CMakeBuildConfiguration_BuildingComplete;
@@ -22,8 +28,6 @@ public class Messages extends NLS {
 	public static String CMakeBuildConfiguration_Configuring;
 	public static String CMakeBuildConfiguration_ExitFailure;
 	public static String CMakeBuildConfiguration_NotFound;
-	public static String CMakeBuildConfiguration_ProcCompCmds;
-	public static String CMakeBuildConfiguration_ProcCompJson;
 	public static String CMakeBuildConfiguration_Failure;
 	public static String CMakeErrorParser_NotAWorkspaceResource;
 	static {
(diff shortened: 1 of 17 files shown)


cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/Messages.java
```diff
@@ -8,12 +8,15 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  *******************************************************************************/
-package org.eclipse.cdt.cmake.core.internal;
+package org.eclipse.cdt.cmake.core;
 
 import org.eclipse.osgi.util.NLS;
 
+/**
+ * @since 1.6
```
jonahgraham, line 16: Messages are not API, so I am adding this
```suggestion
 * @noextend This class is not intended to be subclassed by clients.
 * @noinstantiate This class is not intended to be instantiated by clients.
 * @noreference This class is not intended to be referenced by clients.
 * @since 1.6
```

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakeConsoleWrapper.java
```diff
@@ -21,9 +21,10 @@
 /** Intercepts output to a console and forwards its error stream to a stream that does error parsing for processing.
 
  * @author Martin Weber
- *
+ * @noinstantiate
+ * @noextend
```
jonahgraham, line 25: Unneeded on internal classes.
```suggestion
```

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakePropertiesController.java
```diff
@@ -40,8 +40,10 @@
  * A {@code ICMakePropertiesController} that monitors modifications to the project properties that force
  * us to delete file CMakeCache.txt to avoid complaints by cmake.
  * @author Martin Weber
+ * @noinstantiate
+ * @noextend
```
jonahgraham, line 44: Unneeded on internal classes.
```suggestion
```

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CommandDescriptorBuilder.java
```diff
@@ -28,16 +29,18 @@
  * and/or for performing the actual build of a project.
  *
  * @author Martin Weber
+ * @noinstantiate
+ * @noextend
```
jonahgraham, line 33: Unneeded on internal classes
```suggestion
```

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/IOsOverridesSelector.java
```diff
@@ -19,8 +19,9 @@
  * project build.
  *
  * @author Martin Weber
+ * @noimplement
```
jonahgraham, line 22: Unneeded on internal interfaces
```suggestion
```

Was the intention to make this interface API? If so, it needs to be moved to a non-internal namespace.

jonahgraham (changes requested):
I am applying these suggested changes and will push an update very soon.

jonahgraham (approved):

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/IOsOverridesSelector.java
betamaxbandit: No!
Without this change I get:
"The type org.eclipse.cdt.cmake.core.internal.IOsOverridesSelector is not visible"
It was merely to facilitate moving CMakeBuildConfiguration to API.

betamaxbandit:
LGTM
Thanks @jonahgraham

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/IOsOverridesSelector.java
jonahgraham: > "The type org.eclipse.cdt.cmake.core.internal.IOsOverridesSelector is not visible"

I can only see this error if I don't include the `public` on the line below. This review comment was only about the `@noimplement` which shouldn't be applied to internal classes/interfaces as they are implicitly `@noimplement` already

PR #1020, opened 2025-01-15
use UTF-8 as default encoding

astrograph:
the rest of Eclipse now uses UTF-8 as default encoding.

The tm.terminal code reverts to null when the encoding begins with "Default" and this leads to UTF-8 being selected from Charset.defaultEncoding in stead of ISO-8859 when the displayed default value is selected.


astrograph:
> Thanks @astrograph
> 
> The next release is a major release, so this change is well timed as we have a bit more freedom to change defaults, etc.
> 
> I think a little more needs to be done to change this, for example this line looks like it needs updating too
> 
> https://github.com/eclipse-cdt/cdt/blob/a9a7e8b02716e5a70a4c7fa35c05d2e0cc563998/terminal/plugins/org.eclipse.tm.terminal.view.ui/src/org/eclipse/tm/terminal/view/ui/panels/AbstractExtendedConfigurationPanel.java#L615
> 
> There are a few other places that look like an incorrect default is in place too? [search for `ISO-8859-1` in `terminal/**/*.java`](https://github.com/search?q=repo%3Aeclipse-cdt%2Fcdt+path%3A%2F%5Eterminal%5C%2F%2F+ISO-8859-1+language%3AJava&type=code&l=Java)
> 
> WDYT?

I added a commit with some more changes to reflect the new default encoding

jonahgraham:
There are some IT issues that I am trying to resolve + wait on Eclipse IT team for. Please disregard any failed builds or other failed checks for now.

jonahgraham (approved):
Looks great. Thanks @astrograph for the improvement!

I added a N&N entry

Once CI passes I can submit this.

Issue #1002, opened 2025-01-14
Add comments from GitHub actions


jonahgraham:
Fix/cleanup auto-commenting in PRs from GitHub actions:

- [ ] Add a documentation link when docs are updated (see https://github.com/eclipse-cdt/cdt/pull/995#issuecomment-2578782580)
- [ ] Review JUnit test reports which sometimes generates a comment and sometimes doesn't, this may be simply an out of date configuration for JUnit, have a look at SWT's for some ideas https://github.com/eclipse-platform/eclipse.platform.swt/blob/master/.github/workflows/junit.yml

jonahgraham:
tests are being reported against license checks, 

![image](https://github.com/user-attachments/assets/dfacf570-9255-4907-8321-f881638d95e0)

this is related to OP and this also hides the problem of the license checks failing:

![image](https://github.com/user-attachments/assets/09125c7a-c3cc-4165-abc5-e1e8992d439b)


cc: @betamaxbandit

jonahgraham:
I was made aware of the license check fails because @betamaxbandit was getting emails about the fails, and I wasn't. I had turned off GitHub actions notifications in https://github.com/settings/notifications

![image](https://github.com/user-attachments/assets/348cd5ba-f9f9-4bfb-bfe4-d9c4197121a4)

jonahgraham:
For the JUnit comment doesn't always appear in the PR, there is a message in the build ([link](https://github.com/eclipse-cdt/cdt/actions/runs/12794213715/job/35668845873#step:4:35)):

```
2025-01-15 17:59:29 +0000 - publish -  INFO - There is no pull request for commit 08d6339ba5fdd7b4b469f637d019bd8f65ca8f06
```

jonahgraham:
> this is related to OP and this also hides the problem of the license checks failing:

It doesn't actually hide it, the complete failure of the workflow seems to stop it being fully published to the PR, and then the JUnit results attaches itself randomly to a job, sometimes this one that was failing.

jonahgraham:
_Originally posted by @github-actions[bot] in https://github.com/eclipse-cdt/cdt/issues/1033#issuecomment-2595072188_

Here are test results, but they say number of tests have gone down, which is weird. Some more investigation may be needed here.

## Test Results
   590 files   -  11     590 suites   - 11   12m 59s ⏱️ +11s
10 079 tests  - 121  10 056 ✅  - 121  23 💤 ±0  0 ❌ ±0 
[...]

PR #1023, opened 2025-01-15
Update path to dash-licenses

jonahgraham:
When Eclipse Dash got moved to its own org, the path the shared workflow needed to be updated. This commit does that

Part of #1002


jonahgraham:
Don't merge yet - it is convenient that this is failing at the moment because it is allowing me to test why PRs aren't showing the failure, e.g. I see:

image

But the license check hasn't actually run successfully.

PR #1031, opened 2025-01-16
Revert "workaround test results appearing in incorrect spot"

jonahgraham:
This reverts commit ae3aebcf7f0ddd82b2f83c9963e5621e37beb136.

Part of #1002


jonahgraham:
The loss of the `check_run` functionality is worse than the side effect of it applied to the wrong place. We lose diffs on test runs and no where does main/branch test results get reported.

PR #1032, opened 2025-01-16
Shorten the name of the licensing check

jonahgraham:
This helps it fit in the UI space GitHub gives it

Part of #1002


jonahgraham:
This change shortens the display so it looks nicer and text is more readable as it isn't cut off

Before:

image

After:

image

ruspl-afed (approved):
LGTM

PR #1033, opened 2025-01-16
Search debugger first in selected CBS toolchain.

### LINKED ISSUE #1008: Cannot launch CMake debug session using Local launch target
On Windows, when attempting to launch a debug session using the Local launch target, the launch fails with the message:

Error with command: gdb --version

To reproduce (always)
Prerequisites:
gcc toolchain registered with core build system (on Windows automatic if msys64 installed).
CMake binary is added to environment path - see cmake/org.eclipse.cdt.cmake.ui.tests/manualTests/Bug579242_manual_tests.md
1) Create a CMake helloworld program.
2) Set Launch Mode to Debug .
3) Build program
4) Launch program
  **Expected**: debug session launched successfully and program suspended on breakpoint at main.
  **Actual**: debug session launch failure.
    Problem Occurred popup:
    Error with command: gdb --version
    Cannot run program "gdb": Launching failed  

The problem appears to be introduced with commit:
  commit f259d60aa3f6bbbd120acfd461e35f45fef6f7c0
  Author: ewaterlander <102143930+ewaterlander@users.noreply.github.com> 2024-12-10 17:08:08 

I think the problem is due to org.eclipse.cdt.dsf.gdb.internal.launching.CoreBuildLocalDebugLaunchDelegate.launch(ILaunchConfiguration, String, ILaunch, IProgressMonitor), the following code was removed:

		IToolChain toolChain = buildConfig.getToolChain();
		Path gdbPath = toolChain.getCommandPath(Paths.get("gdb")); //$NON-NLS-1$
		gdbLaunch.setGDBPath(gdbPath != null ? gdbPath.toString() : "gdb"); //$NON-NLS-1$

which means the toolchain's path is not used to set the gdb path.

@ewaterlander

ewaterlander:
For Core Build System local debug target. If there is no absolute path set in the Debugger tab of the launch configuration, try to find the debugger first in the selected toolchain. If the debugger is not found in the toolchain, let GdbLaunch search in PATH.
If an absolute path is set, GdbLaunch will use that.

Fixes #1008

### DIFF
--- a/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/internal/launching/CoreBuildLocalDebugLaunchDelegate.java
+++ b/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/internal/launching/CoreBuildLocalDebugLaunchDelegate.java
@@ -10,18 +10,22 @@
  *******************************************************************************/
 package org.eclipse.cdt.dsf.gdb.internal.launching;
 
+import java.nio.file.Path;
+import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.ExecutionException;
 
 import org.eclipse.cdt.core.build.ICBuildConfiguration;
+import org.eclipse.cdt.core.build.IToolChain;
 import org.eclipse.cdt.debug.core.launch.CoreBuildLaunchConfigDelegate;
 import org.eclipse.cdt.dsf.concurrent.DataRequestMonitor;
 import org.eclipse.cdt.dsf.concurrent.ImmediateExecutor;
 import org.eclipse.cdt.dsf.concurrent.Query;
 import org.eclipse.cdt.dsf.concurrent.RequestMonitorWithProgress;
 import org.eclipse.cdt.dsf.concurrent.Sequence;
+import org.eclipse.cdt.dsf.gdb.IGDBLaunchConfigurationConstants;
 import org.eclipse.cdt.dsf.gdb.internal.GdbPlugin;
 import org.eclipse.cdt.dsf.gdb.internal.Messages;
 import org.eclipse.cdt.dsf.gdb.launching.GdbLaunch;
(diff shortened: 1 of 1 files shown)


ewaterlander:

Here is my fix for #1008

jonahgraham:
> This pull request removes 121 tests.

Please see https://github.com/eclipse-cdt/cdt/issues/1002#issuecomment-2595851157 for discussion about this new feature that comments tests results. I don't know why less tests were run, but the change doesn't seem to be the cause.

jonahgraham:
> > This pull request removes 121 tests.
> 
> Please see [#1002 (comment)](https://github.com/eclipse-cdt/cdt/issues/1002#issuecomment-2595851157) for discussion about this new feature that comments tests results. I don't know why less tests were run, but the change doesn't seem to be the cause.

Never mind - the difference is explained in the log, the maven run did not complete. I have restarted it.

jonahgraham (approved):
This looks reasonable to me. @betamaxbandit can you confirm if this fixes the regression you identified in #1008?

betamaxbandit (approved):
LGTM.

betamaxbandit:
> This looks reasonable to me. @betamaxbandit can you confirm if this fixes the regression you identified in #1008?

I haven't tested this.

@ewaterlander , did you have chance to test this fix on Windows?

jonahgraham:
@betamaxbandit and I did a screen share and tested this on Windows

jonahgraham:
Tests are failing, [catastrophically](https://github.com/eclipse-cdt/cdt/actions/runs/12806024495/job/35718813056?pr=1033#step:7:26646). I don't understand why and can't see that it is related to this change. But I need to investigate a bit before I merge to make sure this isn't somehow exposing some weird bug.

jonahgraham:
> Tests are failing, [catastrophically](https://github.com/eclipse-cdt/cdt/actions/runs/12806024495/job/35718813056?pr=1033#step:7:26646). I don't understand why and can't see that it is related to this change. But I need to investigate a bit before I merge to make sure this isn't somehow exposing some weird bug.

#1035 has the same issue, so I am confident not related to this change now.

ewaterlander:
> @betamaxbandit and I did a screen share and tested this on Windows

I did test it, but on Linux. I could see that the debugger was taken from the toolchain.

PR #1036, opened 2025-01-16
Update to latest versions of build tools

jonahgraham:
This is a follow-up to #986 where I missed this version update.


jonahgraham:
License checking is failing, probably because a bunch of IT is down at EF https://www.eclipsestatus.io/incident/496975?mp=true

jonahgraham:
The [IT outage](https://www.eclipsestatus.io/incident/496975?mp=true) means ECA can't pass so nothing can be merged:
image

Issue #1037, opened 2025-01-16
Some error thrown at dsf gdb tests after that no other tests are executed.


jonahgraham:
Some error thrown at dsf gdb tests after that no other tests are executed.
```
2025-01-16T16:41:44.0644358Z [INFO] org.eclipse.cdt.tests.dsf.gdb 2.3.100-SNAPSHOT ..... FAILURE [  8.004 s]
2025-01-16T16:41:44.0645257Z [INFO] [test-bundle] DSF-GDB multicorevisualizer test fragment 1.0.0-SNAPSHOT SKIPPED
2025-01-16T16:41:44.0646155Z [INFO] org.eclipse.cdt.debug.gdbjtag.core.tests 1.0.400-SNAPSHOT SKIPPED
2025-01-16T16:41:44.0646917Z [INFO] [test-bundle] Debug Core Memory Tests 0.1.0-SNAPSHOT SKIPPED
2025-01-16T16:41:44.0647777Z [INFO] [test-bundle] C/C++ Development Tools Tests Running Support Tests 1.0.0-SNAPSHOT SKIPPED
2025-01-16T16:41:44.0648634Z [INFO] org.eclipse.cdt.testing 8.1.500-SNAPSHOT ........... SKIPPED
2025-01-16T16:41:44.0649335Z [INFO] org.eclipse.cdt.testing 12.0.0-SNAPSHOT ............ SKIPPED
2025-01-16T16:41:44.0650040Z [INFO] org.eclipse.cdt.testing.repo 8.3.0-SNAPSHOT ........ SKIPPED
2025-01-16T16:41:44.0650728Z [INFO] ------------------------------------------------------------------------
2025-01-16T16:41:44.0651263Z [INFO] BUILD FAILURE
2025-01-16T16:41:44.0651713Z [INFO] ------------------------------------------------------------------------
2025-01-16T16:41:44.0652254Z [INFO] Total time:  21:50 min
2025-01-16T16:41:44.0652639Z [INFO] Finished at: 2025-01-16T16:41:44Z
2025-01-16T16:41:44.0653368Z [INFO] ------------------------------------------------------------------------
2025-01-16T16:41:44.1523155Z [ERROR] Failed to execute goal org.eclipse.tycho:tycho-surefire-plugin:4.0.10:test (default-test) on project org.eclipse.cdt.tests.dsf.gdb: An unexpected error occurred while launching the test runtime (process returned error code 13). 
```

[...]

PR #1042, opened 2025-01-17
Run DSF-GDB tests when releng may have changed

jonahgraham:
On a recent update to the target platform, the DSF-GDB tests did not run because none of the changes matched the dsf filter. Going forward, include running DSF tests when any releng changes happen.

This replaces commit 8d0642568c0. That commit ran the tests all the time, which was too slow.

Part of #1037


jonahgraham:
Known fails from #816

PR #1045, opened 2025-01-17
Restore the empty activity IDs

### LINKED ISSUE #1044: Refresh button appearing when it isn't enabled
The refresh button is not supposed to be visible unless explicitly turned on by the user, see the FAQ entry for example: https://github.com/eclipse-cdt/cdt/blob/main/FAQ/README.md#must-the-debugging-views-refresh-automatically

However, last week I submitted https://github.com/eclipse-cdt/cdt/pull/974 and it looks that caused the refresh button to be visible always, I am not quite sure how yet, but I am going to try to fix that so it achieves the desired effect.

Note this is a regression since CDT 11.6 introduced by #974

jonahgraham:
In #974 I removed activities we were not using anymore, but it turns out if LaunchUtils.enableActivity is called on a removed id, strange things happen, such as other activities incorrectly being enabled. This may be a bug in platform, or just undefined behavior. But because extenders do have calls to LaunchUtils.enableActivity in their code, leave the empty activity ids behind.

Fixes #1044

### DIFF
--- a/releng/org.eclipse.cdt/plugin.properties
+++ b/releng/org.eclipse.cdt/plugin.properties
@@ -17,3 +17,12 @@ cheatsheet.helloworld.desc=Create a Hello World Program in C++
 cheatsheet.helloworld.name=Hello World
 
 activity.name = Debug Update Modes
+activity.description.dsf = Debug Services Framework
+activity.name.dsf = CDT-DSF Debug Services Framework
+
+activity.description.dsfgdb = C/C++ debugging with the DSF GDB debugger
+activity.name.dsfgdb = CDT DSF-GDB - GDB Debugging
+activity.description.cdtdebug = Commands for C/C++ debugging
+activity.name.cdtdebug = CDT Debug - C/C++ Development Tools
+activity.description.cdigdb = C/C++ debugging with the CDI (legacy) GDB debugger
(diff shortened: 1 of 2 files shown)


jonahgraham:
Concretely, this PR makes sure that the refresh button doesn't appear in the UI when it isn't enabled explicitly by the user.

jonahgraham:
I have figured out *why* this change matters. Eclipse Platform UI has an optimization where if the number of enabled activity IDs == number of properly defined activity IDs then all activities are considered enabled ([link to code](https://github.com/eclipse-platform/eclipse.platform.ui/blob/745385a64d99bfcf8c2a72139c72ee1829599ef2/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/internal/activities/MutableActivityManager.java#L734-L737)). This caused refresh button to be enabled because we enabled the dsfgdb id ([link](https://github.com/eclipse-cdt/cdt/blob/ae9f14585683df6bfac9c10debaa78238a86bf09/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/gdb/launching/GdbLaunchDelegate.java#L94)) but had removed the ID from the defined IDs, so we hit this optimization in platform and the refresh button (and a bunch of other things presumably) got enabled.

jonahgraham:
Known fails from #816

PR #1046, opened 2025-01-17
Improve CMake build option handling and API

### LINKED ISSUE #818: Overhaul CMake build settings
Overhaul CMake build settings

org.eclipse.cdt.cmake.core.internal.CMakeBuildConfiguration currently uses a mixture of APIs to store settings relevant to the CMake build. These are:

1)  ICBuildConfiguration.setProperty/getProperty
  
2)  CMakeBuildConfiguration.getPropertiesController()

Having 2 settings APIs in use is confusing and inefficient.

1) org.eclipse.cdt.core.build.CBuildConfiguration.getSettings()
  protected Preferences getSettings()
  This is used to implement the *property API in ICBuildConfiguration, eg:

    ICBuildConfiguration.setProperty(String, String)
    ICBuildConfiguration.getProperty(String)
  
  It uses org.osgi.service.prefs.Preferences
  
  When a ICBuildConfiguration is created/accessed, this is central to storing the build configuration toolchain, launch mode etc.
  It is also used by the UI (CMakeBuildTab) to store CMake settings like the generator and -D define arguments, which can be set using the Launch Bar Launch Configuration dialog. For example:
  org.eclipse.cdt.cmake.ui.internal.CMakeBuildTab.initializeFrom(ILaunchConfiguration)
  
2) org.eclipse.cdt.cmake.core.internal.CMakeBuildConfiguration.getPropertiesController()
  private CMakePropertiesController getPropertiesController()  
    This uses a file to store settings:
      ".settings/CDT-cmake.yaml"
  This arrangement was put in place [1], but never really utilised.
  There are a lot of CMake settings stored in org.eclipse.cdt.cmake.core.properties.ICMakeProperties which are used during a CMake build.

It would make sense to rationalise the settings arrangement and do things using one of the APIs above.
I would vote for using the Preferences approach as it is the standard approach in Eclipse.

Questions
A) Will removing the YAML file have any impact on the cmake4eclipse (https://github.com/15knots/cmake4eclipse) integration in CDT?

B) While doing this work, it would be wise to consider the impact of any future implementation of CMake file API (https://github.com/eclipse-cdt/cdt/issues/419)

[1] https://github.com/eclipse-cdt/cdt/pull/814#issuecomment-2156104747
> Hi @15knots , Can you say what the purpose of the YAML file is please?
I planned to allowto persist option values for the cmake command-line on a per-projecz level in the yaml file. But when I found out that some of the cmake options are passed through totally undocumented properties from a project not being part of CDT (launchbar plugin) I gave up. You may safely delete the stuff dealing with that or find a way to combine both the YAML stettings and the launchbar settings. Martin

### LINKED ISSUE #1055: Overhaul CMake build options
While working on https://github.com/eclipse-cdt/cdt/pull/1046 I came across a few broken cmake build options.

1) fix typo in defintion and api names:
  org.eclipse.cdt.cmake.core.properties.ICMakeProperties.setWarnUnitialized(boolean)
2) remove cmake deprecated options
  org.eclipse.cdt.cmake.core.properties.ICMakeProperties.setWarnUnused(boolean)

This list is not exhaustive; there may be other items to address.

The option values are assigned here:
org.eclipse.cdt.cmake.core.internal.CommandDescriptorBuilder.makeCMakeCommandline(Path)

	public CommandDescriptor makeCMakeCommandline(Path toolChainFile) throws CoreException {
		List<String> args = new ArrayList<>();
		List<String> env = new ArrayList<>();

		// defaults for all OSes...
		args.add(CMakeBuildConfiguration.BUILD_COMMAND_DEFAULT);
		/* add general settings */
		if (cmakeProperties.isWarnNoDev())
			args.add("-Wno-dev"); //$NON-NLS-1$
		if (cmakeProperties.isDebugTryCompile())
			args.add("--debug-trycompile"); //$NON-NLS-1$
		if (cmakeProperties.isDebugOutput())
			args.add("--debug-output"); //$NON-NLS-1$
		if (cmakeProperties.isTrace())
			args.add("--trace"); //$NON-NLS-1$
		if (cmakeProperties.isWarnUnitialized())
			args.add("--warn-unitialized"); //$NON-NLS-1$
		if (cmakeProperties.isWarnUnused())
			args.add("--warn-unused"); //$NON-NLS-1$

cc: @jonahgraham

betamaxbandit:
Summary:

- Add some new API to make it easier for ISVs to provide defaults.
- Fully connect UI elements to CMake build process
- Add some missing UI elements (such as customizing generator)
- CMake generator default within CDT changed to Ninja

Details:

Add API to set CMake generator default (eg Ninja) ISV can set their desired CMake generator by overriding
`CMakeBuildConfiguration.getDefaultProperties`. ISVs can also further fine tune the build process by overriding
`CMakeBuildConfiguration.getDefaultProperties`

Remove API `IOsOverrides` and related code. `IOsOverrides` was a partial implementation to achieve builds in Docker containers, however the work was not complete and it the extra code was complicating some basic use cases of setting defaults

Add support for all generators to CMake build settings UI page by using a Combo instead of radio buttons. The non-deprecated generators that are built-in to CDT populate the Combo, but additional generators can be manually entered in the Combo.

Rename clean command to clean target to better reflect its use as the argument passed to cmake's --target command line.

Add all target for the argument passed to cmake's --target command line when doing a normal build.

Clarify usage of UI overrides and change the UI to be "use defaults" (i.e. invert the checkbox). This is a **breaking** change as it means user projects that were using UI overrides will revert to using defaults. This is done on purpose as so many little things have changed in CMake settings, that reverting to defaults on upgrade seems like a logical decision. In addition *use defaults* matches the other GUIs in Eclipse, for example the MBS build command settings.

Populate all defaults in getDefaultProperties() so that all CMake build settings are displayed as used (greyed out) and can be used as a starting point when editing settings.

Simplify some of the code in CMakeBuildTab.

Fix parsing of extra args so that quoted strings work.

Refactored manual tests document and brought it up to date.

Correct command line option for CMake's --warn-unused-vars

Correct command line option for CMake's --warn-uninitialized

Overall this is an API breaking change and the CHANGELOG-API.md has been updated with all the API changes in and around ICMakeProperties, including fixing typos in WarnUninitialized methods.

Fixes #1055
Fixes #818
Part of #1000


betamaxbandit:
Hi @jonahgraham ,
I've created a draft PR so we can iron out any issues before releasing to wider community.

WRT our discussion on the yaml file yesterday, as far as I can tell the org.eclipse.cdt.cmake.core.internal.CMakePropertiesController.save(ICMakeProperties) method is never called by CDT code, which means the yaml file will not exist for regular (non-ISV extended) CDT products.
Let me know if you think I should still consider the yaml file questions for an ISV modified CDT product.

You'll see that I've added an API for setting the cmake generator.
I've also used that API in the CMakeBuildConfiguration constructors and added logic so it's only done the first time. Maybe there is a better way of doing things just the first time. Anyway it means it's easy for vanilla CDT to use this mechanism and set the generator and have those values used in the UI (CMakeBuildTab.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
```diff
@@ -122,6 +127,30 @@ public CMakeBuildConfiguration(IBuildConfiguration config, String name, IToolCha
...
+		if (getProperty(CMAKE_PROPERTIES_INITIALSED) == null) {
+			ICMakeProperties properties = CMakePropertiesFactory.getProperties();
+			IOsOverrides windowsOverrides = properties.getWindowsOverrides();
+			windowsOverrides.setDefaultGenerator(CMakeGenerator.UnixMakefiles);
+
+			IOsOverrides linuxOverrides = properties.getLinuxOverrides();
+			linuxOverrides.setDefaultGenerator(CMakeGenerator.UnixMakefiles);
+			// reset(true) causes the CMake generator to be reset to it's default value.
+			properties.reset(true);
+			setDefaultCMakeProperties(properties);
```
jonahgraham, line 145: This isn't ok - you shouldn't have non-final methods called by constructor.

Can you change this around so that the interface has `getDefaultCMakeProperties` that extenders can override if they don't like the built-in defaults?

Here is a diff showing what I mean, but it isn't tested, so may require fixing up

<details>
<summary>expand for patch contents</summary>

```diff
diff --git a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
index fe7f365c83d..a0eeee42967 100644
--- a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
+++ b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
@@ -77,7 +77,6 @@ import org.osgi.service.prefs.Preferences;
  */
 public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMakeBuildConfiguration {
 
-	private static final String CMAKE_PROPERTIES_INITIALSED = "cmake.properties.initialised"; //$NON-NLS-1$
 	public static final String CMAKE_USE_UI_OVERRIDES = "cmake.use.ui.overrides"; //$NON-NLS-1$
 	public static final boolean CMAKE_USE_UI_OVERRIDES_DEFAULT = false;
 	public static final String CMAKE_GENERATOR = "cmake.generator"; //$NON-NLS-1$
@@ -93,8 +92,6 @@ public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMa
 	// lazily instantiated..
 	private ICMakePropertiesController pc;
 
-	ICMakeProperties defaultProperties;
-
 	private Map<IResource, IScannerInfo> infoPerResource;
 	/**
 	 * whether one of the CMakeLists.txt files in the project has been modified and saved by the
@@ -116,7 +113,6 @@ public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMa
 
 		ICMakeToolChainManager manager = Activator.getService(ICMakeToolChainManager.class);
 		toolChainFile = manager.getToolChainFileFor(getToolChain());
-		setDefaultCMakeProperties();
 	}
 
 	public CMakeBuildConfiguration(IBuildConfiguration config, String name, IToolChain toolChain) {
@@ -127,30 +123,21 @@ public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMa
 			ICMakeToolChainFile toolChainFile, String launchMode) {
 		super(config, name, toolChain, launchMode);
 		this.toolChainFile = toolChainFile;
-		//		// moule
-		setDefaultCMakeProperties();
-	}
-
-	private void setDefaultCMakeProperties() {
-		// Only set defaults on initial construction
-		if (getProperty(CMAKE_PROPERTIES_INITIALSED) == null) {
-			ICMakeProperties properties = CMakePropertiesFactory.getProperties();
-			IOsOverrides windowsOverrides = properties.getWindowsOverrides();
-			windowsOverrides.setDefaultGenerator(CMakeGenerator.UnixMakefiles);
-
-			IOsOverrides linuxOverrides = properties.getLinuxOverrides();
-			linuxOverrides.setDefaultGenerator(CMakeGenerator.UnixMakefiles);
-			// reset(true) causes the CMake generator to be reset to it's default value.
-			properties.reset(true);
-			setDefaultCMakeProperties(properties);
-		}
 	}
 
 	@Override
-	public void setDefaultCMakeProperties(ICMakeProperties defaultProperties) {
-		this.defaultProperties = defaultProperties;
-		setProperty(CMAKE_GENERATOR, getOsOverrides(defaultProperties).getGenerator().getCMakeName());
-		setProperty(CMAKE_PROPERTIES_INITIALSED, Boolean.TRUE.toString());
+	public ICMakeProperties getDefaultCMakeProperties() {
+		ICMakeProperties properties = CMakePropertiesFactory.createProperties();
+		IOsOverrides windowsOverrides = properties.getWindowsOverrides();
+		windowsOverrides.setDefaultGenerator(CMakeGenerator.UnixMakefiles);
+
+		IOsOverrides linuxOverrides = properties.getLinuxOverrides();
+		linuxOverrides.setDefaultGenerator(CMakeGenerator.UnixMakefiles);
+		// reset(true) causes the CMake generator to be reset to it's default value.
+		properties.reset(true);
+
+		setProperty(CMAKE_GENERATOR, getOsOverrides(properties).getGenerator().getCMakeName());
+		return properties;
 	}
 
 	/**
@@ -192,7 +179,7 @@ public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMa
 			}
 
 			//			ICMakeProperties cmakeProperties = getPropertiesController().load();
-			ICMakeProperties cmakeProperties = getPropertiesController().load(defaultProperties);
+			ICMakeProperties cmakeProperties = getPropertiesController().load(getDefaultCMakeProperties());
 			runCMake |= !Files.exists(buildDir.resolve("CMakeCache.txt")); //$NON-NLS-1$
 
 			// Causes CMAKE_BUILD_TYPE to be set according to the launch mode
@@ -334,7 +321,7 @@ public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMa
 			project.deleteMarkers(ICModelMarker.C_MODEL_PROBLEM_MARKER, false, IResource.DEPTH_INFINITE);
 
 			//			ICMakeProperties cmakeProperties = getPropertiesController().load();
-			ICMakeProperties cmakeProperties = getPropertiesController().load(defaultProperties);
+			ICMakeProperties cmakeProperties = getPropertiesController().load(getDefaultCMakeProperties());
 			CommandDescriptorBuilder cmdBuilder = new CommandDescriptorBuilder(cmakeProperties,
 					new SimpleOsOverridesSelector());
 			CommandDescriptor command = cmdBuilder.makeCMakeBuildCommandline(getCleanCommand());
@@ -436,7 +423,7 @@ public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMa
 	@SuppressWarnings("unchecked")
 	public <T> T getAdapter(Class<T> adapter) {
 		if (ICMakePropertiesController.class.equals(adapter)) {
-			//			return (T) pc;
+			//						return (T) pc;
 			return (T) getPropertiesController();
 		}
 		return super.getAdapter(adapter);
diff --git a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/ICMakeBuildConfiguration.java b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/ICMakeBuildConfiguration.java
index 5769704217e..913d3e369bc 100644
--- a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/ICMakeBuildConfiguration.java
+++ b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/ICMakeBuildConfiguration.java
@@ -1,3 +1,13 @@
+/*******************************************************************************
+ * Copyright (c) 2015, 2025 QNX Software Systems and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
 package org.eclipse.cdt.cmake.core;
 
 import org.eclipse.cdt.cmake.core.properties.ICMakeProperties;
@@ -14,5 +24,5 @@ public interface ICMakeBuildConfiguration {
 	 * {@link CMakeBuildConfiguration#CMAKE_GENERATOR}. The CMake generator property can then be accessed using
 	 * the {@link ICBuildConfiguration#getProperty(String)}.
 	 */
-	void setDefaultCMakeProperties(ICMakeProperties defaultProperties);
+	ICMakeProperties getDefaultCMakeProperties();
 }
diff --git a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/properties/CMakePropertiesFactory.java b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/properties/CMakePropertiesFactory.java
index 5d679ded86f..8c5c5302ceb 100644
--- a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/properties/CMakePropertiesFactory.java
+++ b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/properties/CMakePropertiesFactory.java
@@ -1,9 +1,22 @@
+/*******************************************************************************
+ * Copyright (c) 2015, 2025 QNX Software Systems and others.
+ *
+ * This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License 2.0
+ * which accompanies this distribution, and is available at
+ * https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *******************************************************************************/
 package org.eclipse.cdt.cmake.core.properties;
 
 import org.eclipse.cdt.cmake.core.internal.properties.CMakePropertiesBean;
 
+/**
+ * @since 2.0
+ */
 public class CMakePropertiesFactory {
-	public static ICMakeProperties getProperties() {
+	public static ICMakeProperties createProperties() {
 		return new CMakePropertiesBean();
 	}
 }
```
</details>

jonahgraham (changes requested):
> WRT our discussion on the yaml file yesterday, as far as I can tell the org.eclipse.cdt.cmake.core.internal.CMakePropertiesController.save(ICMakeProperties) method is never called by CDT code, which means the yaml file will not exist for regular (non-ISV extended) CDT products.
> Let me know if you think I should still consider the yaml file questions for an ISV modified CDT product.

I don't think you need to consider the yaml file - thank you for pointing out that `save` was never called. I guess this was incomplete work once upon a time.

I guess that you will start calling save now though (in a subsequent commit)? Perhaps after migrating from yaml to something else?

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
jonahgraham: Addendum, 

extenders can do something like this:

```java

	@Override
	public ICMakeProperties getDefaultCMakeProperties() {
		// get the built-in CDT defaults
		ICMakeProperties properties = super.getDefaultCMakeProperties();
		
		// provide my customizations (I don't know what Renesas/extenders may actually want to do here)
		properties.setWarnUnitialized(true);

		return properties;
	}

```

jonahgraham: And finally - maybe ;) - I think calling:

```
		setProperty(CMAKE_GENERATOR, getOsOverrides(properties).getGenerator().getCMakeName());
```

from within this method is incorrect as that method has side effects by changing project preferences too. It may be best to call `setProperty` on the return value of `getDefaultCMakeProperties`:

```java

	private ICMakeProperties getDefaultPropertiesInternal() {
		ICMakeProperties properties = getDefaultCMakeProperties();
		setProperty(CMAKE_GENERATOR, getOsOverrides(properties).getGenerator().getCMakeName());
		return properties;
	}
```

and then in build/clean:

```java
			ICMakeProperties cmakeProperties = getPropertiesController().load(getDefaultPropertiesInternal());
```

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
```diff
@@ -591,17 +632,9 @@ private class SimpleOsOverridesSelector implements IOsOverridesSelector {
...
 			// see comment in CMakeIndexerInfoConsumer#getFileForCMakePath()
-			final String os = Platform.getOS();
-			if (Platform.OS_WIN32.equals(os)) {
-				overrides = cmakeProperties.getWindowsOverrides();
-			} else {
-				// fall back to linux, if OS is unknown
-				overrides = cmakeProperties.getLinuxOverrides();
-			}
-			return getUiOrOsOverrides(overrides);
+			return getUiOrOsOverrides(getOsOverrides(cmakeProperties));
```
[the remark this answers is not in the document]
betamaxbandit: Apologies, that was some clumsy refactoring last thing on a Friday when I needed to finish work.

betamaxbandit (commented):
Hi @jonahgraham ,
thanks for taking the time to review this.

I like all your suggestions here and think the structure is better the way you have suggested.

However, it removes a key requirement. Each time the build is performed with this patch, it uses the default values. If the user changes the generator value in the UI, that change is ignored. Hence the need for the CMAKE_PROPERTIES_INITIALSED.

I'm struggling to come up with a better way of doing things at the moment, so if you have any ideas I'd be grateful of them please.

betamaxbandit:
> Hi @jonahgraham , thanks for taking the time to review this.
> 
> I like all your suggestions here and think the structure is better the way you have suggested.
> 
> However, it removes a key requirement. Each time the build is performed with this patch, it uses the default values. If the user changes the generator value in the UI, that change is ignored. Hence the need for the CMAKE_PROPERTIES_INITIALSED.
> 
> I'm struggling to come up with a better way of doing things at the moment, so if you have any ideas I'd be grateful of them please.

Maybe using CMakePropertiesController.save(ICMakeProperties) will help with the issue I mentioned in my previous comment about CMAKE_PROPERTIES_INITIALSED. But as well as saving the props, I'd need to do some other stuff in the load, like:

setProperty(CMAKE_GENERATOR, getPlatformOsOverrides(properties).getGenerator().getCMakeName());

jonahgraham:
@betamaxbandit I think if the load/save are matched then the problem is resolved - right now it seems a bit mismatched between the incomplete yaml and the launch config properties. Can we get it down to only *one* place that user information is stored?

betamaxbandit:
Hi @jonahgraham ,

I have implemented all your suggestions AFAIK.

These are some notes about the status of this PS. Sorry it's taking so long to do this, but I'm facing a number of issues with this. I wonder if it would be easier to split into smaller tickets?

0) Current block
Getting the correct defaults when the user opens the UI.
After creating a LC with run launch mode, when then switching to debug mode, the correct defaults are not used.
  Can this be done as a separate change?

1) LC names do not include launch mode
CoreBuildLocalLaunchConfigProvider does not create different launch configurations based on the run mode. When debug launch mode is selected and the LC is opened, I'd expect it to create a LC with debug appended to the name. And similarly, when run launch mode is selected and the LC is opened, I'd expect it to create a LC with run appended to the name.
However, the cbuildconfig does append the launch mode to it's name, so there can be 2 different cbuildconfig (debug and run), but only a single LC.

Q) Should run and debug launch configurations share the same or different names? 

Q) Why does the CMakeBuildTab use properties and not LC attrs to save values?

2) YAML file.
Do we need this? Adding in a save will generate it and complicate things as they are.
Having the yaml file is an extra feature so should be done in a different ticket (eg add portability or similar).

jonahgraham:
@betamaxbandit let's have a chat about this in a call and then we'll record any outcomes here.

ewaterlander:

This issue was raised first in the discussion in #814

Then my conclusion was that CMakePropertiesController could be deprecated, but I restored it, because it would break usage by betamaxbandit.

The Core Build System Build Settings tab stores values in the preferences and not in the Launch Configuration, because then it can store different CMake arguments for Run and Debug. In Managed Build projects you can also set different Make arguments for the Run and the Debug build configuration. I think this is a good feature.

When you change launch mode you keep seeing the same launch configuration. This is always. The tabs change, but it is all stored in the same .launch file.

CMakePropertiesController made no difference for launch mode.

ewaterlander:
> Q) Should run and debug launch configurations share the same or different names?
> 
> Q) Why does the CMakeBuildTab use properties and not LC attrs to save values?
> 
>     2. YAML file.
>        Do we need this? Adding in a save will generate it and complicate things as they are.
>        Having the yaml file is an extra feature so should be done in a different ticket (eg add portability or similar).

CBS projects still have separate build configurations for Run and Debug like MBS projects, only they are hidden for the user. Instead of explicitly setting an active build configuration in MBS projects, the build configuration is implicitly selected by changing the launch mode in CBS projects.

Build configurations store their settings in the preferences. The methods are there. I see no reason to do it differently.
Like I said in my previous comment, changing launch mode is only a different view on the same launch configuration.

CMake settings are build settings, not launch settings. I think we should not make separate launch configurations for different build settings. Same is true for Make settings in the CBS Makefile projects.

We don't need the YAML file. We can use the preferences. The infrastructure is there.

betamaxbandit:
Hi @ewaterlander ,
Thanks for your comments.

> This issue was raised first in the discussion in https://github.com/eclipse-cdt/cdt/pull/814

I'll update that when I complete this change.

I'm not planning changing the way the build settings are stored. I was just thinking aloud.

I plan to deprecate CMakePropertiesController.

Cheers John

betamaxbandit:
Hi @jonahgraham ,
I've updated my draft PR. I need to add the JUnit test and do further testing tomorrow, but I think it is nearly there now. Hope you agree?
Cheers John

jonahgraham:
@betamaxbandit , @Kummallinen  and I had a meeting/review on this today. In short this is progressing in the correct direction. Along with the review, a part of the outcome of this review is we decided having a fully fleshed out example of an ISV CMake customized project is a useful to both demonstrate for other ISVs and to ensure that the API we create has a workable use case. I have provided that in #1052

debug/org.eclipse.cdt.debug.core/META-INF/MANIFEST.MF
betamaxbandit, line 1: I will remove this change in my next PS

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
```diff
@@ -162,7 +203,8 @@ public IProject[] build(int kind, Map<String, String> args, IConsole console, IP
 				runCMake = true;
 			}
 
-			ICMakeProperties cmakeProperties = getPropertiesController().load();
+			ICMakeProperties cmakeProperties = getCMakeProperties();
```
betamaxbandit, line 206: @jonahgraham ,
Now that getPropertiesController() is no longer called, the Runnable cmakeCacheDirtyMarker param to CMakePropertiesController is not setup and thereby this code is not executed:

deleteCMakeCache = true;

I'm not sure if this matters.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/properties/CMakePropertiesBean.java
```diff
@@ -172,4 +176,13 @@ public void reset(boolean resetOsOverrides) {
 			windowsOverrides.reset();
 		}
 	}
+
+	@Override
+	public IOsOverrides getPlatformOsOverrides(ICMakeProperties cmakeProperties) {
```
betamaxbandit, line 181: From your verbal feedbac, this is item:
3) getPlatformOsOverrides static? DONE

cmake/org.eclipse.cdt.cmake.core.tests/src/org/eclipse/cdt/cmake/core/CMakeBuildConfigurationTests.java
```diff
@@ -0,0 +1,201 @@
+/*******************************************************************************
```
betamaxbandit, line 1: From your verbal feedbac, this is item:
1) copyright - new test.  DONE

betamaxbandit (commented):
From your verbal feedback, this is item:
2) code cmakebuildtab 158, remove side effects DONE

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
```diff
@@ -124,6 +126,45 @@ public CMakeBuildConfiguration(IBuildConfiguration config, String name, IToolCha
...
+		 */
+		IOsOverrides osOverrides = properties.getPlatformOsOverrides(properties);
+		// CMAKE_GENERATOR
+		String cmakeGeneratorCmakeDefault = osOverrides.getGenerator().getCMakeName();
+		// CMAKE_ARGUMENTS
+		String extraArgsCmakeDefault = osOverrides.getExtraArguments().stream().map(String::trim)
+				.collect(Collectors.joining(" ")); //$NON-NLS-1$
+
+		String useUiOverrides = getProperty(CMAKE_USE_UI_OVERRIDES);
+		if (Boolean.parseBoolean(useUiOverrides)) {
```
betamaxbandit, line 152: I've integrated the change you emailed; removed setProperty

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
```diff
@@ -582,50 +624,33 @@ public void shutdown() {
...
-				if (!buildCommand.isBlank() && !BUILD_COMMAND_DEFAULT.equals(buildCommand)) {
-					retVal.setUseDefaultCommand(false);
-					retVal.setCommand(buildCommand);
-				}
-			}
-			return retVal;
+			return cmakeProperties.getPlatformOsOverrides(cmakeProperties);
 		}
 	} // SimpleOsOverridesSelector
+
```
betamaxbandit, line 634: I've integrated the change you emailed; added getDefaultProperties

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakePropertiesController.java
```diff
@@ -41,6 +41,7 @@
  * us to delete file CMakeCache.txt to avoid complaints by cmake.
  * @author Martin Weber
  */
+@Deprecated
```
jonahgraham, line 44: Lets delete instead of deprecating internal code.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
jonahgraham: All that dead code should be deleted along with my deletion comment on `ICMakePropertiesController` as it all depended on `save` being called.  The `if (deleteCMakeCache) {` block in `build` is currently unreachable.

NOTE: The deletion of `CMakeCache.txt` is a good idea, but we should do it completely as a new feature. CMake gives a reasonable error for cases it does need to be deleted, but it would be nice to automate (or semi-automate) that.

jonahgraham (changes requested):

cmake/org.eclipse.cdt.cmake.core.tests/src/org/eclipse/cdt/cmake/core/CMakeBuildConfigurationTests.java
betamaxbandit, line 1: I saw that all my tests fail with the message:

"Toolchain is missing for build configuration"

So I have added a mocked toolchain.

There are other failing tests, but I'm hoping these are not related to my change. When these fixes go in and I check again.

betamaxbandit (commented):
As well as test failures, I have an outstanding thing to check. That is the default value for the generator on windows and linux is acceptable.

cmake/org.eclipse.cdt.cmake.core.tests/src/org/eclipse/cdt/cmake/core/ExtendedCMakeBuildConfiguration.java
betamaxbandit, line 1: I did not mean to commit this. I will remove when I have checked test results.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/internal/CMakePropertiesController.java
jonahgraham: CMakePropertiesEvolutionTest is to test how it evolves when serializing, it does not do anything useful without serializing AFAICT

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
```diff
@@ -124,6 +117,45 @@ public CMakeBuildConfiguration(IBuildConfiguration config, String name, IToolCha
...
+	public ICMakeProperties getCMakeProperties() {
+		// Create CMake default properties
+		ICMakeProperties properties = CMakePropertiesFactory.createProperties();
+		// Set defaults for both platforms
+		IOsOverrides windowsOverrides = properties.getWindowsOverrides();
+		windowsOverrides.setDefaultGenerator(CMakeGenerator.Ninja);
+		IOsOverrides linuxOverrides = properties.getLinuxOverrides();
+		linuxOverrides.setDefaultGenerator(CMakeGenerator.Ninja);
+		// reset(true) causes the CMake generator to be reset to it's default value.
+		properties.reset(true);
```
jonahgraham, line 130: What is the purpose of this? ICMakeProperties is a temporary object now as we have removed serialization and any chance of this being used with the UI in its current form.

cmake/org.eclipse.cdt.cmake.core.tests/src/org/eclipse/cdt/cmake/core/CMakeBuildConfigurationTests.java
```diff
@@ -0,0 +1,208 @@
...
+		List<String> linuxExtraArguments = linuxOverrides.getExtraArguments();
+		assertThat(linuxExtraArguments, contains("-DLinuxtest0=0", "-DLinuxtest1=1"));
+	}
+
+	private class CMakeBuildConfigurationExtended extends CMakeBuildConfiguration {
+
+		public CMakeBuildConfigurationExtended(IBuildConfiguration config, String name, IToolChain toolChain) {
+			super(config, name, toolChain);
+		}
+
```
jonahgraham, line 182: A key set of tests is missing still, there are no tests for when an extender of CMakeBuildConfiguration wants to provide different defaults for things like the generator and extra args.

jonahgraham (changes requested):
The TODO in the example project needs to be completed as well. I have merged the required commits.

TODO: https://github.com/eclipse-cdt/cdt/blob/5e62200e60deccb0b43f341153a4db5f5b02ccc2/cmake/org.eclipse.cdt.cmake.example/src/org/eclipse/cdt/cmake/example/ExtendedCMakeBuildConfiguration.java#L37

jonahgraham:
@[jonahgraham](https://github.com/jonahgraham) jonahgraham [force-pushed](https://github.com/eclipse-cdt/cdt/compare/6a77bf818da8a42f051f0dce1c80e0787ea2fad6..afb8c41f8685d50f82baff7293c22c666430794a) the IDE-82683-REQ-012 branch from [6a77bf8](https://github.com/eclipse-cdt/cdt/commit/6a77bf818da8a42f051f0dce1c80e0787ea2fad6) to [afb8c41](https://github.com/eclipse-cdt/cdt/commit/afb8c41f8685d50f82baff7293c22c666430794a) 
[now](https://github.com/eclipse-cdt/cdt/pull/1046#event-16046254474)

^^^ was rebase onto main.

jonahgraham:
As discussed on a call today between @betamaxbandit, @Kummallinen  and myself we have decided to remove the incomplete work related to IOsOverrides. This code was worked on to support use cases like building in docker containers and automatically switching cmake commands, but that work was never completed. For now that extra API is a lot of overhead in complication.

Therefore I have removed it as a new commit on this PR, it will be squashed later, but I have kept it as an independent commit so that @betamaxbandit can review what I have done. @betamaxbandit have a look at https://github.com/eclipse-cdt/cdt/pull/1046/commits/12180666eb74086bef0dba83b7da9963cba4776c when you get a chance.

jonahgraham:
As discussed on a call today between @betamaxbandit, @Kummallinen and myself I added some additional refactoring which I have again left as an extra commit ([link to commit in pull request](https://github.com/eclipse-cdt/cdt/pull/1046/commits/50cd03ad8416604b500f0fb7d4dbd00c80ba5e06)). 

Please see the commit message for a summary of all the changes. 

@betamaxbandit can you have a look. If this is all ok then I will squash the commits and update the description as this has grown to more than just the API change, but now includes the user visible changes that naturally flow out of the API changes.

cmake/org.eclipse.cdt.cmake.ui/src/org/eclipse/cdt/cmake/ui/internal/CMakeBuildTab.java
```diff
@@ -37,16 +39,17 @@ public class CMakeBuildTab extends CommonBuildTab {
 	 * Checkbox allowing user to choose these settings over the default operating system defaults.
```
betamaxbandit, line 39: Comment should be updated to something like:

	/**
	 * Checkbox, controls whether to use UI specified values or defaults set in
	 * {@link CMakeBuildConfiguration#getDefaultProperties()}.
	 * See {@link CMakeBuildConfiguration#CMAKE_USE_DEFAULT_CMAKE_SETTINGS}
	 */

cmake/org.eclipse.cdt.cmake.ui/src/org/eclipse/cdt/cmake/ui/internal/messages.properties
```diff
@@ -1,15 +1,14 @@
...
 CMakeBuildTab_Generator=Generator
-CMakeBuildTab_Ninja=Ninja
 CMakeBuildTab_NoneAvailable=No Toolchains Available for this Target
 CMakeBuildTab_Settings=CMake Settings
 CMakeBuildTab_Toolchain=Toolchain
-CMakeBuildTab_UnixMakefiles=Unix Makefiles
-CMakeBuildTab_useUICmakeSettings=Use these settings
-CMakeBuildTab_useUICmakeSettingsTip=Use these settings instead of the operating system defaults
+CMakeBuildTab_useDefaultCmakeSettings=Use default CMake settings
+CMakeBuildTab_useDefaultCmakeSettingsTip=Use the default CMake settings that are provided by the tool chain and C build configuration
```
betamaxbandit, line 11: Consider changing
C build configuration
to
Core Build System

betamaxbandit (commented):
LGTM
Just a few minor suggestions.
I've done some testing on windows and looks good.
I like the new UI modifications.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
```diff
@@ -69,28 +64,28 @@
...
 /**
- * @since 1.6
+ * @since 2.0
  */
-public class CMakeBuildConfiguration extends CBuildConfiguration {
+public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMakeBuildConfiguration {
 
-	public static final String CMAKE_USE_UI_OVERRIDES = "cmake.use.ui.overrides"; //$NON-NLS-1$
-	public static final boolean CMAKE_USE_UI_OVERRIDES_DEFAULT = false;
+	public static final String CMAKE_USE_DEFAULT_CMAKE_SETTINGS = "cmake.use.default.cmake.settings"; //$NON-NLS-1$
```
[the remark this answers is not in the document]
jonahgraham: I agree - but I went furhter. See latest commit that moves these constants to the ICMakeBuildConfiguration interface and I added some basic javadocs for all the properties. I also fixed up the out of date getCMakeProperties javadoc.

cmake/org.eclipse.cdt.cmake.ui/src/org/eclipse/cdt/cmake/ui/internal/CMakeBuildTab.java
jonahgraham: I decided this comment added little value and removed it. i.e. all the fields connect to their given field in ICMakeBuildConfiguration and leaving the code to document itself is better than repeating it. This is true because this is a private field, if it was API having javadoc would be more useful.

jonahgraham:
There have been many discussions about how to handle CMake preferences. I cross reference some of those key points here. AFAICT this PR now resolves all these questions in the generally agreed upon way.

From https://github.com/eclipse-cdt/cdt/pull/1046#issuecomment-2605350227

> We don't need the YAML file. We can use the preferences. The infrastructure is there.

From https://github.com/eclipse-cdt/cdt/pull/814#issuecomment-2156104747

> You may safely delete the stuff dealing with that or find a way to combine both the YAML stettings and the launchbar settings.

From https://github.com/eclipse-cdt/cdt/issues/818#issuecomment-2159165062

>> A) Will removing the YAML file have any impact on the cmake4eclipse (https://github.com/15knots/cmake4eclipse) integration in CDT?

> No, it won't (from the author of cmake4eclipse)

From https://github.com/eclipse-cdt/cdt/issues/818#issuecomment-2160257436

> I also vote for the preferences approach (option 1).

jonahgraham:
@betamaxbandit I have done what I hope is the final cleanup and squash on this work.

jonahgraham:
PS I updated the title and description in https://github.com/eclipse-cdt/cdt/pull/1046#issue-2796014939 of the PR to reflect where the PR ended up.

betamaxbandit (commented):
LGTM.
I like the move of constants into ICMakeBuildConfiguration interface

I tried your latest commit (a02340) on Windows 10 and all looks good.

jonahgraham (approved):
@betamaxbandit has reviewed the changes https://github.com/eclipse-cdt/cdt/pull/1046#pullrequestreview-2575481046

Issue #1048, opened 2025-01-21
Restrict Pin View and Open New View Actions to CPP application debug context.


raghucssit:
We want to restrict the visibility of the Pin View and Open New View toolbar Actions to be availbale only when there is a CPP application under debug.
This is because Pin View does is never enabled other than CPP application debug time. Only a disabled icon sits on the toolbar. Open New View is also associated with it because if we Pin a view and need to duplicate View to resume other things.
So both are enabled or disabled together only when there is a CPP debug context.

This includes  converting View Actions to Command Handler framework.

This is similar to https://github.com/eclipse-cdt/cdt/issues/865

PR #1049, opened 2025-01-21
Restrict Pin View and Open View to active CPP debug context.

raghucssit:
That is if any CPP application is under debug. Because Pin View is never enabled other than CPP application debug session.

This also includes Action to Command migration.

Fixes https://github.com/eclipse-cdt/cdt/issues/1048


raghucssit:
This is how fix looks like.
@
pin-view-open-view-fix

raghucssit:
Git diff is strange. I removed old Action delegates and added new Handlers. But it shows as old files are modified.

iloveeclipse:
> Git diff is strange. I removed old Action delegates and added new Handlers. But it shows as old files are modified.

Not strange, but clever. It detects a move/rename, because the remaining code similarity is high enough to the old one.

raghucssit:
Oh bad.. There is a compilation problem. I will fix it. I did not have the project which is dependent on OpenNewViewAction so did not find in ws.

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/OpenNewViewHandler.java
```diff
@@ -1,5 +1,5 @@
 /*******************************************************************************
- * Copyright (c) 2010, 2014 Texas Instruments and others
+ * Copyright (c) 2025 Advantest Europe GmbH and others.
```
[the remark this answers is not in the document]
raghucssit: Now I have fixed the copyright.

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinCommandEnablementTester.java
```diff
@@ -0,0 +1,65 @@
...
+			Set<IViewPart> pinnedViews = PinViewHandler.getPinnedViews();
+			if (pinnedViews.size() > 0 && args.length == 1) {
+				for (IViewPart viewPart : pinnedViews) {
+					WorkbenchPart vPart = (WorkbenchPart) viewPart;
+					if (args[0].equals(vPart.getPartName())) {
+						return true;
+					}
+				}
+			}
+			// Check if any CDT launch is active.
```
jonahgraham, line 52: I am concerned that this is too narrow a check. I know of at least one CDT extender that doesn't extend GdbLaunch (not open source). The [PDALaunch](https://github.com/eclipse-cdt/cdt/blob/5224c92b6770a7ecf68ec7d92476bbc23a674f01/dsf/org.eclipse.cdt.examples.dsf.pda/src/org/eclipse/cdt/examples/dsf/pda/launch/PDALaunch.java#L49) is supposed to demonstrate this, but that code has bitrotten and no longer works.  

The actions used org.eclipse.cdt.debug.internal.ui.pinclone.PinCloneUtils.isPinnable(IWorkbenchPart, ISelection) to determine if the pin should be enabled. You could see if any of the launches have an adapter of type IPinProvider, and if they do, show the button.

jonahgraham (changes requested):
This looks pretty good, but there is new, unexpected, behaviour when I am debugging both Java and C in the same workspace. The pin context + new view buttons are visible and enabled on the Java context. Prior to this change, they were visible, but only the new view was enabled. See:

image

The interesting irony, is that with pin command enabled, it is working to pin Java contexts:

image

I think you should keep enablement condition otherwise pin is enabled on illegal contexts, even within CDT:

image

I think the enablement should be based on org.eclipse.cdt.debug.internal.ui.pinclone.PinCloneUtils.isPinnable(IWorkbenchPart, ISelection)

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinCommandEnablementTester.java
raghucssit: `org.eclipse.cdt.dsf.gdb.launching.GdbLaunch` does not provide an adapter for `IPinProvider`.  I need to find a way to reuse the `org.eclipse.cdt.debug.internal.ui.pinclone.PinCloneUtils.isPinnable(IWorkbenchPart, ISelection) `. It was bit difficult to achieve that. So I use the approach currently in place.

jonahgraham: I wasn't sure by your wording in the last comment. Are you still looking into this? I would like to not break known existing extenders, I don't have access to their source now, but I think they still extend DsfLaunch, so at a minimum that should be required, along with checking if the launch is of that type, rather than exactly GdbLaunch. 

The GdbExtendedLaunch example in the CDT source shows what many many extenders do. With the change as currently written, the pin and open new button disappear for those users.

Here is a screenshot that shows the buttons missing, along with how to change the launch type:

image

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinViewHandler.java
```diff
@@ -301,19 +220,55 @@ private void updatePinContextColor(DebugContextPinProvider provider) {
...
+
+		public ViewDisposeListener(IViewPart vPart) {
+			viewPart = vPart;
+		}
+
+		@Override
+		public void widgetDisposed(DisposeEvent e) {
+			pinned.remove(viewPart);
+		}
+
```
jonahgraham, line 272: Unreferenced class?

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinViewHandler.java
```diff
@@ -25,123 +26,43 @@
...
-	private String fPinnedContextLabel = ""; //$NON-NLS-1$
+public class PinViewHandler extends AbstractHandler {
+	private DebugContextPinProvider fProvider;
 	private String fLastKnownDescription = ""; //$NON-NLS-1$
-	private IAction fAction;
+	private String fPinnedContextLabel = ""; //$NON-NLS-1$
 	private IPartListener2 fPartListener;
-	private DebugContextPinProvider fProvider;
+	private static Set<IViewPart> pinned = new HashSet<>();
+	Image image;
```
jonahgraham, line 59: Add access modifier
```suggestion
	private Image image;
```

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinCommandEnablementTester.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+public class PinCommandEnablementTester extends PropertyTester {
+
+	private static final String GDB_LAUNCH_TYPE = "org.eclipse.cdt.dsf.gdb.launching.GdbLaunch"; //$NON-NLS-1$
+	private static final String PIN_VIEW_COMMAND_PROP_TEST_NAME = "canPinViewEnabled"; //$NON-NLS-1$
+
+	@Override
+	public boolean test(Object receiver, String property, Object[] args, Object expectedValue) {
+		if (PIN_VIEW_COMMAND_PROP_TEST_NAME.equals(property)) {
+			// Check if the View was pinned, then button must be there to Unpin.
```
jonahgraham, line 42: This code doesn't quite match the comment because the test of `if (args[0].equals(vPart.getPartName()))`  only works for the original view and only works in English because the strings in the plugin.xml are not translated.

To test, open a second view on Variables and pin it. Stop the debug session and the pin disappears. But if you apply the pin to the original view it remains on closing.

raghucssit:
Migrating Action Contribution to Command/Handler is tedious.
The reason is platform maintains an `Instance` of `IViewActionDelegate` for each action contribution. i.e. an instance per view in this case.
As `Action` is always there(Not show/hide), `IViewActionDelegate` takes care of `enable/disable`. And `IViewActionDelegate` has the complete state of the `Action` like `isChecked`, `IViewPart` where it is added. Based on this controlling enable/disable has every information needed.

On the other hand, `Command/Handler`, Uses a `single instance of Handler` and `single instance of Property Tester` for all the contributions on the workbench. This make it hard to track from which Part/Command enablement request is coming. Command asks Property Tester  each time when there is selection change on the workbench. It only supplies the changed selection. From this little information Property Tester cannon decide if it can display/hide the command.
This is the reason I maintained a list of pinned views and passed the view name as an argument from the command where  is is requested.

Even if I plan to contribute a dedicated Command/Handler per view still problem cannot be solved because we can duplicate the View and the duplicated view still use the shared handler.

Command/Handler is good for reusing the common state of the Command on all the contributions. And Property Tester never knows what was the state of the Command in case of Toggle if we plan to decide show/hide based on the toggle state.(Currently that is how it is done for Pin Action. i.e If it is pinned(checked) never touch it for enable/disable).

I will evaluate how can I work around this limitation.

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinCommandEnablementTester.java
jonahgraham: In reply to the above note and https://github.com/eclipse-cdt/cdt/pull/1049#issuecomment-2609955619

Note that you can update the comment rather than the code, but they should match. However the language part is a problem regardess.

> Even if I plan to contribute a dedicated Command/Handler per view still problem cannot be solved because we can duplicate the View and the duplicated view still use the shared handler.

I would go simpler rather then more complicated. The pins could be left visible on all views if any view is still pinned?

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinCommandEnablementTester.java
```diff
@@ -0,0 +1,65 @@
...
+			Set<IViewPart> pinnedViews = PinViewHandler.getPinnedViews();
+			if (pinnedViews.size() > 0 && args.length == 1) {
+				for (IViewPart viewPart : pinnedViews) {
+					WorkbenchPart vPart = (WorkbenchPart) viewPart;
+					if (args[0].equals(vPart.getPartName())) {
+						return true;
+					}
+				}
+			}
```
jonahgraham, line 51: @raghucssit what I was thinking is this, leave command visible everywhere if any pin is in use anywhere. That removes need for argument and removes current bug where it depends on whether it is primary or secondary view that the pin is applied to.

```suggestion
			// Check if any View was pinned, then leave pin visible on all views
			if (!PinViewHandler.getPinnedViews().isEmpty()) {
				return true;
			}
```

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinCommandEnablementTester.java
jonahgraham: See https://github.com/eclipse-cdt/cdt/pull/1049/files/9a457cbd907698cbb64790b0ed2dc7929716202c#r1927439200

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinCommandEnablementTester.java
raghucssit: I am working on Non declarative approach. I have registered a `IDebugContextListener` with `Debug Service`. Now I get debug changes. Based on the change I will poll the views on the workbench contribute Action dynamically.
Hope this is not an overkill..:-)

jonahgraham: Please see https://github.com/eclipse-cdt/cdt/pull/1049#pullrequestreview-2568048108 AFAICT the pin command shows when Java process is selected with your version because there is some gdblaunch present pin commands are visible.

jonahgraham: > This will enable Pin Command on all the Views if only one View is pinned.

Yes - this is intent of my recommendation, make pin commands either visible or not visible globally rather than trying to do it per view. This is a simplification that I assume would resolve the issue. i.e. pin command is only visible if user is actually doing CDT work.

raghucssit:
@jonahgraham I think i fixed the suggestion you gave me. I actually installed `IDebugContextListener` and reusing the `PinCloneUtils.isPinnable(null, debugContext)`.
Please check my latest commit.
Here is how it works.

https://github.com/user-attachments/assets/dd8b9ded-9826-464d-b2d8-02e556991ef3

debug/org.eclipse.cdt.debug.ui/plugin.xml
```diff
@@ -1402,6 +1332,96 @@
...
+          locationURI="toolbar:org.eclipse.debug.ui.VariableView?after=additions">
+       <command
+             commandId="org.eclipse.cdt.debug.ui.PinViewCommand"
+             icon="icons/elcl16/toolbar_pinned.gif"
+             label="%PinView.name "
+             style="toggle">
+          <visibleWhen
+                checkEnabled="false">
+             <test
+                   args="Variables"
```
jonahgraham, line 1346: Now that you aren't using the arg, please remove them from the plugin.xml
```suggestion
```

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinCommandEnablementTester.java
```diff
@@ -0,0 +1,82 @@
...
+ * within Debug Launch Manager.
+ * </p>
+ *
+ * @author Raghunandana Murthappa
+ */
+public class PinCommandEnablementTester extends PropertyTester implements IDebugContextListener {
+
+	private static final String PIN_VIEW_COMMAND_PROP_TEST_NAME = "canPinViewEnabled"; //$NON-NLS-1$
+	private ISelection debugContext;
+	private boolean isDclInstalled;
```
jonahgraham, line 41: > @jonahgraham I think i fixed the suggestion you gave me. I actually installed IDebugContextListener and reusing the PinCloneUtils.isPinnable(null, debugContext).

Sadly this violates the API contract of PropertyTester. See [javadoc](https://help.eclipse.org/latest/nftopic/org.eclipse.platform.doc.isv/reference/api/org/eclipse/core/expressions/PropertyTester.html#:~:text=So%20property%20testers%20should%20always%20be%20implemented%20in%20a%20stateless%20fashion.) key part quoted:

>  So property testers should always be implemented in a stateless fashion.

If you are listening to the active debug context, then I think lots of commands do that already. Have you had a look at how other debug commands are implemented? I am quite short of time now, but I can find you a specific example later if you want.

jonahgraham (changes requested):

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinCommandEnablementTester.java
jonahgraham: > find you a specific example later if you want.

For example, look at how `debugContext` variable is used in org.eclipse.debug.examples.ui ([source link](https://github.com/eclipse-platform/eclipse.platform/blob/e44cc1b8be5011ce3e8a459d0b41777ac6333869/debug/org.eclipse.debug.examples.ui/plugin.xml#L385-L406))

CDT also has reverse commands (E.g. `org.eclipse.cdt.debug.ui.command.reverseStepInto`) and the property tester `ReverseDebuggingPropertyTester` and expression definition `org.eclipse.cdt.debug.ui.testIsReverseDebuggingEnabled` which uses `debugContext` variable. 

The CDT case is interesting, but more complicated because it is a retargettable command, whilst the o.e.debug.examples.ui is simpler.

jonahgraham (changes requested):
This looks good, the behaviour seems all correct.

I have marked resolved all the comments that are complete so it should be easier to see remaining items

> If things are good then i will squash and push all the commits. I have made different commits to differentiate the changes to help review.

I appreciate the different commits, it made it much easier to review the updates.

In addition, there is a new error introduced because there is another plugin.xml that still needs updating. Please look at https://github.com/eclipse-cdt/cdt/blob/f70a8b35c18a21a8cfb9132ad9de84caebac0d78/memory/org.eclipse.cdt.debug.ui.memory.memorybrowser/plugin.xml#L43-L65

dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/debugview/layout/PinCommandEnablementTester.java
```diff
@@ -0,0 +1,59 @@
...
+			IWorkbenchWindow window = PlatformUI.getWorkbench().getActiveWorkbenchWindow();
+			if (window != null) {
+				IWorkbenchPage page = window.getActivePage();
+				if (page != null) {
+					IViewPart debugPart = page.findView(IDebugUIConstants.ID_DEBUG_VIEW);
+					if (debugPart != null) {
+						ISelection debugContext = DebugUITools.getDebugContextForPart(debugPart);
+						if (debugContext instanceof IStructuredSelection structSel) {
+							Object firstEle = structSel.getFirstElement();
+							if (firstEle instanceof IDMVMContext) {
```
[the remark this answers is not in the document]
raghucssit: > In addition, I am guessing that you can't use `debugContext` as mentioned here [#1049 (comment)](https://github.com/eclipse-cdt/cdt/pull/1049#discussion_r1927990059) Did you try and it didn't work, or was there some limitation? I don't mind the answer, I would like to learn about any possible limitations you have found with this solution for my future work.

Yes. We can't use `debugContext` variable.
This is because `org.eclipse.core.expressions.WithExpression#evaluate(IEvaluationContext context)` checks first if `debugContext` active, If there no debugContext active then it will simply disable the command and don't ask the property tester. So if we have pinned the view and then if all the launch configs are terminated it removes the contribution. But it should be there because we need to unpin it.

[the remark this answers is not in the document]
raghucssit: Thanks a lot for this code suggestion. This is want I was looking for. But did not find. This is more trust able as already tested and working.

raghucssit:
@jonahgraham I have fixed all the comments and tested. It works fine. Thanks for all the suggestions and guidelines.
Build failed due to some repository is not reachable. May be infrastructure isssue.

jonahgraham:
> Build failed due to some repository is not reachable. May be infrastructure isssue.

Yes, I'll keep an eye on it and kick it off when services are back from their maintenance downtime https://www.eclipsestatus.io/

dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/debugview/layout/PinCommandEnablementTester.java
```diff
@@ -0,0 +1,58 @@
...
+		if (PIN_VIEW_COMMAND_PROP_TEST_NAME.equals(property)) {
+
+			// We enable Pin Command on all views if it is pinned in any View.
+			if (!PinViewHandler.getPinnedViews().isEmpty()) {
+				return true;
+			}
+
+			IAdaptable debugContext = DebugUITools.getDebugContext();
+			IDMVMContext dmvmContext = Adapters.adapt(debugContext, IDMVMContext.class);
+			return DsfSession.isSessionActive(dmvmContext.getDMContext().getSessionId());
```
jonahgraham, line 54: Possible NPE here. I will apply this change if I am ready to merge the PR before you get a chance to action this comment.

```suggestion
			return dmvmContext != null && DsfSession.isSessionActive(dmvmContext.getDMContext().getSessionId());
```

dsf/org.eclipse.cdt.dsf.ui/src/org/eclipse/cdt/dsf/debug/internal/ui/debugview/layout/PinCommandEnablementTester.java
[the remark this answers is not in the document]
jonahgraham: >  it will simply disable the command and don't ask the property tester.

:-) Thanks for that insight.

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinViewHandler.java
jonahgraham: @raghucssit you marked this as resolved, but the class still seems to be present and unreferenced. To be clear I am referring to the package private inner class `ViewDisposeListener`. Is there some work or use case outstanding here that this class represents?

raghucssit: Oh, I missed that. We don't need this. I was planning to use `ViewDisposeListener` instead of `IPartListener2`. But it did not fulfill the needed functionality. I forgot to remove this. Now it is removed.

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/internal/ui/actions/PinViewHandler.java
```diff
@@ -301,19 +218,41 @@ private void updatePinContextColor(DebugContextPinProvider provider) {
...
+				overlayColor = descOverlayColor;
 			}
 		}
+
+		return false;
+	}
+
+	@Override
+	public void dispose() {
+		pinned = null;
```
jonahgraham: Make `pinned` final and `clear()` it instead of setting to null as the setting to null has a bunch of possible NPEs it causes if this handler gets disposed.

I feel I am probably overthinking this, but since in theory a new instance of the handler could be instantiated, the static `pinned` should remain usable.

I have applied this change in the push I did just now.

jonahgraham (approved):
Thanks @raghucssit 

I have rebased this change, updated the commit comment to say "Fixes" to automatically close #1048 when we merge and applied the final trivial fix review comment I just created. 

Once build passes this will be merged.

iloveeclipse:
@jonahgraham : many thanks for a very helpful review.

PR #1051, opened 2025-01-22
Allow the new API for CMakeBuildConfiguration to be extended

jonahgraham:
PR #1010 added the ability to extend CMakeBuildConfiguration and CMakeBuildConfigurationProvider by making the classes public API, they used to be internal API.

This change makes it easier to reuse the code in the provider and configuration by allowing extenders to provide their own implementations of CMakeBuildConfiguration. This has been achieved by adding createCMakeBuildConfiguration methods to control which CMakeBuildConfiguration is constructed.

Follow up to #1010


jonahgraham:
@betamaxbandit please have a look at this and let me know your thoughts.

PR #1052, opened 2025-01-22
Provide an example of extending CMake project type in Core Build

jonahgraham:
In issue #1000 a lot of the work is on exposing a better API to ISVs.
This new project serves these purposes:

- Ensure the exposed API works in practice
- Example for ISVs to use if they want to extend CMake
- Provide an in-tree example of the API in use, so future editors of CDT have better chance of understanding *why* items are done as they are.


jonahgraham:
@betamaxbandit please have a look at this and let me know your thoughts. 

In particular see the TODO in new class ExtendedCMakeBuildConfiguration where you can demonstrate the new API being added in #1046

cmake/org.eclipse.cdt.cmake.example/plugin.xml
```diff
@@ -0,0 +1,55 @@
...
+            id="org.eclipse.cdt.cmake.example.tag"
+            label="CMake Examples for ISVs">
+      </tag>
+      <template
+            icon="icons/cmake_logo-main.png"
+            id="org.eclipse.cdt.cmake.ui.example.newExtendedProjectTemplate"
+            label="Extended CMake Project (ISV Example)"
+            wizard="org.eclipse.cdt.cmake.example.NewExtendedCMakeProjectWizard">
+         <description>
+            A CMake project the demonstrates how to extend CDT's CMake project type. This project is an example and not shipped with CDT and is intended for ISVs as a demonstration.
```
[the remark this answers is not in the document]
jonahgraham: Thanks. Fixed in latest commit.

cmake/org.eclipse.cdt.cmake.example/src/org/eclipse/cdt/cmake/example/ExtendedCMakeBuildConfiguration.java
```diff
@@ -0,0 +1,38 @@
...
+			ICMakeToolChainFile toolChainFile, String launchMode) {
+		super(config, name, toolChain, toolChainFile, launchMode);
+	}
+
+	public ExtendedCMakeBuildConfiguration(IBuildConfiguration config, String name, IToolChain toolChain) {
+		super(config, name, toolChain);
+	}
+
+	// TODO: Here goes the example extending that is being developed in #1046 https://github.com/eclipse-cdt/cdt/pull/1046
+}
```
betamaxbandit, line 38: Hi @jonahgraham ,
not sure if you were expecting me to add some stuff here. Anyway, I think something like this:

	@Override
	public ICMakeProperties getCMakeProperties() {
		ICMakeProperties properties = super.getCMakeProperties();

		/*
		 * To change the CMake generator (specified with -G <generator>) used on both Windows and Linux host platforms.
		 */
		IOsOverrides windowsOverrides = properties.getWindowsOverrides();
		windowsOverrides.setGenerator(CMakeGenerator.Ninja);
		IOsOverrides linuxOverrides = properties.getLinuxOverrides();
		linuxOverrides.setGenerator(CMakeGenerator.UnixMakefiles);
		return properties;
	}


BTW, I tried using properties.setWarnUnitialized(true), but it causes a "CMake Error: Unknown argument --warn-unitialized" because it has a typo in it. I guess we want to fix that in a separate PR.

jonahgraham: Yes, I am expecting you to add stuff here when you work on #1046 so that your commit in #1046 shows the new API and how to use it in the same Pull Request.

> BTW, I tried using properties.setWarnUnitialized(true), but it causes a "CMake Error: Unknown argument --warn-unitialized" because it has a typo in it. I guess we want to fix that in a separate PR.

Doh!

PR #1053, opened 2025-01-22
Check that all xml files in CDT are well formed

jonahgraham:


jonahgraham:
Created to ensure we don't have a regression on the error that #1054 fixes. This first commit is expected to fail, and once #1054 is merged I will rebase and this should pass.

PR #1054, opened 2025-01-22
Fix XML syntax error in freemarker manifest

jonahgraham:
The extra " inside XML property was actually being processed properly by freemarker (surprisingly!), but XML editors/viewers reported an error.


jonahgraham:
I want to make sure the new check added in https://github.com/eclipse-cdt/cdt/pull/1053 fails before I merge this fix.

PR #1060, opened 2025-01-28
Add some documentation and consistency to use of info vs output stream

### LINKED ISSUE #1059: CMake Clean has inverted colors compared to build
As you can see from these screenshots that blue and black parts of the output seem inverted in the build vs clean

Build:

Image

vs Clean:

Image

 _Originally posted by @jonahgraham in [#691](https://github.com/eclipse-cdt/cdt/issues/691#issuecomment-2619572330)_

jonahgraham:
With some better Javadocs the cross referencing between IConsole and the UI aspects of the console is a little easier to follow.

This resolves #1059 in two parts:

1. For Core Build System this update consistently uses info stream to show information messages and output stream to be stdout of launched build tool. This resolves the "Build Complete" appearing as the output color when doing clean (See screenshots in #1059)

2. CBuildConfiguration.watchProcess(IConsole, IProgressMonitor) incorrectly passed the info stream as the output stream. Mostly this was used for the clean stage of builds. This resolves the CMake output like ("Cleaning all built files...") appearing as the info color when doing clean (See screenshots in #1059)

Fixes #1059

Here are some after screenshots, see #1059 for the before screenshots:

image

image

### DIFF
--- a/build/org.eclipse.cdt.core.autotools.core/src/org/eclipse/cdt/core/autotools/core/AutotoolsBuildConfiguration.java
+++ b/build/org.eclipse.cdt.core.autotools.core/src/org/eclipse/cdt/core/autotools/core/AutotoolsBuildConfiguration.java
@@ -90,7 +90,7 @@ public class AutotoolsBuildConfiguration extends CBuildConfiguration {
 		try {
 			project.deleteMarkers(ICModelMarker.C_MODEL_PROBLEM_MARKER, false, IResource.DEPTH_INFINITE);
 
-			ConsoleOutputStream outStream = console.getOutputStream();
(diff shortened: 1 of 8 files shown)


jonahgraham:
> This pull request removes 1217 tests.

Hmm. A side of effect of making this comment work is that it compares to the last build of HEAD. However since we run different numbers of tests depending on whether debug tests have been suppressed we get this unexpected warning about removing tests. The tests were not removed, they were just not run this time.

betamaxbandit (commented):
LGTM
Neat. Good find and +1 for consistency and with the extra javadoc explanation!

PR #1062, opened 2025-01-28
Fix copy and paste error in CMake manual test instructions

jonahgraham:
Both 5 and 6 were titled the same, but 6 sub steps are correctly for the clean target

Fixup for #1046


jonahgraham:
FYI @betamaxbandit

betamaxbandit:
Thanks @jonahgraham

PR #1063, opened 2025-01-29
Bump org.eclipse.cdt.core/ui bundle version to 9.0.0

jonahgraham:
Some of the changes that are about to arrive have specific impacts of API of core build. Because CBS is largely in org.eclipse.cdt.core/ui any breaking changes to the CBS API necessitate bumping the entire bundle to a new version.

One of the most noticeable changes is expected to be that org.eclipse.cdt.core.build.ICBuildConfigurationProvider.createBuildConfiguration(IProject, IToolChain, String, IProgressMonitor) needs a new parameter to properly connect the target. The upcoming PR that makes the API change will document that change in the CHANGELOG-API.md and N&N documentation.

Part of #1000


jonahgraham:
cc: @betamaxbandit

jonahgraham:
I would like to merge and build this version bump today so that we can contribute an update 12.0 M2 to SimRel so all consumers of CDT API see the version bump as soon as possible. Deadline for this is today.

If anyone has thoughts or concerns, kindly let me know.

betamaxbandit (commented):
LGTM

PR #1066, opened 2025-01-29
Add CDT-LSP to API Baseline

jonahgraham:
See https://github.com/eclipse-cdt/cdt-lsp/pull/406#issuecomment-2622119241


jonahgraham:
I have tested this both using the manual method (https://github.com/eclipse-cdt/cdt/blob/main/POLICY.md#Using-API-Tooling) and the Oomph install method (https://github.com/eclipse-cdt/cdt/blob/main/CONTRIBUTING.md#setup-cdt-for-development-with-oomph) and it seems ready to go to me - I will merge once build completes.

ghentschke:
@jonahgraham Thank you, now the API baseline check works for CDT-LSP repo as well!

PR #1069, opened 2025-02-02
Clarify API around CBuildConfiguration.getBuildDirectory as @nooverride

jonahgraham:


jonahgraham:
@ewaterlander and @betamaxbandit can you review please?

betamaxbandit (approved):
LGTM
However, there is a task (#1000 - IDE-82683-REQ-018) to make the build output folder configurable. At the moment I don't know what this API should look like, but it might impinge on this. Will only know when I get to that task.

jonahgraham:
> LGTM However, there is a task (#1000 - IDE-82683-REQ-018) to make the build output folder configurable. At the moment I don't know what this API should look like, but it might impinge on this. Will only know when I get to that task.

OK - we can make it overridable when you come to that. There are three methods getBuildContainer, getBuildDirectoryURI and getBuildDirectory and when extenders override them it becomes unclear which one should be. By the time you come to that https://github.com/eclipse-cdt/cdt/pull/1071 will probably be merged. If you think that API use is a problem, please let me know as it isn't set in stone yet.

ewaterlander:
> @ewaterlander and @betamaxbandit can you review please?

LGTM.
We only use getBuildDirectoryURI()

PR #1070, opened 2025-02-02
Use CommandLineUtil to split and join command lines in CBS Makefile support

jonahgraham:
Without this the build and clean command entered in the CBS settings could be parsed and re-assembled incorrectly.

A simple example is that an extra space (e.g. "make  clean") would lead to an error when running make like:

```
make: *** empty string invalid as file name.  Stop.
```

This happened because the code used trivial split on " " and join with " " to handle parsing that command line string into array of arguments. This change fixes that by using the functionality already available in CommandLineUtil

TODO:

- [ ] Add missing tests for joining array back into string
- [ ] Implement optional quoting of arguments on argumentsToStringWindowsCreateProcesswind
- [ ] Add quotes around empty parameters for argumentsToString*


jonahgraham:
This was a first attempt at fixing #1072. But it splits and re-joins between the UI and the preference store. This means that users don't get out what they typed in.

Instead I am going to update the API to store in the preference store what users have typed, and just split when create process.

PR #1071, opened 2025-02-02
Add a directory tracker to the CBS Makefile builder

jonahgraham:
Messages such as "make: Entering directory" are now processed so that the scanner info builder knows the correct full path to use for build.

Without this code calls to ToolChain.getResourcesFromCommand() would pass in the wrong directory and resources could not be reliably calculated. With the incorrect calculation the scanner info does not get applied to the file correctly and the file cannot be indexed reliably.

TODO:

- [ ] Decide how to handle make -j as "make: Entering directory" becomes unreliable
- [ ] Consider whether we should search project for resources so that the exact build directory is less important
- [ ] Consider how to report back to the user that the detected resource was not found in the project.

### DIFF
--- a/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/StandardBuildConfiguration.java
+++ b/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/StandardBuildConfiguration.java
@@ -11,12 +11,16 @@
 package org.eclipse.cdt.core.build;
 
 import java.io.IOException;
+import java.net.URI;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Stack;
+import java.util.regex.Matcher;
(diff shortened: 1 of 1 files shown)


jonahgraham:
> TODO:
> 
> * [ ]  Decide how to handle make -j as "make: Entering directory" becomes unreliable
> * [ ]  Consider whether we should search project for resources so that the exact build directory is less important
> * [ ]  Consider how to report back to the user that the detected resource was not found in the project.

I think these TODO items can be split into future issues. When reviewing please let me know if you think they need to be dealt with.

jonahgraham:
@ewaterlander can you have a look at this? Do you have the issues of changing directories in the middle of the build?

BTW I have a handful of patches to Makefile flow in CBS as I am trying to [dogfood](https://en.wikipedia.org/wiki/Eating_your_own_dog_food) while working on https://github.com/eclipse-equinox/equinox/issues/831#issuecomment-2629052476

ewaterlander:
Hi @jonahgraham 

We typically don't do directory changes in our builds, so I did not see this issue.
This change LGTM.

PR #1073, opened 2025-02-02
Use CommandLineUtil to split and join command lines in CBS Makefile support

### LINKED ISSUE #1072: CBS Makefile builder does not parse build / clean commands fully
Entering anything other than a simple string in CBS Makefile builder's build / clean commands don't parse correctly and are not passed to make as expected.

I am referring to these fields:

Image

For example, enter `make  all` with an extra space leads to a make error:

Image

This is happening because the builder is simply splitting on `" "` which leads to an extra empty command line option

https://github.com/eclipse-cdt/cdt/blob/84d99f277ab1e48da8e1c8ce3e6c15d1e8b260ad/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/StandardBuildConfiguration.java#L102

This also means other, more complicated cases, such as quoted strings are not handled properly either:

Image

leads to:

Image

jonahgraham:
Without this the build and clean command entered in the CBS settings could be parsed and re-assembled incorrectly.

A simple example is that an extra space (e.g. "make  clean") would lead to an error when running make like:

```
make: *** empty string invalid as file name.  Stop.
```

This happened because the code used trivial split on " " and join with " " to handle parsing that command line string into array of arguments. This change fixes that by using the functionality already available in CommandLineUtil

Fixes #1072

For example, when building with a quoted argument with a space it works now, and it also quotes the output correctly in the Console so it can be copied and pasted if needed:

image

Note that `make`, and other arguments that don't need quoting are always output quoted. A separate PR will address that.

### DIFF
--- a/NewAndNoteworthy/CHANGELOG-API.md
+++ b/NewAndNoteworthy/CHANGELOG-API.md
@@ -24,6 +24,11 @@ The following classes have been removed or modified in API breaking ways:
   - spelling corrected for methods with Uninitialized in the name
   - setWarnUnused renamed to setWarnUnusedVars and isWarnUnused renamed to isWarnUnusedVars
 
+### StandardBuildConfiguration.setBuildCommand(String[]) and StandardBuildConfiguration.setCleanCommand(String[]) removed
+
+These methods (in `org.eclipse.cdt.core.build.StandardBuildConfiguration`) made it difficult to save and load users build and clean command without modifying it.
+They have been replaced with methods that take only a `String` for consistent parsing of command lines.
+See [#1072](https://github.com/eclipse-cdt/cdt/issues/1072) for more details on motivation for this change.
 
 ## API Changes in CDT 11.5.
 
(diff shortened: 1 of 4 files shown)


NewAndNoteworthy/CHANGELOG-API.md
```diff
@@ -24,6 +24,11 @@ The following classes have been removed or modified in API breaking ways:
   - spelling corrected for methods with Uninitialized in the name
   - setWarnUnused renamed to setWarnUnusedVars and isWarnUnused renamed to isWarnUnusedVars
 
+### StandardBuildConfiguration.setBuildCommand(String[]) and StandardBuildConfiguration.setCleanCommand(String[]) removed
```
jonahgraham, line 27: @ewaterlander I have gone back and forth a little on straight out removing this API in favour of my new code. Please let me know if you are depending on the array version of setBuildCommand and setCleanCommand.

jonahgraham:
> Note that `make`, and other arguments that don't need quoting are always output quoted. A separate PR will address that.

See #1074

NewAndNoteworthy/CHANGELOG-API.md
ewaterlander: Hi Jonah,

I'm not depending on the array versions of these methods.

PR #1074, opened 2025-02-02
Improve CommandLineUtil to string methods so they look prettier

jonahgraham:
Initially CommandLineUtil.argumentsToString was used to provide properly quoted string when interacting with GDB. But it is also useful to print a string that can be copied + pasted to a terminal for the user.

When doing this the always quote every argument looks less nice, so this change updates the code to only quote argument if needed.

Tests have also been added for the quoting.

Improves look and feel of changes in #1073


jonahgraham:
This change removes the extra `'` around make and other arguments that don't need it. 

Before (with #1073):

image

After (with #1073 and #1074):

image

Issue #1058, opened 2025-01-28
Build Error from compile commands generator on MBS projects with build location outside workspace


Kummallinen:
**Describe the bug**
When building an MBS project that has the build location set to outside of the workspace the build errors with an NPE in CompilationDatabaseGenerator.

```
!ENTRY org.eclipse.core.resources 4 2 2025-01-20 18:31:28.848
!MESSAGE Problems occurred when invoking code from plug-in: "org.eclipse.core.resources".
!STACK 0
java.lang.NullPointerException: Cannot invoke "org.eclipse.core.runtime.IPath.append(String)" because "buildDirectory" is null
	at org.eclipse.cdt.managedbuilder.internal.core.jsoncdb.generator.CompilationDatabaseGenerator.generate(CompilationDatabaseGenerator.java:93)
	at org.eclipse.cdt.managedbuilder.internal.core.CommonBuilder.build(CommonBuilder.java:515)
	at org.eclipse.cdt.managedbuilder.internal.core.CommonBuilder.build(CommonBuilder.java:471)
	at org.eclipse.core.internal.events.BuildManager$2.run(BuildManager.java:1077)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:296)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:352)
	at org.eclipse.core.internal.events.BuildManager$1.run(BuildManager.java:441)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:47)
	at org.eclipse.core.internal.events.BuildManager.basicBuild(BuildManager.java:444)
	at org.eclipse.core.internal.events.BuildManager.basicBuildLoop(BuildManager.java:555)
	at org.eclipse.core.internal.events.BuildManager.basicBuildLoop(BuildManager.java:503)
	at org.eclipse.core.internal.events.BuildManager.build(BuildManager.java:585)
	at org.eclipse.core.internal.resources.Workspace.buildInternal(Workspace.java:595)
	at org.eclipse.core.internal.resources.Workspace.build(Workspace.java:491)
	at org.eclipse.ui.actions.BuildAction$1.runInWorkspace(BuildAction.java:288)
	at org.eclipse.core.internal.resources.InternalWorkspaceJob.run(InternalWorkspaceJob.java:43)
[...]

PR #1075, opened 2025-02-03
Issue #1058: Moved generation of the compile_commands.json

alicetrifu:
Moved compile_commands.json file to performPrebuildGeneration, so the file to be generated only when Makefiles generation is set to be on.

### DIFF
--- a/build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/CommonBuilder.java
+++ b/build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/CommonBuilder.java
@@ -510,10 +510,6 @@ public class CommonBuilder extends ACBuilder implements IIncrementalProjectBuild
 			}
 
 			for (int i = 0; i < num; i++) {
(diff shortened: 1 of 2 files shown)


build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/CommonBuilder.java
```diff
@@ -509,11 +509,7 @@ private IProject[] build(int kind, IProject project, IBuilder[] builders, boolea
 				}
 			}
 
-			for (int i = 0; i < num; i++) {
-				if (isGenerateFileOptionEnabled()) {
-					CompilationDatabaseGenerator generator = new CompilationDatabaseGenerator(getProject(), activeCfg);
-					generator.generate();
-				}
+			for (int i = 0; i < num; i++) {
```
jonahgraham, line 512: The code cleanliness checks are going to fail because of trailing whitespace here
```suggestion
			for (int i = 0; i < num; i++) {
```

jonahgraham (commented):
This change LGTM - needs to pass the build, but then it is ready to go.

PR #1081, opened 2025-02-12
Fix mismatched id/name for CBuildConfigProvider

jonahgraham:
This seems to only cause issues when using the plugin.xml GUI editor because it cannot find the exsd file due to the incorrect id.


jonahgraham:
@betamaxbandit I think this plus https://github.com/eclipse-pde/eclipse.pde/issues/1608 and https://github.com/eclipse-pde/eclipse.pde/issues/1611 address the issues you raised today about the plugin.xml editor not working well.

betamaxbandit:
@jonahgraham , cool. Thanks.
I look forward to getting the fix.

In my oomph installed "Eclipse IDE for Eclipse Committers (includes Incubating components) Version: 2024-12 (4.34.0) Build id: 20241128-0757" if I do "Check for updates", will I get the #1609 change?

jonahgraham:
No, you won't get my fix in PDE until you update to latest I-build for 4.35, or wait until 2025-03 release.

PR #1076, opened 2025-02-04
The active Launch Target is now saved in the ICBuildConfiguration

betamaxbandit:
In addition to the active launch mode and toolchain, the active Launch Target is now used to determine the Core Build Configuration (ICBuildConfiguration).

The build output folder name now ends with the name of the Launch Target (eg: cmake.debug.win32.x86_64.Local)

Addresses Issue: CDT CMake Improvements #1000, IDE-82683-REQ-016 Launch Targets used in ICBuildConfiguration

Issue #1065, opened 2025-01-29
resume[gdb] (org.eclipse.cdt.tests.dsf.gdb.tests.MIRunControlTest) is a flaky test


jonahgraham:
Since fixing #816 we don't have a bunch of DSF-GDB tests consistently failing, but the update to newer Ubuntu version seems to make resume[gdb] (org.eclipse.cdt.tests.dsf.gdb.tests.MIRunControlTest) flaky. 

I cannot reproduce this failure on my local machine running same version of Ubuntu and tools, nor can I see any negative behavior in the IDE around resume. Therefore I have classed this a flaky test rather than a bug in CDT itself.

PR #1084, opened 2025-02-13
Stops CMake build output folders being named "default"

betamaxbandit:
The default IBuildConfiguration is no longer used by projects that use ICBuildConfigurationProvider.

For CMake, Makefile and other projects the build output folder is sometimes named "default" rather than the pattern
toolName.launchMode.toolchain OS.toolchain Arch.launchTarget Id (eg: cmake.debug.win32.x86_64.Local). PR #1076 exposes new API (ICBuildConfigurationProvider.getCBuildConfigName) to encourage this naming pattern.

The "sometimes" is variable and often happens when a project is first created when the active launch target is Local and the launch mode is "run", but not always. This gives a random, inconsistent impression to CDT.

The Platform project always contains a IBuildConfiguration with the name IBuildConfiguration.DEFAULT_CONFIG_NAME. It seems the original Core Build system design went to some length to fit in with this and always make use of this IBuildConfiguration when pairing it with a new ICBuildConfiguration.

With this PR, this no longer happens, allowing CDT code to be simplified and the build folder naming made consistent, always adhering to ICBuildConfigurationProvider.getCBuildConfigName.

Addresses Issue: CDT CMake Improvements #1000, IDE-82683-REQ-024 Default CMake build folder


betamaxbandit:
@jonahgraham ,
I have pushed a commit with failing junit tests for this.
The test is still WIP and it does pull in a number of dependencies into a single test. Let me know if that is a problem.

The fixes are not yet in place, but I have marked where I think changes should happen.
I'll do further testing on this before pushing the fixed PR.

There are a number of places in CDT code where the following line appears:

  names.add(IBuildConfiguration.DEFAULT_CONFIG_NAME);

I need to investigate whether these can be removed too.

jonahgraham:
> This pull request removes 1217 

Please disregard this part of the Test Results method in https://github.com/eclipse-cdt/cdt/pull/1084#issuecomment-2655190440 - the removed tests are simply not run on this PR to make build run faster.  https://github.com/eclipse-cdt/cdt/issues/1002 is umbrella for adding better auto-comments to PRs.

core/org.eclipse.cdt.core.tests/model/org/eclipse/cdt/core/build/DefaultBuildConfigurationNameTests.java
```diff
@@ -0,0 +1,152 @@
...
+
+/**
+ * Tests that a ICBuildConfiguration is never created with the default name.
+ *
+ *
+ * Tests org.eclipse.cdt.cmake.core.CMakeBuildConfigurationProvider and friends.
+ * org.eclipse.cdt.core.build.ICBuildConfigurationProvider.getCBuildConfiguration(IBuildConfiguration, String)
+ * TBC
+ */
+public class DefaultBuildConfigurationNameTests extends BaseTestCase5 {
```
jonahgraham, line 44: :+1: I'm loving the TDD

jonahgraham (commented):
In case you have the cdt-lsp projects closed, make sure to take into account how it uses ICBuildConfigurationManager.getBuildConfiguration. For example [here](https://github.com/eclipse-cdt/cdt-lsp/blob/eeb7fd76bee5e4074d7c05c287b58031e2ae7d91/bundles/org.eclipse.cdt.lsp.clangd/src/org/eclipse/cdt/lsp/clangd/internal/config/ClangdConfigurationFileManager.java#L134-L145)

betamaxbandit:
Hi @jonahgraham ,
I've updated my draft PR with that actual implementation for this ticket.
I still have some tidy up to do in the JUnit tests.

betamaxbandit:
@jonahgraham ,
my most recent PR fixes the problem I was having in the JUnits where the initial conditions for each test were not being reset. I've solved this now by adding a resetCoreBuildLaunchBarTracker(). This allows my tests to behave now as if each one were running in a new workspace and the resulting active build configuration is set as expected.

I'm very pleased with how this work has gone because I believe quarantining the default build config to the noConfigs bin is totally the correct thing to do. It stops the "default" name appearing in the build output directory sometimes. More importantly I think the correct build config for the active launchbar controls is being set straight away now, without relying on a build action first.

I have a little bit of tidying to do to the junits, but I think functionally they are complete.

jonahgraham:
The failing test is known flaky - https://github.com/eclipse-cdt/cdt/issues/1065 - I have restarted the build anyway.

core/org.eclipse.cdt.core.tests/model/org/eclipse/cdt/core/build/DefaultBuildConfigurationNameTestsNoToolchains.java
```diff
@@ -0,0 +1,177 @@
...
+		description.setNatureIds(
+				new String[] { CProjectNature.C_NATURE_ID, CCProjectNature.CC_NATURE_ID, CMakeNature.ID });
+		project.setDescription(description, null);
+		return project;
+	}
+
+	private void removeDiscoveredToolchains() throws CoreException {
+		Collection<IToolChain> allToolChains = toolchainManager.getAllToolChains();
+		for (IToolChain toolchain : new ArrayList<>(allToolChains)) {
+			System.out.println(String.format("Removing toolchain '%s' for test '%s'", toolchain.getName(), getName()));
```
jonahgraham, line 171: Remove or comment out please
```suggestion
```

core/org.eclipse.cdt.core.tests/model/org/eclipse/cdt/core/build/DefaultBuildConfigurationNameTests.java
```diff
@@ -0,0 +1,438 @@
...
+				for (Map.Entry<String, String> property : properties.entrySet()) {
+					String tcValue = mockToolchain.getProperty(property.getKey());
+					// If toolchain doesn't have this property, it doesn't care
+					if (tcValue != null && !property.getValue().equals(tcValue)) {
+						return false;
+					}
+				}
+				return true;
+			}
+		});
```
jonahgraham, line 419: This is the trick:

```suggestion
		when(mockToolchain.matches(anyMap())).thenCallRealMethod();
```

core/org.eclipse.cdt.core.tests/model/org/eclipse/cdt/core/build/DefaultBuildConfigurationNameTests.java
```diff
@@ -0,0 +1,438 @@
...
+		when(mockToolchain.getTypeId()).thenReturn("tc_typeId");
+		when(mockToolchain.getId()).thenReturn("tcId");
+		// I don't understand why it is necessary to mock IToolChain.matches ???
+		when(mockToolchain.matches(anyMap())).thenAnswer(new Answer<>() {
+			@Override
+			public Object answer(InvocationOnMock invocation) throws Throwable {
+				Map<String, String> properties = invocation.getArgument(0);
+				// Copied from IToolChain.matches(Map<String, String>)
+				/*
+				 * Is this logic correct? With values
```
jonahgraham, line 405: I don't know the answer - the IToolChain version has this comment "// If toolchain doesn't have this property, it doesn't care" - so it seems on purpose

It was written this way because of https://bugs.eclipse.org/bugs/show_bug.cgi?id=548980

jonahgraham (commented):
This LGTM - set to ready to review when you are satisfied.

jonahgraham:
This looks good to me - it exposes a pre-existing bug in cdt-lsp I have raised here https://github.com/eclipse-cdt/cdt-lsp/issues/435

Issue #968, opened 2024-12-13
Make CDT LSP editor more visible and easier to onboard in CDT 12


jonahgraham:
The goal is for CDT users to have a way to easily try out the CDT-LSP feature and start to transition their workflows to the long term CDT implementation.

Make CDT LSP visible in the following ways:

1. Add to the top of the CEditor a button(s) that says "Try LSP"
2. Add to the top of the LSP Editor a button(s) that says "go back to normal CEditor"
3. Add "Quick-fix" on CEditor errors to say "Try LSP" - See https://github.com/eclipse-cdt/cdt-lsp/issues/379

More details on 1 and 2:

Contribute a banner to the CEditor and CLspEditor to promote the new C/C++ editing experience based on LSP. The banner in CEditor enables a quick switch to the LSP editor:

![image](https://github.com/user-attachments/assets/777c0f8f-1d04-4a90-ab9c-f4e69a7fa929)

And in the CLspEditor it enables a quick switch back to the traditional CEditor.

![image](https://github.com/user-attachments/assets/91b58306-a1dc-470f-8362-607876ead08c)

If there is no project specific setting for `prefer_lsp` a quick confirmation dialog is displayed with options for help and feedback:

![image](https://github.com/user-attachments/assets/7c33d4fe-69a8-4cf8-b1fb-7ca80e07020c)

and the other way:

![image](https://github.com/user-attachments/assets/a69a87d8-4f32-4e58-a651-2583db50183e)

If the editor is open on a specific project and that project has project specific preferences for `prefer_lsp`, instead of automatically changing the setting, the property page is opened:

![image](https://github.com/user-attachments/assets/16c2ce3b-5b62-4199-8392-5d6800f621c7)

A notification  (typically in the bottom right of the IDE) is also displayed to the user offering more information and access to the preferences.

![image](https://github.com/user-attachments/assets/e8d069de-0939-4503-8755-a130595570d0)

In dark mode the banner looks like this:

![Image](https://github.com/user-attachments/assets/73f669b9-8f69-42f5-a600-3951826ef120)

And the notification (typically in the bottom right of the IDE) looks like this:

![Image](https://github.com/user-attachments/assets/5cc417bb-35c3-49fe-a32b-e12efbbc5384)

jonahgraham:
Updated original issue comment to read:

The goal is for CDT users to have a way to easily try out the CDT-LSP feature and start to transition their workflows to the long term CDT implementation.

[...]

PR #1088, opened 2025-02-16
Add internal API to CEditor to allow CDT-LSP to contribute a "try it" banner

jonahgraham:
Part of https://github.com/eclipse-cdt/cdt/issues/968


jonahgraham:
I would like to merge this ASAP so that https://github.com/eclipse-cdt/cdt/pull/1088 can build successfully and we can get this into M3.

I have tried to keep this code as simple as possible and with the use of `SafeRunner` to wrap any possible exceptions. There is no new public API here, only an internal interface implemented by CDT LSP and wired up with OSGi services.

Any thoughts/objections?

PR #1095, opened 2025-02-23
CMake built-ins detection successful when compile_commands.json compiler

betamaxbandit:
path is relative

Under certain conditions*, the compiler param in the command field of the compile_commands.json is a relative path rather than absolute. When this happens, the built-ins detection was not successful and the following exception was thrown:

!ENTRY org.eclipse.cdt.core 4 0 2025-02-23 20:32:10.752 !MESSAGE Error: Cannot run program "gcc": Launching failed !STACK 0
java.io.IOException: Cannot run program "gcc": Launching failed
	at org.eclipse.cdt.utils.spawner.Spawner.exec(Spawner.java:450)
	at org.eclipse.cdt.utils.spawner.Spawner.<init>(Spawner.java:147)
	at org.eclipse.cdt.utils.spawner.Spawner.<init>(Spawner.java:134)
	at org.eclipse.cdt.utils.spawner.ProcessFactory$Builder.start(ProcessFactory.java:273)
	at org.eclipse.cdt.utils.spawner.ProcessFactory.exec(ProcessFactory.java:366)
	at org.eclipse.cdt.core.CommandLauncher.execute(CommandLauncher.java:189)
	at org.eclipse.cdt.jsoncdb.core.internal.builtins.CompilerBuiltinsDetector.detectBuiltins(CompilerBuiltinsDetector.java:111)
	at org.eclipse.cdt.jsoncdb.core.CompileCommandsJsonParser.detectBuiltins(CompileCommandsJsonParser.java:290)
	at org.eclipse.cdt.jsoncdb.core.CompileCommandsJsonParser.processJsonFile(CompileCommandsJsonParser.java:193)
	at org.eclipse.cdt.jsoncdb.core.CompileCommandsJsonParser.parse(CompileCommandsJsonParser.java:455)
	at org.eclipse.cdt.cmake.core.CMakeBuildConfiguration.processCompileCommandsFile(CMakeBuildConfiguration.java:361)
	at org.eclipse.cdt.cmake.core.CMakeBuildConfiguration.build(CMakeBuildConfiguration.java:241)

This meant that source file includes were not indexed and could not be opened using Open Declaration (F3) and info markers of the following type appeared in the Problems view:

  gcc -E -P -dM -Wp,-v
"...extCmakegcc\\build\\cmake.run.win32.x86_64.Local\\detect_compiler_builtins.c"
  Cannot run program "gcc": Launching failed

  Error: Program "gcc" not found in PATH
  PATH=[...]
    extCmakegcc		Unknown	Compiler Builtins Detector Problem
  gcc -E -P -dM -Wp,-v
"...extCmakegcc\\build\\cmake.run.win32.x86_64.Local\\detect_compiler_builtins.c"
  Cannot run program "gcc": Launching failed

This patch fixes the environment issue for Core Build projects, by calling ICBuildConfiguration.setBuildEnvironment(Map<String, String>) as part of the built-ins detection setup, thereby supporting absolute and relative compiler paths.

Addresses Issue: CDT CMake Improvements #1000, IDE-82683-REQ-011 Source code navigation and Built-ins detection

*: CMake produces relative compiler path

When the CMAKE_<LANG>_COMPILER variable (eg CMAKE_C_COMPILER) is set in the CMakeLists.txt *after* the project() or language commands, it causes a relative path to be used. For example, in the CMakeLists.txt below, gcc is set after project() command:

  cmake_minimum_required(VERSION 3.10)
  project (track2)
  set(CMAKE_C_COMPILER gcc)
  add_executable(${PROJECT_NAME} track2.c)

The above script creates a relative compiler path in compile_commands.json:
  "command": "gcc   -O3 -DNDEBUG -o ...

Normally the CMAKE_C_COMPILER variable should be set before the project() comannd.

  "command": "C:\\msys64\\mingw64\\bin\\c++.exe -IC...

### DIFF
--- a/jsoncdb/org.eclipse.cdt.jsoncdb.core/src/org/eclipse/cdt/jsoncdb/core/internal/builtins/CompilerBuiltinsDetector.java
+++ b/jsoncdb/org.eclipse.cdt.jsoncdb.core/src/org/eclipse/cdt/jsoncdb/core/internal/builtins/CompilerBuiltinsDetector.java
@@ -17,10 +17,13 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Objects;
+import java.util.Properties;
 
 import org.eclipse.cdt.core.CCorePlugin;
 import org.eclipse.cdt.core.ConsoleOutputStream;
 import org.eclipse.cdt.core.ICommandLauncher;
+import org.eclipse.cdt.core.build.ICBuildConfiguration;
+import org.eclipse.cdt.core.build.ICBuildConfigurationManager;
 import org.eclipse.cdt.core.resources.IConsole;
 import org.eclipse.cdt.jsoncdb.core.IParserPreferences;
 import org.eclipse.cdt.jsoncdb.core.IParserPreferencesAccess;
@@ -29,6 +32,8 @@ import org.eclipse.cdt.jsoncdb.core.participant.IRawSourceFileInfo;
 import org.eclipse.cdt.jsoncdb.core.participant.builtins.IBuiltinsDetectionBehavior;
 import org.eclipse.cdt.jsoncdb.core.participant.builtins.IBuiltinsOutputProcessor;
 import org.eclipse.cdt.jsoncdb.core.participant.builtins.OutputSniffer;
+import org.eclipse.cdt.utils.spawner.EnvironmentReader;
+import org.eclipse.core.resources.IBuildConfiguration;
 import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.CoreException;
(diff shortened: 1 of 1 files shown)


betamaxbandit:
Hi @jonahgraham 
I've created a draft PR top fix this env issue.

1) Is the commit message too long?

2) I could not think of a way of creating a JUnit for this.
Ideally I'd like to test the return result of, to check it contains the expected env:
org.eclipse.cdt.jsoncdb.core.internal.builtins.CompilerBuiltinsDetector.getEnvp(IProject)
But that method is private.
Any ideas on how to tackle this please?

3) I do have detailed instructions for a manual test. Should I simply attach those to this PR?

jonahgraham:
This code doesn't work right, interestingly it suffers from the same problem as the last point in this comment https://github.com/eclipse-cdt/cdt/pull/1067#issuecomment-2657796998

This is the return value of getenvp:

image


Without your patch it looks like this:

image


Sadly neither is correct. In the first the `null` is clearly wrong, in the second the PATH is not pre-pended to the incoming PATH.

I think I will fix this @betamaxbandit - but I will send back to you for review.

jonahgraham:
In the new code the `null` arrives here

https://github.com/eclipse-cdt/cdt/blob/2a62d586213af5307a5188c68be2370d81efdf32/core/org.eclipse.cdt.core/src/org/eclipse/cdt/internal/core/envvar/EnvironmentVariableManager.java#L464

This code assumes that the value being prepended to is non-null. I will include the fix for this alongside this commit.

jonahgraham:
@betamaxbandit and I reviewed this together in a pair programming session

PR #1096, opened 2025-02-25
Avoid null value being converted to "null" string in environment

jonahgraham:
If the CDT Variable is APPEND or PREPEND, and the incoming environment did not contain that variable name, the resulting environment would have null.

For example, if PATH was not set on the incoming environment, and PATH was supposed to be prepended with /usr/bin, this method would have set PATH=/usr/bin:null

This change ensures that the delimeter + null are not prended/appended when the incoming value is null.

Note that this fixes the "null" appearing in the output on the last point in this comment https://github.com/eclipse-cdt/cdt/pull/1067#issuecomment-2657796998


jonahgraham:
@betamaxbandit and I reviewed this together in a pair programming session

PR #1097, opened 2025-02-25
Include system environment when debugging core build

jonahgraham:
When doing a local launch, the only environment passed to the debuggee was the modified variables, typically just PATH. This meant all the other environment was lost.

This change does the same thing as all the other calls to setBuildEnvironment and passes in the initial system environment.

I added a small note to the API of setBuildEnvironment to try to avoid this error in the future.

Note that this fixes the missing incoming environment part of the last point in this comment https://github.com/eclipse-cdt/cdt/pull/1067#issuecomment-2657796998

### DIFF
--- a/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/ICBuildConfiguration.java
+++ b/core/org.eclipse.cdt.core/src/org/eclipse/cdt/core/build/ICBuildConfiguration.java
@@ -151,6 +151,9 @@ public interface ICBuildConfiguration extends IAdaptable, IScannerInfoProvider {
 	 * @param env
 	 *            build environment
 	 * @since 6.1
+	 * @implNote Ensure you pass a new non-empty map containing the base environment
+	 * as this method modifies the passed in map adding the ICBuildConfiguration specific
+	 * overrides, such as applying the project's build environment.
 	 */
 	default void setBuildEnvironment(Map<String, String> env) {
 	}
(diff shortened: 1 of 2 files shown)


betamaxbandit (commented):
Q: Is a similar change necessary in the Run equivalent?

org.eclipse.cdt.debug.internal.core.launch.CoreBuildLocalRunLaunchDelegate.launch(ILaunchConfiguration, String, ILaunch, IProgressMonitor)

jonahgraham:
> Q: Is a similar change necessary in the Run equivalent?
> 
> org.eclipse.cdt.debug.internal.core.launch.CoreBuildLocalRunLaunchDelegate.launch(ILaunchConfiguration, String, ILaunch, IProgressMonitor)

No, the run already does the correct thing.

---

@betamaxbandit said in https://github.com/eclipse-cdt/cdt/pull/1095#issuecomment-2681871198

> I downloaded your latest PS for this change and applied changes from: #1096 #1097 But when I attempt to launch a debug session on Windows, I get an error:
> 
> ```
> !MESSAGE Failure to start debug session
> !STACK 0
> java.util.concurrent.ExecutionException: org.eclipse.core.runtime.CoreException: Failed to execute MI command:
> -gdb-set env =:: = ::\\
> Error message from debugger back end:
> Argument required (environment variable to set).
> 	at org.eclipse.cdt.dsf.concurrent.Query.get(Query.java:112)
> 	at org.eclipse.cdt.dsf.gdb.internal.launching.CoreBuildLocalDebugLaunchDelegate.launch(CoreBuildLocalDebugLaunchDelegate.java:149)
> 	at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:777)
> 	at org.eclipse.debug.internal.core.LaunchConfiguration.launch(LaunchConfiguration.java:688)
> 	at org.eclipse.debug.internal.ui.DebugUIPlugin.buildAndLaunch(DebugUIPlugin.java:1054)
> 	at org.eclipse.debug.internal.ui.DebugUIPlugin$1.run(DebugUIPlugin.java:1257)
> 	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
> Caused by: org.eclipse.core.runtime.CoreException: Failed to execute MI command:
> -gdb-set env =:: = ::\\
> Error message from debugger back end:
> Argument required (environment variable to set).
> 	... 7 more
> ```

This is due the the "special" environment variables on Windows that have no key and should have been exlcuded.

jonahgraham:
I am removing this from RC1 consideration - this fix isn't quite right, for the obvious as stated in the above comment, but also because it does too many `-gdb-set env` - it is doing one for every environment variable, but it should only be doing it for the variables that have different values than in the incoming environment.

jonahgraham:
My latest commit saves my current state of work - I think this is correct, but some more thought it needed.

PR #1107, opened 2025-03-07
Prepare repo for CDT 12.1.0

jonahgraham:


jonahgraham:
This PR is needed before we can release a future CDT 12.1.0. For now I am having some trouble as it is saying most bundles need a version bump, but I don't think they should. I will revisit this later in the development cycle. In the meantime it means some cleanup may be needed if other patches get merged.

jonahgraham:
:partying_face: now that CI is back up and running from both the IT outage + the out of date config fixed with this PR I can now promote a build to SimRel 2025-06 M2 this week.

PR #1109, opened 2025-03-08
Add accessor for included pty instance

laeubi:
Currently if I get passed a Process from by a method call I can check it for the Spawner and can already call special methods like hangup(). But there is no way to get access the the PTY used to create the Spawner so for example one can't call PTY#setTerminalSize or query any other properties.

This now adds a new method pty() that returns this instance for further investigation and actions.

See
- https://github.com/eclipse-platform/eclipse.platform/pull/1762#issuecomment-2708081348

### DIFF
--- a/core/org.eclipse.cdt.core.native/about.properties
+++ b/core/org.eclipse.cdt.core.native/about.properties
@@ -1,5 +1,5 @@
 ###############################################################################
-# Copyright (c) 2014, 2024 Contributors to the Eclipse Foundation
+# Copyright (c) 2014, 2025 Contributors to the Eclipse Foundation
 #
 # See the NOTICE file(s) distributed with this work for additional
 # information regarding copyright ownership.
@@ -24,7 +24,7 @@ blurb=C/C++ Development Tooling Native Utilities\n\
 Version: {featureVersion}\n\
 Build id: {0}\n\
 \n\
-Copyright (c) 2014, 2024 Contributors to the Eclipse Foundation
+Copyright (c) 2014, 2025 Contributors to the Eclipse Foundation
 \n\
 See the NOTICE file(s) distributed with this work for additional\n\
 information regarding copyright ownership.\n\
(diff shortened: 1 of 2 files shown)


laeubi:
The cleancheck fails with

```
[ERROR] Changes not staged for commit:
[ERROR] 	modified:    core/org.eclipse.cdt.core.native/about.properties
```

but I can't see how this relates to my commit...

laeubi:
@jonahgraham @akurtakov any chance to push this forward?

laeubi:
It seems the cleanup is modify the `about.properties` file but I can't really reproduce this locally... any hint would be appreciated!

laeubi:
Looks like it was actually the **Year** that needs to be changed here...

laeubi:
Build is finally green!

stbischof:
+1

jonahgraham:
Thanks @laeubi for this fix - a build that includes this should be published to SimRel on Monday if all goes to plan. The continuous build is here https://download.eclipse.org/tools/cdt/builds/cdt/main/ and (assuming enough services stay up - https://www.eclipsestatus.io/) should be populated in a few hours.

Note that the `@since` will be updated to the correct value when I resolve #1107

laeubi:
> a build that includes this should be published to SimRel on Monday if all goes to plan. 

This sounds great! will it be the 12.1.0 release then? Or are there special Milestone releases?

jonahgraham:
> This sounds great! will it be the 12.1.0 release then? Or are there special Milestone releases?

It will be 12.1.0 M2 - the 12.1.0 release is planned to line up with 2025-06 release.

However, I am very open to the idea of doing the 12.1.0 release early (with 2025-06 M3 or RC1) if that helps dependency management. Please let me know and I can bring to the CDT call on 14 May - agenda: https://github.com/eclipse-cdt/cdt/discussions/1157.

laeubi:
@akurtakov @merks If I reference some CDT site in the platform target, would it be enough if we have the stable site at RC1? If yes this sounds like a good idea and I can test with the 12.1.0 M2 first.

akurtakov:
What would the CDT in platform target be needed for ? Haven't you moved what's needed to platform?

laeubi:
@akurtakov I only moved the UI for displaying a VT100 terminal, but we still need to launch the process in a PTY what is [org.eclipse.cdt.core.native](https://github.com/eclipse-cdt/cdt/tree/main/core/org.eclipse.cdt.core.native) + native fragments.

akurtakov:
Eclipse Platform p2 repo is supposed to be standalone - aka to not depend on any other p2 repo.  If your changes keep things that way - fine by me.

PR #1110, opened 2025-03-11
fix: remove the extra space

Sheena-143:
remove the extra space


jld01:
Hi @Sheena-143, there are no files changed in this PR. Can you amend it please?

Issue #1126, opened 2025-04-04
Build order of reference projects is incorrect


wntrkid:
**Description**
It is possible to arrange the build order of referenced projects in managed C projects. But when building the order is incorrect.

**To Reproduce**

1. Create managed C projects i.e. "Grey", "Hello", "StrTest".
2. For project "StrTest" go to "properties > C/C++ General Paths and Symbols... > References"
3. Check "Hello" and "Gray" and bring "Hello" to the top of the list
4. Build project "StrTest" and check the global build console.

**Version:**
- Eclipse 2023-09
- CDT Version 11.3.0

![Image](https://github.com/user-attachments/assets/ebe08fe5-685d-438c-9648-523117e5a6b9)

![Image](https://github.com/user-attachments/assets/1c10c46a-99c7-4c07-98f2-1126321bb552)

wntrkid:
What I found is that org.eclipse.core.internal.resources.Workspace.class is computing a digraph representing the build order or all projects but without respecting the project order as defined in the CDT "references" (stored in .cproject, search for "org.eclipse.cdt.core.cfg.export.settings.sipplier").

However the CDT implementation of builder "org.eclipse.cdt.managedbuilder.internal.core.CommonBuilder.class" does of course take care for the referenced projects but again with incorrect order.

At the end the projects are got build twice, both times in wrong order.

PR #1127, opened 2025-04-04
Preserve the order of referenced projects (#1126)

AndreasMehler:
With this it is guaranteed the reference projects are kept in order. This will fix the build problems described in #1126 with multiple projects referenced that may have dependencies between each other and the order matters.

Tested with CDT 11.3.0


AndreasMehler:
mistaken, will not fix the problem. closed

PR #1131, opened 2025-04-08
Remove use of deprecated constant to allow Platform to delete the constant

jonahgraham:
See https://github.com/eclipse-platform/eclipse.platform/pull/1784#pullrequestreview-2702035505


ruspl-afed (approved):
+1 for switching from internal to API

PR #1136, opened 2025-04-09
p2 composites for CDT LSP 3.1

jonahgraham:


jonahgraham:
Part of https://github.com/eclipse-cdt/cdt-lsp/issues/480

PR #1138, opened 2025-04-10
Fix for the compiler path in compile_commands.json file

alicetrifu:
Compiler path was wrongly created when anything was set on the "Prefix" option from "Cross Settings". The code needs to check if anything is set there before creating the default path for the compiler.


build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -229,10 +229,10 @@ private List<CompilationDatabaseInformation> populateObjList(IProject project, I
 						outputLocation + "", inputStrings, sourceLocation, outputLocation); //$NON-NLS-1$
 
 				IBuildMacroProvider provider = ManagedBuildManager.getBuildMacroProvider();
-				String compilerName = CompilationDatabaseGenerator.getCompilerName(tool);
+				String compilerName = CompilationDatabaseGenerator.getCompilerName(tool, config, cmdLInfo);
```
jonahgraham, line 232: Minor change - instead of passing `cmdLInfo` to `getCompilerName`, move this line below `String commandLine = cmdLInfo.getCommandLine();` and pass `commandLine` to `getCompilerName`

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -469,9 +469,9 @@ private String findCompilerInPath(ITool tool, IConfiguration config) {
 		return null;
 	}
 
-	private static String getCompilerName(ITool tool) {
-		String compilerCommand = tool.getToolCommand();
-		String[] arguments = CommandLineUtil.argumentsToArray(compilerCommand);
+	private static String getCompilerName(ITool tool, IConfiguration config, IManagedCommandLineInfo cmdLInfo) {
```
jonahgraham, line 472: `tool` and `config` arguments are unused AFAICT.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -438,11 +438,11 @@ public boolean visit(IResourceProxy proxy) throws CoreException {
 
 	}
 
-	private String findCompilerInPath(ITool tool, IConfiguration config) {
+	private String findCompilerInPath(ITool tool, IConfiguration config, IManagedCommandLineInfo cmdLInfo) {
```
jonahgraham, line 441: I think the `toolMap` cache is now incorrect - the key should probably be tool + compiler name before it is expanded, or even just the compiler name before it is expanded

jonahgraham (changes requested):
This looks like roughly the correct solution. Just some cleanup and review the caching.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -229,10 +229,10 @@ private List<CompilationDatabaseInformation> populateObjList(IProject project, I
 						outputLocation + "", inputStrings, sourceLocation, outputLocation); //$NON-NLS-1$
 
 				IBuildMacroProvider provider = ManagedBuildManager.getBuildMacroProvider();
-				String compilerName = CompilationDatabaseGenerator.getCompilerName(tool);
 				String commandLine = cmdLInfo.getCommandLine();
+				String compilerName = CompilationDatabaseGenerator.getCompilerName(commandLine);
```
jonahgraham, line 233: The replace on the next line does not work reliabliy. Please revisit this bit of code. For example, if commandLine has this contents:

```
"name of compiler" -O0 input.c -i output.o
```

Then `compilerName` will be `name of compiler` and after the replace the command line will be:

```
"" -O0 input.c -i output.o
```

which leads to, `resolvedOptionFileContents` being:

```
"/full/path/to/name of compiler" "" -O0 input.c -i output.o
```

Instead of doing `replace` use the parsed information of `CommandLineUtil.argumentsToArray` to convert the `commandLine` into two parts, the compilerName (i.e. argv[0]) and the rest of the command line as a string.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -438,15 +438,15 @@ public boolean visit(IResourceProxy proxy) throws CoreException {
 
 	}
 
-	private String findCompilerInPath(ITool tool, IConfiguration config) {
-		if (toolMap.containsKey(tool)) {
-			return "\"" + toolMap.get(tool) + "\""; //$NON-NLS-1$//$NON-NLS-2$
+	private String findCompilerInPath(ITool tool, IConfiguration config, String compilerName) {
+		String cacheKey = tool + " " + compilerName; //$NON-NLS-1$
```
jonahgraham, line 442: The `toString` of ITool is just the externalized name of the tool, not really suitable for using as a key here.

That said and on further examination, we only need to consider `tool` for the cache key if the compilerName is not unique enough. AFAICT it is unique with a compilation, so probably easiest to just drop tool entirely here.

```suggestion
		String cacheKey = compilerName; //$NON-NLS-1$
```

jonahgraham (commented):
This works well in the normal cases, but there are some corner cases that aren't handled well, see the individual comments.

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -469,13 +471,20 @@ private String findCompilerInPath(ITool tool, IConfiguration config) {
...
 		}
 		return arguments[0];
 	}
 
+	private static String getCompilerArgs(String commandLine) {
+		String[] arguments = CommandLineUtil.argumentsToArray(commandLine);
+		if (arguments.length <= 1) {
+			return ""; //$NON-NLS-1$
+		}
+		return String.join(" ", Arrays.copyOfRange(arguments, 1, arguments.length)); //$NON-NLS-1$
```
jonahgraham, line 487: This is an invalid way of converting a list to a command line string. For example `gcc "My file.c" -o "My file.o"` would turn into `"/usr/bin/gcc" My file.c -o My file.o` which won't be interpreted properly. Please ensure quoting is done as appropriate and meets the spec of compile_commands.json linked [here](https://clang.llvm.org/docs/JSONCompilationDatabase.html#:~:text=command%3A%20The%20compile%20command%20as%20a%20single%20shell%2Descaped%20string.%20Arguments%20may%20be%20shell%20quoted%20and%20escaped%20following%20platform%20conventions%2C%20with%20%E2%80%98%22%E2%80%99%20and%20%E2%80%98%5C%E2%80%99%20being%20the%20only%20special%20characters.%20Shell%20expansion%20is%20not%20supported.)

org.eclipse.cdt.utils.CommandLineUtil.argumentsToString(Collection<String>, boolean) may or may not actually match the requirements. I hope that it does as it would be simpler than implementing a new case. If argumentsToString does not meet requirements, I recommend adding a new method to CommandLineUtil and the appropriate tests to CommandLineUtilTest

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -438,15 +440,15 @@ public boolean visit(IResourceProxy proxy) throws CoreException {
 
 	}
 
-	private String findCompilerInPath(ITool tool, IConfiguration config) {
-		if (toolMap.containsKey(tool)) {
-			return "\"" + toolMap.get(tool) + "\""; //$NON-NLS-1$//$NON-NLS-2$
+	private String findCompilerInPath(ITool tool, IConfiguration config, String compilerName) {
```
jonahgraham, line 443: `tool` is now unused. 

```suggestion
	private String findCompilerInPath(IConfiguration config, String compilerName) {
```

build/org.eclipse.cdt.managedbuilder.core/src/org/eclipse/cdt/managedbuilder/internal/core/jsoncdb/generator/CompilationDatabaseGenerator.java
```diff
@@ -70,7 +71,7 @@ public final class CompilationDatabaseGenerator {
 	 * Checked on each build
 	 * Used before we look up the environment
 	 */
-	private Map<ITool, String> toolMap = new HashMap<>();
+	private Map<String, String> toolMap = new HashMap<>();
```
jonahgraham, line 74: Because we now have a key of a string it is much less obvious what that key is, so please document it. Perhaps a line like this in the javadoc: `Map of compiler name (as calculated by {@link #getCompilerName(String)}) to the absolute path of the compiler.`

jonahgraham (approved):

ghentschke:
@alicetrifu  Is this PR still work in progress?

alicetrifu:
Hello @ghentschke , 

This should be done. I see that the logs are not available anymore for this, but the comments should be fixed. I see that meanwhile this got into conflict. Should I fix this?

ghentschke:
There is still a failed unit test. I am not sure if it relates to this PR.

alicetrifu:
@ghentschke , This does not seem to be related to my PR indeed. But I am not sure why I have the problem with the Licence. I had this in the past as well and @Kummallinen rerun this for me. I think there are some issues with my permissions. Any idea on what I should do next to fix this ? Thank you!

ghentschke:
> But I am not sure why I have the problem with the Licence. 

I think there is a general problem with the license server at the moment. See https://github.com/eclipse-cdt/cdt-lsp/pull/535#issuecomment-3247237534

alicetrifu:
Thamk you @ghentschke! Do I need to make any additional changes here? Not sure how to fix the failing test since I am sure it's not releated to this PR. Please let me know what you think about the status here. Thank you!

Issue #1142, opened 2025-04-14
Reduce redundency of ToggleInstructionStepModeCommand


raghucssit:
ToggleInstructionStepModeCommand is currently contributed at 4 palces.
1. Main Toolbar
2. Main Menu-->Run
3. Debug View Toolbar
4. Debug View context menu.

![Image](https://github.com/user-attachments/assets/c756aa8d-32fc-4215-9c0d-c8e6ac2db24f)

This is seems to be redundent. I think we can restrict it to Debug View. Because this command makes sanse only if a CDT application is under debug. Adding this to Main Toolbar and Run Menu pollutes those areas.

Background:
Eclipse workbench shows all the contributions declared in `plugin.xml` if the corresponding plugin is not loaded on eclipse startup. Below is the platform ui issue related to it.
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2900

As we know the above issue. If CDT and JDT both are installed and if user is interested only in Java Development still ToggleInstructionStepModeCommand is shown in the Main Toolbar. So we can remove this command from Main Toolbar and Run Menu.

PR #1143, opened 2025-04-14
Remove ToggleInstructionStepModeCommand from Run Menu and Main Toolbar.

raghucssit:
As ToggleInstructionStepModeCommand is contributed in too many places. We can remove it from Main Toolbar and Run Command.

see https://github.com/eclipse-cdt/cdt/issues/1142


iloveeclipse:
@jonahgraham : a review is welcome.

raghucssit:
@jonahgraham The motivation for this fix is, " If property tester which decides the command has to the visible or not is not loaded then command is visible by default". Once the Property Tester is loaded by any dependents on the platform then it checks with the property tester and hides it. This is a bug/limitation in the eclipse platform.
Here is the detailed discussion:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2900
As ToggleInstructionStepModeCommand is contributed to main toolbar, it is by default is visible even without any context.
So I have decided to remove it from there.
And it continues to be available at 'Debug View' toolbar.
Please check if we can remove this from main toolbar.

jonahgraham:
> @jonahgraham : a review is welcome.

:cry: I'm sorry that this slipped through the cracks - I was having extended down time when this came in.

Kummallinen:
Sorry only just saw the notification for this.

For embedded development this is commonly used so would annoy our users if it were removed. While we can put it back for our products users of Embedded CDT would likely miss it & users of multiple plugins sets may end up with duplicates if vendors need to add this back in themselves

iloveeclipse:
The original issue is that the button is shown in non-CDT context as long as CDT is not loaded. This is a limitation of an extension point because it has to use tester classes to check the button / CDT debugging state, and we don't want CDT to be loaded if it is not used.

What about following options: 
1) Declare an activity that is disabled by default & bound that to the button. Whoever needs that can enable activity (either by product customization or via Preferences->Capabilities).
2) Declare an activity that is disabled by default & bound that to the button. Additionally, add code to the corresponding bundle activator that would enable the activity on bundle startup and disable on shutdown.
3) Declare an activity that is disabled by default & bound that to the button. Additionally, enable the activity first time CDT debugger starts (and do not disable it later).
4) Any other ideas ? ...

WDYT?

ruspl-afed:
What if we add an "id" attribute for these contributions to let interested parties filter them out with E3 activity?

raghucssit:
> contributions

Contribution Element already have an `id` attribute.
But Menu Manager don't use this `id` for filtering, Instead it uses actual `Command ID`.
I am working on supporting this `id` for activity filtering in Menu Manger.
Please check below image where I added attribute to Menu Contribution.
image

Is my understanding correct regarding your suggestion ?

ruspl-afed:
> I am working on supporting this id for activity filtering in Menu Manger

Thank you!

Since command `id` and menu contribution `id` is a kind of API (for example to filter it out with activity) I would recommend to avoid `internal` segment in it and find better formulation for it, considering the following:
1) start from contributing bundle symbolic name
2) add segments like `menu` and `command` that points to an extension and element
3) end with a key for concrete contribution ("toggle instruction step mode" in our case)

As a result, we may have something like 
**`org.eclipse.cdt.debug.ui.menu.command.toggle_instruction_step_mode`**
And it's relatively easy to detect and filter such an `id` for both machines and humans.

We use this approach for our products and recommend it to our customers because it greatly simplifies code maintenance.

iloveeclipse:
@raghucssit : please update this PR, it has merge issues.
@jonahgraham : any chance for a review?

iloveeclipse:
Ideally users shouldn't need to use "low-level" and not well known features like Capabilities to have a clean and not overloaded UI. 

So while one could deactivate the extra buttons/menus via Capabilities, better to have clean UI from beginning. So would be nice if someone from CDT team could review this PR. Maybe @jld01?

jonahgraham:
I think this is ok - I have added it to the agenda of the CDT call today https://github.com/eclipse-cdt/cdt/discussions/1340 to progress this.

I checked the help - and the two references to it in the help only refer to the Debug View anyway:

- https://help.eclipse.org/latest/topic/org.eclipse.cdt.doc.user/tasks/cdt_t_disassembly.htm
- https://help.eclipse.org/latest/topic/org.eclipse.cdt.doc.user/reference/cdt_u_dbg_view.htm

Kummallinen (commented):
See my previous comment about moving to a new bundle.

(forgot to make that as a proper review comment)

jonahgraham:
This seems stuck again - moving to a draft until we have a proposal that is widely acceptable by current CDT committers.

PR #1158, opened 2025-05-07
Only import the terminal.control

laeubi:
The terminal control is migrated to the Eclipse platform, as those it should be handled like a thrid-party dependency an not directly included with an exact version in the feature.

This now changes it to a requirement with a lower bound of 5.5 excluding the next major version. That way CDT and other consumers can gracefully upgrade to the new platform variant.

In general I think one would not need the feature at all today, as dependencies will be pulled in anyways if installing any of the other ones, but removing it is currently a bit out of scope, so this seems the most compatible way.

FYI @merks

### DIFF
--- a/terminal/features/org.eclipse.tm.terminal.control.feature/feature.xml
+++ b/terminal/features/org.eclipse.tm.terminal.control.feature/feature.xml
@@ -35,15 +35,9 @@
    </license>
 
    <requires>
+      <import plugin="org.eclipse.tm.terminal.control" version="5.5.0" match="compatible"/>
       <import plugin="org.eclipse.core.runtime"/>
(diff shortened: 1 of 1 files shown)


laeubi:
@jonahgraham can you review/merge?

merks:
I don’t have a strong opinion. Though, as you suggest, the import is not necessary. 

This change will result in that bundle not being in the update site I believe. It is in the Platform’s update site so will be available for simrel aggregation.

jonahgraham (approved):

jonahgraham:
> I will be removing the source from CDT too for this bundle.

See https://github.com/eclipse-cdt/cdt/pull/1166

laeubi:
Great!

PR #1159, opened 2025-05-07
Update to Eclipse 2025-03

laeubi:
To migrate to the moved terminal.control CDT needs to upgrade to the next release (2025-06), as a first step this upgrades to the current release (2025-03).

FYI @merks

### DIFF
--- a/core/org.eclipse.cdt.ui/META-INF/MANIFEST.MF
+++ b/core/org.eclipse.cdt.ui/META-INF/MANIFEST.MF
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.cdt.ui; singleton:=true
-Bundle-Version: 9.0.0.qualifier
+Bundle-Version: 9.0.100.qualifier
 Bundle-Activator: org.eclipse.cdt.ui.CUIPlugin
 Bundle-Vendor: %providerName
 Bundle-Localization: plugin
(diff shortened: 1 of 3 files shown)


laeubi:
@jonahgraham any idea about the code clean error? Could it be that something compiles differently? Should I just bump the version of affected bundles?

jonahgraham:
> Should I just bump the version of affected bundles?

Yes.

We need to remove TypedListener use  to follow https://github.com/eclipse-platform/eclipse.platform.swt/pull/1615

More details:

> @jonahgraham any idea about the code clean error? Could it be that something compiles differently? 

Yes, but generally this should not happen when dependencies change. In this case it happens because CDT uses internals in particular the change in compiler output is introduced by this change https://github.com/eclipse-platform/eclipse.platform.swt/pull/1615

With 4.34 this code:

https://github.com/eclipse-cdt/cdt/blob/170e654b4796bad1453ae85a427b97317d67a69a/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/internal/ui/ImageCombo.java#L261

calls `Method org/eclipse/swt/widgets/TypedListener."<init>":(Lorg/eclipse/swt/internal/SWTEventListener;)V` but with 4.35 it calls `Method org/eclipse/swt/widgets/TypedListener."<init>":(Ljava/util/EventListener;)V`

Since `Method org/eclipse/swt/widgets/TypedListener."<init>":(Ljava/util/EventListener;)V` is not public in 4.34, we need to bump dependency.

(The reason the constructor call changes is because `ModifyListener` now extends `EventListener` directly instead of via `SWTEventListener`)

This means the code is no longer ABI compatible with the min version of dependency as listed. This looks a lot like https://bugs.eclipse.org/bugs/show_bug.cgi?id=536448

jonahgraham:
PS this is the steps I did to figure out what changed have been documented in  see https://github.com/eclipse-cdt/cdt/pull/1161

laeubi:
I can't find why my ECA is not accepted but adjusted the build now.

laeubi:
I created https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/6062

laeubi:
@jonahgraham is anything left here or can it be merged?

jonahgraham:
Lets bring this to Platform 2025-06 M3 once it comes out Friday.

Assigning to myself.

jonahgraham:
Replaced by https://github.com/eclipse-cdt/cdt/pull/1165

PR #1169, opened 2025-05-27
Set CC and CXX environment variables for clang toolchain

### LINKED ISSUE #1140: CMake toolchain selection from LaunchBar apparently ignored
I am using the latest CDT sources under PDE. I have both GNU and LLVM toolchains installed via MSYS2.

When I create a new CMake project, the default toolchain is `gcc`. If I then select `clang` via the _Launch Configuration_ settings dialog, the modified toolchain selection is apparently ignored:

Image

The project continues to build using the GNU toolchain:

```
CMAKE_CXX_COMPILER:FILEPATH=C:/msys64/ucrt64/bin/c++.exe
CMAKE_C_COMPILER:FILEPATH=C:/msys64/ucrt64/bin/cc.exe
```

This issue persists even if I clean the project and delete the build folder.

After selecting the `clang` toolchain, I expect the project to build using `clang`.

betamaxbandit:
Sets the C and C++ compiler environment variables so CMake uses Clang instead of the system default compiler.

Fixes #1140

### DIFF
--- a/build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
+++ b/build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
@@ -11,8 +11,12 @@
 package org.eclipse.cdt.build.gcc.core;
 
 import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 
 import org.eclipse.cdt.core.build.IToolChainProvider;
+import org.eclipse.cdt.core.envvar.EnvironmentVariable;
 import org.eclipse.cdt.core.envvar.IEnvironmentVariable;
 
 /**
@@ -28,7 +32,29 @@ public class ClangToolChain extends GCCToolChain {
 
 	public ClangToolChain(IToolChainProvider provider, Path pathToToolChain, String arch,
 			IEnvironmentVariable[] envVars) {
-		super(provider, pathToToolChain, arch, envVars);
+		super(provider, pathToToolChain, arch, addClangEnvVars(envVars));
+	}
+
+	private static IEnvironmentVariable[] addClangEnvVars(IEnvironmentVariable[] envVars) {
+		List<IEnvironmentVariable> envVarsNew = new ArrayList<>(Arrays.asList(envVars));
+		/*
+		 * Set CC and CXX environment variables for clang and clang++. This is equivalent to setting these in the
+		 * CMakeLists.txt:
+		 * set(CMAKE_C_COMPILER clang)
+		 * set(CMAKE_CXX_COMPILER clang++)
+		 */
+		addIfAbsent("cc", "clang", envVarsNew); //$NON-NLS-1$ //$NON-NLS-2$
+		addIfAbsent("cxx", "clang++", envVarsNew); //$NON-NLS-1$ //$NON-NLS-2$
+
+		return envVarsNew.toArray(new EnvironmentVariable[0]);
+	}
+
+	private static void addIfAbsent(String name, String value, List<IEnvironmentVariable> envVarsNew) {
+		boolean isAbsent = envVarsNew.stream().noneMatch(ev -> ev.getName().equals(name));
+
+		if (isAbsent) {
+			envVarsNew.add(new EnvironmentVariable(name, value, IEnvironmentVariable.ENVVAR_APPEND, null));
+		}
 	}
 
 	@Override


build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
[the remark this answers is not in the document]
Kummallinen: Additional point, in CI you'd likely just set the same environment variables in the docker container or CI pipeline used for build to nudge CMake into selecting Clang not GCC.

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
```diff
@@ -28,7 +32,29 @@ public class ClangToolChain extends GCCToolChain {
...
+
+	private static IEnvironmentVariable[] addClangEnvVars(IEnvironmentVariable[] envVars) {
+		List<IEnvironmentVariable> envVarsNew = new ArrayList<>(Arrays.asList(envVars));
+		/*
+		 * Set CC and CXX environment variables for clang and clang++. This is equivalent to setting these in the
+		 * CMakeLists.txt:
+		 * set(CMAKE_C_COMPILER clang)
+		 * set(CMAKE_CXX_COMPILER clang++)
+		 */
+		addIfAbsent("cc", "clang", envVarsNew); //$NON-NLS-1$ //$NON-NLS-2$
```
betamaxbandit, line 46: cc and cxx should be upper case!

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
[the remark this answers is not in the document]
jld01: > The environment variables are to select the compiler for a CMake project which is compiler agnostic (fairly common case) without needing to create a toolchain file when one isn't needed.

I don't think the end user will necessarily appreciate a distinction like this. All they observe is a toolchain selection UI which gives the impression that they can select the toolchain (both path and compiler name) that applies to the project.

If it is common practice for developers using CMake to rely on environment to select toolchain _technology_ (`gcc` or `clang`) in the absence of a toolchain file, then perhaps we just need to clarify within the UI that the selection applies within the Eclipse IDE only and may be overriden by directives in `CMakeLists.txt`, by CMake command line arguments or by toolchain file association.

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
```diff
@@ -28,7 +32,29 @@ public class ClangToolChain extends GCCToolChain {
...
 	public ClangToolChain(IToolChainProvider provider, Path pathToToolChain, String arch,
 			IEnvironmentVariable[] envVars) {
-		super(provider, pathToToolChain, arch, envVars);
+		super(provider, pathToToolChain, arch, addClangEnvVars(envVars));
+	}
+
+	private static IEnvironmentVariable[] addClangEnvVars(IEnvironmentVariable[] envVars) {
+		List<IEnvironmentVariable> envVarsNew = new ArrayList<>(Arrays.asList(envVars));
+		/*
+		 * Set CC and CXX environment variables for clang and clang++. This is equivalent to setting these in the
```
jld01, line 41: Do we also need to set the ASM environment variable?

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
betamaxbandit: > Does envVarsNew already include the system environment,
No; its the env from the core build and/or any env vars set by the user for a "User Defined Toolchain".

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
betamaxbandit: > Do we also need to set the ASM environment variable?
Good question. I do not think so. CMake appears to set ASM automatically to clang when CC is set to clang.

I tested this using the attached project.
[cmakeASM.zip](https://github.com/user-attachments/files/20555699/cmakeASM.zip)

In the screendshot below, it can be seen that the ASM is set to clang.exe.

cmakeASM helloworld example

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
```diff
@@ -28,7 +34,81 @@ public class ClangToolChain extends GCCToolChain {
...
+			while (iterator.hasNext()) {
+				IEnvironmentVariable ev = iterator.next();
+				if (ev.getName().equalsIgnoreCase(name)) {
+					found = true;
+					// Replace if existing name is in lower-case and new name is in upper-case
+					if (!ev.getName().equals(name) && name.equals(name.toUpperCase())) {
+						// Original value is respected.
+						String evOldValue = ev.getValue();
+						iterator.remove();
+						envVars.add(new EnvironmentVariable(name, evOldValue));
```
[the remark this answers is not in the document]
betamaxbandit: Yes I see that now. Renaming the variable name is tricky. Perhaps we should forget about it.

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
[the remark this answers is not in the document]
betamaxbandit: Yes I agree.
I have tested this and can reproduce the problem:

In a windows cmd shell,
set CC=clang
set CXX=clang++
Then launch eclipse CDT from this shell.
In a CMake project, in Build Settings, set Toolchain to:
  C:\msys64\ucrt64\bin\gcc.exe
Build project.
Expected: Builds using gcc.
Actual: Builds using clang, eg:
  "-- The C compiler identification is Clang 19.1.7
  -- The CXX compiler identification is Clang 19.1.7
  "

build/org.eclipse.cdt.build.gcc.core.tests/src/org/eclipse/cdt/build/gcc/core/tests/TestClangToolChain.java
```diff
@@ -0,0 +1,218 @@
...
+	 * Where envVars is null.
+	 *
+	 * Expected:
+	 *   IToolChain.getVariables() has items "CC=a/clang.exe", "CXX=a/clang++.exe"
+	 */
+	@Test
+	public void clangExe() throws Exception {
+		IToolChain tc = new ClangToolChain(null, Paths.get("a/clang.exe"), null, null);
+		IEnvironmentVariable[] variables = tc.getVariables();
+		assertThat(Arrays.asList(variables), hasItems( //
```
jonahgraham, line 50: Nice use of assertions here :smile:

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/GCCToolChain.java
jld01, line 1: Can we also support tools with a major version suffix? It is quite common for package managers to allow installation of multiple versions of the same toolchain distinguished by a major version suffix (eg `gcc-10`, `g++-11`).

betamaxbandit: Hi @jld01 ,

sure. I haven't seen these version suffixes before. Can you point me at any installers that use that so I can see for myself please?

jld01: For example on Ubuntu 24.04 LTS, I can install `gcc-13` using `apt install gcc-13`. List of installed files at https://packages.ubuntu.com/noble/amd64/gcc-13/filelist includes: `/usr/bin/gcc-13`

betamaxbandit: > For example on Ubuntu 24.04 LTS, I can install `gcc-13` using `apt install gcc-13`. List of installed files at https://packages.ubuntu.com/noble/amd64/gcc-13/filelist includes: `/usr/bin/gcc-13`

Thanks John.

I've updated the code to accommodate version suffixes.

I've refactored how I do things a bit and used regex more. I'd appreciate another pair of eyes on my regex please.

cc @jonahgraham

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
```diff
@@ -23,12 +31,131 @@
...
+		boolean isCXX = cxxMatcher.matches();
+		boolean isCC = ccMatcher.matches();
+		if (isCC || isCXX) {
+			// Compiler is either clang or clang++
+			String mappedFilename;
+			if (isCXX && CC.equals(desiredType)) {
+				// Compiler is CPP type and desired type is C
+				String version = cxxMatcher.group(1) != null ? "-" + cxxMatcher.group(1) : EMPTY; //$NON-NLS-1$
+				String ext = cxxMatcher.group(2) != null ? cxxMatcher.group(2) : EMPTY;
+				mappedFilename = "clang" + version + ext; //$NON-NLS-1$
```
jld01, line 143: It seems we are splitting the filename into multiple parts using regex groups and then re-combining them here. Could we simplify by using a group that captures the entire filename instead? The existing regex groups describing the optional parts would be nested inside this top-level group. Similarly for `GCCToolChain`.

jld01 (commented):
Looks good, @betamaxbandit, but I wonder if we can simplify? I may have missed something.

build/org.eclipse.cdt.build.gcc.core/src/org/eclipse/cdt/build/gcc/core/ClangToolChain.java
jld01: Hi @betamaxbandit, in your `clang` tool regex you have: `clang(?:-(\\d+))?(\\.exe)?`

You have capturing groups for the optional major version (group 1) and the optional .exe suffix (group 2).

You could use nested groups as follows: `(clang(?:-(\\d+))?(\\.exe)?)`

In that case, group 1 would be the full file name, group 2 would be the major version and group 3 would be the optional .exe suffix. So you would not need to re-assemble the full file name from its constituent parts. In fact, the inner groups could all be non-capturing: `(clang(?:-\\d+)?(?:\\.exe)?)`

I am suggesting that you could eliminate some code that way.

jld01: Hi @betamaxbandit, do you have time to look at this simplification at present? If not, I suggest we merge your existing change for the time being. It's a good improvement.

jonahgraham:
@betamaxbandit  / @jld01 I have added this to the agenda for the call today. I think @jld01's comment of:

>  I suggest we merge your existing change for the time being. It's a good improvement.

Is a good summary and we should probably rebase and merge this change and create new PRs for any follow-ups.

Issue #1175, opened 2025-05-30
Improve toolchain identification on macOS


jld01:
The LLVM toolchain supplied by recent versions of the XCode Command Line Tools package is not identified by CDT, so the toolchain is not presented for selection in the MBS project wizard.

To improve the out-of-box experience for new CDT users on macOS, it would also be useful to automatically locate tools provided by the Homebrew package manager on macOS, in a similar manner to the way we support the MSYS2 package manager on Windows.

jld01:
I am closing this issue. CDT now detects the version of `clang` provided by _Xcode Command Line Tools_ and will locate Homebrew installations.

PR #1176, opened 2025-05-30
Also use clang executable to identify LLVM tools installation

jld01:
Use either `clang` or `llvm-ar` to locate an LLVM tools installation. The XCode _Command Line Tools_ do not include `llvm-ar`. Other installations of LLVM tools that lie ahead of `/usr/bin` on the `PATH` will take precedence.

Relates to: #1175


jld01:
This PR is intended for CDT 12.2.0.

PR #1179, opened 2025-06-02
Fixes User Defined Toolchain clang type forgetfulness

betamaxbandit:
When a User Defined Toolchain of type clang was created (Preferences > C/C++ > Core Build Toolchains > User Defined Toolchains) the Type was reported correctly as clang. But after restarting the workbench the Type was reported as GCC. This is now fixed so the Type always shows as clang for a clang toolchain.

clang tc added - before workbench restart


jonahgraham (approved):
@betamaxbandit LGTM - this can be merged once #1185 is ready

Issue #1180, opened 2025-06-04
'C/C++ Projects' view hangs eclipse(Over 15 seconds) if there are more than 70k files in a directory within the project


raghucssit:
'C/C++ Projects' view hangs i there are more than 70k files in a working tree
This is discovered some file/code file generator generated large number of files in a cpp project.
When user tries to open the directory it has the large number of files it made the eclipse unresponsive for over 15 seconds.

![Image](https://github.com/user-attachments/assets/ddf0b007-49b3-400d-b6ac-ce5ca39d0ab6)

**Solution proposed:**
Please check below PR for planned fix.
https://github.com/eclipse-platform/eclipse.platform.ui/pull/810
JFace Viewr now has a viewer limit feature which allows it to display only limited number of items and then a special Item called ExpandNode. On Selection of ExpandNode then it populates the next 'limit' number of elements. This phase by phase population of elements keeps the UI responsive.
This limit can be configured in General Preference page. And you can set to any limit based on the computer capacity.

![Image](https://github.com/user-attachments/assets/bcf0c3ec-1205-4b7d-bd25-ba04fa524cf9)

PR #1181, opened 2025-06-04
Workbench viewer limit is setup for 'C/C++ Projects' view.

raghucssit:
C/C++ Projects view makes eclipse unresponsive if it has more than 70k files in a directory inside a cpp project. So we can set up viewr limit for the JFace viewer.

see https://github.com/eclipse-cdt/cdt/issues/1180

### DIFF
--- a/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/internal/ui/cview/CView.java
+++ b/core/org.eclipse.cdt.ui/src/org/eclipse/cdt/internal/ui/cview/CView.java
@@ -111,6 +111,7 @@ import org.eclipse.ui.part.PluginTransfer;
 import org.eclipse.ui.part.ResourceTransfer;
 import org.eclipse.ui.part.ShowInContext;
 import org.eclipse.ui.part.ViewPart;
+import org.eclipse.ui.views.WorkbenchViewerSetup;
 import org.eclipse.ui.views.framelist.FrameList;
 
 /**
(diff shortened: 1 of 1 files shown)


raghucssit:
@jonahgraham Please check this fix.. This is similar to 'Package Explorer'. 'Package Explorer' has this viewer setup done from eclipse 4.30.. Working good so far. I assume it is a safe fix.

jonahgraham:
@raghucssit there is a warning on the [new method call](https://github.com/eclipse-platform/eclipse.platform.ui/blob/6a56d3924b3d6a1ea33d43bcc0168d0a5ea5fa51/bundles/org.eclipse.ui.workbench/eclipseui/org/eclipse/ui/views/WorkbenchViewerSetup.java#L84-L87) - quoted here:

```
	 * <strong>EXPERIMENTAL</strong>. This class or interface has been added as part
	 * of a work in progress. There is no guarantee that this API will work or that
	 * it will remain the same. Please do not use this API without consulting with
	 * the API development team.
	 * </p>
	 *
```

iloveeclipse:
> This change looks good to me, and the fact the API is experimentally marked isn't a big deal, we just need to reach out to Platform UI folk to make sure they know we are using it.

@jonahgraham : Raghu and me were the original Platform UI team developed that code :)

And yes, good point, we should remove the warning, it is now not experimental anymore. We simply forget about that...

@raghucssit: please create PR that removes it.

jonahgraham (approved):

iloveeclipse:
> please create PR that removes it.

Fixed via https://github.com/eclipse-platform/eclipse.platform.ui/pull/3026.

@jonahgraham : we are ready to merge now?

jonahgraham:
Yes, I rebased it to pick up #1185 and it can be merged if the code cleanliness passes.

jonahgraham:
Thanks @raghucssit for the fix

Issue #1186, opened 2025-06-05
Clean up LLDB-MI final launch sequence


jld01:
We should eliminate GDB-specific MI commands from the final launch sequence for LLDB-MI as these will always fail. We should also report the underlying LLDB version to the MI trace for diagnostic purposes.

PR #1187, opened 2025-06-06
Report LLDB version and clean up LLDB-MI launch sequence

jld01:
We use CLI 'version' to replace the MI '-gdb-version' command.

Relates to: #1186


ruspl-afed:
AH, I missed the moment when it became "ready for review"

jld01:
Hi @ruspl-afed, you wrote:
>     1. do we continue to introduce new non-final public types (and what for)?

I am following the existing DSF conventions for reasons of consistency here. If you think we should adopt a more restrictive policy, I suggest raising a new discussion topic.

>     2. does this API change fully comply with the https://github.com/eclipse-cdt/cdt/blob/main/POLICY.md#api ?

I think it does, yes.

jld01:
Yes. In practise, it is not particularly unusual for the likelihood of name coincidence to be considered sufficiently low that the change is treated as non-breaking.

PR #1190, opened 2025-06-11
Document the current steps for releasing CDT

jonahgraham:
This document is to capture all the steps that I do as part of the releasing/maintenance/processing of CDT releng. As I have prepared the CDT 12.1 release I have written down every step I have taken. Once I have completed CDT 12.1 release I will merge this into the repo so this knowledge is properly captured.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,105 @@
+This is the Release plan and TODO list for CDT.
+
+## Steps for Release
+
+### Items at the beginning of development
+
+Prepare main branch for next release
+
+- Create a GitHub issue with the `endgame` label to track work towards the release
+- Update the target platform in cdt.target
+    - Update the Eclipse platform to the latest I-build if possible, it may be best to stay on last release of platform until platform's i-builds are stable e.g. https://download.eclipse.org/eclipse/updates/4.37-I-builds/ - later in the development cycle this can be locked down to milestone versions for stability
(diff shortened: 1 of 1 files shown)


jonahgraham:
Main things left to do:

- N&N handling
- GitHub Milestones
- PMI
- Check an old endgame to make sure everything is captured properly.

jonahgraham:
Another new item for this document:

- If core native parts change, we need to release *before* Eclipse Platform because Eclipse Platform depends on us. See https://github.com/eclipse-platform/eclipse.platform.releng.aggregator/pull/3046#issuecomment-2970743413

RELEASING.md
```diff
@@ -0,0 +1,53 @@
+This is the Release plan and TODO list for CDT.
+
+## Steps for Release
+
+Items at the beginning of development
+
+
+Prepare main branch for next release
```
jld01, line 8: Should we add creation of the new N&N page to this list? It seems useful to create it early so that we can add to it throughout the release cycle.

ruspl-afed:
What do you think about switching p2 update site creation to JustJ technology (used by [EMF](https://download.eclipse.org/modeling/emf/emf/builds/), [Mylyn](https://download.eclipse.org/mylyn/updates/) and many others)? In my experience, this could greatly simplify the p2 delivery process. 

However, I don't know how good it is for 
1) service releases from branches
2) products in executable form

May be @merks can advice regarding these two use cases

merks:
If one plans to produces service releases and non-service releases at the same time then this would be a problem:

image

A bunch of builds maintain both an update site for a product and an update site for just the tools that can be installed into any product, e.g., 

- https://ci.eclipse.org/cbi/view/p2RepoRelated/job/p2repo-sbom/

One can also produce downloads, including downloads of products:

https://download.eclipse.org/birt/updates/release/4.20.0/index.html

I think dealing with service releases at the same time as regular release would be the worst hurdle.  Normally one just increases versions, not matter which segment and that produces a new set of milestones one of which is eventually promoted as the release...

Issue #1204, opened 2025-06-15
NullPointerException reading Mach-O load commands


jld01:
A `NullPointerException` is triggered when presenting source files relating to a MachO64 executable file in the Project Explorer:
```
java.lang.NullPointerException: Cannot read field "cmd" because "loadcommand" is null
	at org.eclipse.cdt.utils.macho.MachO64.createStabsReader(MachO64.java:1799)
	at org.eclipse.cdt.utils.macho.MachO64.getSymbolReader(MachO64.java:1820)
	at org.eclipse.cdt.utils.macho.parser.MachOBinaryObject64.getAdapter(MachOBinaryObject64.java:415)
	at org.eclipse.cdt.internal.core.model.Binary.addSourceFiles(Binary.java:328)
	at org.eclipse.cdt.internal.core.model.Binary.computeChildren(Binary.java:302)
	at org.eclipse.cdt.internal.core.model.Binary.buildStructure(Binary.java:288)
	at org.eclipse.cdt.internal.core.model.Openable.generateInfos(Openable.java:276)
	at org.eclipse.cdt.internal.core.model.CElement.openWhenClosed(CElement.java:428)
```
The executable contains Mach-O load commands generated by the clang toolchain:
```
% clang --version
Apple clang version 17.0.0 (clang-1700.0.13.3)
Target: arm64-apple-darwin24.3.0
```
The issue is caused by the presence of load commands that are not recognised by the Mach-O symbol table reader.

jld01:
Mach-O executables on Apple silicon (AArch64) are now parsed without errors.

PR #1205, opened 2025-06-15
Extend Mach-O load command support

jld01:
We skip load commands that are not processed by the symbol loader and log commands that are not recognized.

Relates to: #1204

### DIFF
--- a/core/org.eclipse.cdt.core/.settings/.api_filters
+++ b/core/org.eclipse.cdt.core/.settings/.api_filters
@@ -26,4 +26,68 @@
             </message_arguments>
         </filter>
     </resource>
+    <resource path="utils/org/eclipse/cdt/utils/macho/MachO64.java" type="org.eclipse.cdt.utils.macho.MachO64$LoadCommand">
+        <filter id="336658481">
+            <message_arguments>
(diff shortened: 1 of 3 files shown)


jonahgraham:
> Are you aware of any issues with the license checker @jonahgraham?

Not before you asked, but the problem isn't really a CDT problem as this is the problematic bundle which we have been using for a while successfully. Sometimes the metadata about bundles fluctuates, so I would merge as is and then if it hasn't settled on its own in a day or two we can put in the license review request so that the metadata resolves itself.

[INFO] maven/mavencentral/com.fasterxml.jackson.core/jackson-core/2.19.0

Note that license tests passed fine on https://github.com/eclipse-cdt/cdt/actions/runs/15619654328 after you merged https://github.com/eclipse-cdt/cdt/pull/1199 and we haven't changed anything.

jld01:
@jonahgraham wrote:
> Sometimes the metadata about bundles fluctuates, so I would merge as is and then if it hasn't settled on its own in a day or two we can put in the license review request so that the metadata resolves itself.

Thanks for the prompt response. I'll try re-running the job tomorrow and then merge regardless, as you suggest.

Issue #1202, opened 2025-06-14
Migrate GitShowInContextHandler behaviour to Egit?


laeubi:
Currently the Terminal view has a dependency to Egit because of `GitShowInContextHandler` accesses the HistoryPageInput to get a list of files/path.

It would be better to implement that in Egit directly e.g. using the Adapter pattern. A `HistoryPageInput` then might be made adaptable to either a `ShowInContext` itself or an `ISelection` ... @jonahgraham can you tell what is the use-case for this? It seems a quite special crafted thing when the selection itself is empty.

jonahgraham:
> Using an adapter is the correct thing, but I think what is needed (if not already present) is an adapter from HistoryPageInput -> java.io.File or org.eclipse.core.runtime.IPath or org.eclipse.core.resources.IResource

☝ was done already for the use case originally resolved with this egit code here. It was done in https://github.com/eclipse-egit/egit/commit/316d0971adb9d9a14ea2a3f31bec41fd1b877d9e - so I think we can simply remove the egit dependency from CDT/Terminal entirely

> Can someone give a hint how to test the code? I know there is show in in the Project Explorer but how does it relates to the terminal view?

In the git perspective right click and select as follows (this seems to be what didn't work in [original bug report](https://bugs.eclipse.org/bugs/show_bug.cgi?id=460277)):

![Image](https://github.com/user-attachments/assets/38d5fc15-46c8-4d4c-bb00-9e021f8ae26e)

Also, if you show history of a file, you can click on that file and do Show in too, the input in that case is mentioned `HistoryPageInput`:

![Image](https://github.com/user-attachments/assets/0040c16c-37bc-499b-8ea7-6d252e9d078d)

jonahgraham:
> remove the egit dependency from CDT/Terminal entirely

PR to do this: https://github.com/eclipse-cdt/cdt/pull/1208

PR #1208, opened 2025-06-17
Remove workaround for egit not providing path information for show in terminal

jonahgraham:
To fix https://bugs.eclipse.org/bugs/show_bug.cgi?id=460277 many years ago some workarounds were added to the terminal to allow adapting git inputs to locations to open a terminal with.

Since that workaround was added it is no longer necessary because egit now adapts objects and provides selections in a way that means we don't need it anymore. In particular https://github.com/eclipse-egit/egit/commit/316d0971adb9d9a14ea2a3f31bec41fd1b877d9e adapts inputs to Files that we know how to open with.

Fixes https://github.com/eclipse-cdt/cdt/issues/1202


jonahgraham:
More info on license check fail https://github.com/eclipse-cdt/cdt/pull/1205#issuecomment-2977761562

PR #1213, opened 2025-06-19
Only import org.eclipse.tm.terminal.connector.local

laeubi:
FYI @jonahgraham 

See
- https://github.com/eclipse-platform/eclipse.platform/pull/1973


laeubi:
@jonahgraham now there is [a feature on platform](https://github.com/eclipse-platform/eclipse.platform/tree/master/terminal/features/org.eclipse.platform.terminal.feature) where I added all migrated terminal stuff I wonder if it would be an option to even **delete** the (then empty) features from CDT and instead use the platform one? As we likely want to adjust the EPPs as well (even though I have not found where the terminal stuff is included there) its seems a good opportunity to get rid of them with the next release.

PR #1219, opened 2025-06-25
FAQ/README.md: Update kernel indexing instructions and cleanup

aduskett:
The last time the instructions for indexing the kernel were updated was before the migration from the wiki with commit 351be05b. In fact, even the example project name is "5.1!"  Since then, several dialog options have changed.

- Update the instructions to reflect current menu and dialog options.
- Update the example project name from 5.1 to 6.12
- Re-arrange the ordering of the instructions so the user isn't bouncing around the dialog tree unnecessarily.


jld01:
Thank you for the PR, @aduskett. Can you re-base this change on the latest CDT sources in _main_ branch please? This will allow our automated checks to complete successfully.

Issue #1220, opened 2025-06-26
Disable Toggle Instruction Step Mode command using activity support.


raghucssit:
Eclipse platform has a limitation:
If a property tester is not loaded then it shows the contribution item by default. For example Toggle Instruction Step Mode command is contributed at Main Menu -> Run menu and Main toolbar. When we start the eclipse if the CDT debug UI plugin is not loaded then this command is visible by default.

This unexpected show up of the menu item has made some eclipse users uncomfortable.

We had tried to fix in other way in the past but they were not technically/functionally possible due to legacy reasons.

How we got this idea ?
CDT already has similar Activity support for `Terminal` command. Thanks @iloveeclipse for the idea.
So I have decided to implement fix for this command, But with controlling activity disablement based on CDT debug ui plugin is loaded or not.

Discussion on limitation of the visibility on the eclipse plaform:
https://github.com/eclipse-platform/eclipse.platform.ui/issues/2900

Other approach to completely remove this command from Run Menu and Main toolbar.
https://github.com/eclipse-cdt/cdt/pull/1143

PR #1221, opened 2025-06-26
Disable Toggle Instruction Step Mode command using activity support.

raghucssit:
TISM command is always visible if the CDT Debug UI plug-in is not loaded yet. This happened due to platform limitation.So handle visibility using activity support.

see https://github.com/eclipse-cdt/cdt/issues/1220

### DIFF
--- a/debug/org.eclipse.cdt.debug.ui/plugin.properties
+++ b/debug/org.eclipse.cdt.debug.ui/plugin.properties
@@ -282,3 +282,10 @@ popup.stepIntoSelection.name=Step Into Selection
 DebuggerConsoleView.name=Debugger Console
 
 GenericTarget.name=Generic Target
+
+# Toggle Instruction Step Mode Command Activity support
+org.eclipse.cdt.debug.ui.category.name=C/C++ Debug UI
+org.eclipse.cdt.debug.ui.category.description=Controls visibility of C/C++ Debug UI contributions
+
+org.eclipse.cdt.debug.ui.activity.toggleInstructionStepMode.name=Instruction Step Mode Button on main Toolbar
+org.eclipse.cdt.debug.ui.activity.toggleInstructionStepMode.description=Control visibility of 'Toggle Instruction Step Mode' button on the main Toolbar
--- a/debug/org.eclipse.cdt.debug.ui/plugin.xml
+++ b/debug/org.eclipse.cdt.debug.ui/plugin.xml
@@ -1302,6 +1302,7 @@
              disabledIcon="icons/dlcl16/instr_step.gif"
              helpContextId="toggle_instruction_step_mode_action_context"
              icon="icons/elcl16/instr_step.gif"
+             id="main.run.menu:org.eclipse.cdt.debug.ui.ToggleInstructionStepModeCommand"
              label="%ToggleInstructionStepModeAction.label"
              style="toggle"
              tooltip="%ToggleInstructionStepModeAction.tooltip">
@@ -1321,6 +1322,7 @@
              disabledIcon="icons/dlcl16/instr_step.gif"
              helpContextId="toggle_instruction_step_mode_action_context"
              icon="icons/elcl16/instr_step.gif"
+             id="main.toolbar:org.eclipse.cdt.debug.ui.ToggleInstructionStepModeCommand"
              label="%ToggleInstructionStepModeAction.label"
              style="toggle"
              tooltip="%ToggleInstructionStepModeAction.tooltip">
@@ -2675,5 +2677,35 @@
            name="%GenericTarget.name">
      </wizard2>
   </extension>
+  
+  <!-- Activity definition that controls the visibility of  Toggle Instruction Step Mode command-->
+  <extension point="org.eclipse.ui.activities">
+   <category
+         id="org.eclipse.cdt.debug.ui.category"
+         name="%org.eclipse.cdt.debug.ui.category.name"
+         description="%org.eclipse.cdt.debug.ui.category.description">
+   </category>
+
+   <activity
+         id="org.eclipse.cdt.debug.ui.activity.toggleInstructionStepMode"
+         name="%org.eclipse.cdt.debug.ui.activity.toggleInstructionStepMode.name"
+         description="%org.eclipse.cdt.debug.ui.activity.toggleInstructionStepMode.description">
+   </activity>
+
+   <activityPatternBinding
+         activityId="org.eclipse.cdt.debug.ui.activity.toggleInstructionStepMode"
+         pattern="org\.eclipse\.cdt\.debug\.ui/main\.toolbar:org\.eclipse\.cdt\.debug\.ui\.ToggleInstructionStepModeCommand">
+   </activityPatternBinding>
+   <activityPatternBinding
+         activityId="org.eclipse.cdt.debug.ui.activity.toggleInstructionStepMode"
+         pattern="org\.eclipse\.cdt\.debug\.ui/main\.run\.menu:org\.eclipse\.cdt\.debug\.ui\.ToggleInstructionStepModeCommand">
+   </activityPatternBinding>
+
+   <categoryActivityBinding
+         activityId="org.eclipse.cdt.debug.ui.activity.toggleInstructionStepMode"
+         categoryId="org.eclipse.cdt.debug.ui.category">
+   </categoryActivityBinding>
+</extension>
+  
     
 </plugin>


debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/ui/CDebugUIPlugin.java
```diff
@@ -323,6 +332,26 @@ public IStatus runInUIThread(IProgressMonitor monitor) {
 			}
 		};
 		wjob.schedule();
+
+		// We need to enable Toggle Instruction Step Mode only if Debug UI plug-in is loaded.
+		setActivityEnabled(TISM_ACTIVITY_ID, true);
```
iloveeclipse, line 337: Please move this code to the job above

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/ui/CDebugUIPlugin.java
```diff
@@ -319,10 +328,29 @@ public void start(BundleContext context) throws Exception {
...
 				startupInUIThread();
+				// We need to enable Toggle Instruction Step Mode only if Debug UI plug-in is loaded.
+				setActivityEnabled(TISM_ACTIVITY_ID, true);
 				return Status.OK_STATUS;
 			}
 		};
 		wjob.schedule();
+
+		// Disable the activity when the platform shuts down.
+		IWorkbench workbench = PlatformUI.getWorkbench();
```
iloveeclipse, line 339: Raghu, I meant **all**new code should go in the job.
The point is, the start() method of activator should do as less as possible for multiple reasons.

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/ui/CDebugUIPlugin.java
```diff
@@ -436,4 +467,26 @@ public static <T> T getService(Class<T> service) {
 		return ref != null ? context.getService(ref) : null;
 	}
 
+	private void setActivityEnabled(String activityId, boolean enabled) {
+		IWorkbench workbench = PlatformUI.getWorkbench();
+		if (workbench == null) {
```
iloveeclipse, line 472: Can it be null here? If yes, the other code doesn't check for it. 
What you probably want is `PlatformUI.isWorkbenchRunning()`.

raghucssit: Yes.. If workbench is null PlatformUI itself throws exception..

debug/org.eclipse.cdt.debug.ui/src/org/eclipse/cdt/debug/ui/CDebugUIPlugin.java
```diff
@@ -435,5 +435,4 @@ public static <T> T getService(Class<T> service) {
 		ServiceReference<T> ref = context.getServiceReference(service);
 		return ref != null ? context.getService(ref) : null;
 	}
-
```
iloveeclipse, line 438: Please undo unrelated change

raghucssit:
We can already define an id for a contribution item. And that can be used for activity support.
I have updated PR to remove all the code needed to activate or deactivate in the Bundle Activator.

debug/org.eclipse.cdt.debug.ui/plugin.properties
```diff
@@ -282,3 +282,10 @@ popup.stepIntoSelection.name=Step Into Selection
 DebuggerConsoleView.name=Debugger Console
 
 GenericTarget.name=Generic Target
+
+# Toggle Instruction Step Mode Command Activity support
+org.eclipse.cdt.debug.ui.category.name=C/C++ Debug UI
+org.eclipse.cdt.debug.ui.category.description=Controls visibility of C/C++ Debug UI contributions
+
+org.eclipse.cdt.debug.ui.activity.toggleInstructionStepMode.name=Toggle Instruction Step Mode Command
```
iloveeclipse, line 290: I believe it should be short and look like `Instruction Step Mode Button on main Toolbar`

debug/org.eclipse.cdt.debug.ui/plugin.properties
```diff
@@ -282,3 +282,10 @@ popup.stepIntoSelection.name=Step Into Selection
...
+org.eclipse.cdt.debug.ui.activity.toggleInstructionStepMode.description=Activity to control visibility of Toggle Instruction Step Mode command
```
iloveeclipse, line 291: Probably better: `Control visibility of 'Toggle Instruction Step Mode'  button on the main toolbar`

debug/org.eclipse.cdt.debug.ui/plugin.xml
```diff
@@ -2675,5 +2677,35 @@
            name="%GenericTarget.name">
      </wizard2>
   </extension>
+  
+  <!-- Activity definition that controls the visibility of  Toggle Instruction Step Mode command-->
+  <extension point="org.eclipse.ui.activities">
+   <category
+         id="org.eclipse.cdt.debug.ui.category"
+         name="%org.eclipse.cdt.debug.ui.category.name"
+         description="%Controls visibility of C/C++ Debug UI contributions">
```
iloveeclipse, line 2686: Should be `%org.eclipse.cdt.debug.ui.category.description`

iloveeclipse (changes requested):
Please after changes attach a screenshot of the Capabilities page where the new activity category & sub items are shown.

debug/org.eclipse.cdt.debug.ui/plugin.xml
```diff
@@ -1302,6 +1302,7 @@
              disabledIcon="icons/dlcl16/instr_step.gif"
              helpContextId="toggle_instruction_step_mode_action_context"
              icon="icons/elcl16/instr_step.gif"
+             id="menu:org.eclipse.ui.run.ToggleInstructionStepModeCommand"
```
ruspl-afed, line 1305: > `org.eclipse.ui.run`

What is the reason to use the prefix from the hosting menu for the local contribution id?

ruspl-afed: > It is the FQN of the contribution. i.e. location+command

I'm not sure that I fully follow your logic here. Target location is already specified with `locationURI` attribute. Why should we prefix an id of CDT contribution with a prefix from Platform is not clear for me. Following this logic vendors with create ids for their contributions started from `org.eclipse` instead of `com.vendor` that will look weird.

raghucssit: > Target location is already specified with `locationURI` attribute.

True. But `id` field was optional and never used. I agree with you we should use vendor prefix like `org.eclipse.cdt.**` but my logic is we should be able to identify this contribution only by  it's id. That is it's location it may be anywhere on the platform may be `main menu/main toolbar/view toolbar `etc and then actual `command id`(I can put command id FQN here but it becomes very long).

It is like `Main Menu-->Run-->ToggleInstructionStepModeCommand`. This is the logic behind that id.
Also it makes easy if we see the `patternBinding` what actually is enabled/disabled.
`pattern="org.eclipse.cdt.debug.ui/toolbar:org.eclipse.debug.ui.main.toolbar.ToggleInstructionStepModeCommand"`

If I use `org.eclipse.cdt.debug.ui.ToggleInstructionStepModeCommand` this is again same thing used for command. Or `otherwise org.eclipse.cdt.debug.ui.menu.run.ToggleInstructionStepModeCommand` something like this. These two hides the location where it is contributed. Yes we can get location from `locationURI` but my idea is to make the id itself self explanatory of the location.

iloveeclipse: @raghucssit :  I believe what Alexander meant was: let use
`id="menu:org.eclipse.cdt.debug.ui.ToggleInstructionStepModeCommand"` instead.

Same for toolbar:
`id="toolbar:org.eclipse.cdt.debug.ui.ToggleInstructionStepModeCommand"`

raghucssit: Fixed.

ruspl-afed: Although I never thought about notation like `<location>:<fqn_id>` it looks reasonable and solves the initial problem.
Thank you for your contribution @raghucssit 

Do you think we can generalize such a way of defining contribution id to a kind of best practice and support it with PDE @iloveeclipse ?

PR #1223, opened 2025-06-27
Detect Homebrew installations on macOS hosts

jld01:
We provide immediate access to tools installed using the Homebrew package manager as for the MSYS2 package manager.

Relates to: #1175


jld01:
This change allows CDT to access build/debug tools provided by the Homebrew package manager for improved usability on macOS hosts. This works in a similar manner to the way we pick up tools provided by the MSYS2 package manager on Windows hosts.

The MBS toolchain descriptions for _MacOSX GCC_ and _LLVM with Clang (MacOSX)_ are modified to use a new macOS configuration environment variable supplier. I have kept the name of this supplier generic to accommodate the possibility of supporting other macOS package managers in the future.

PR #1233, opened 2025-07-04
Move to migrated platform terminal

laeubi:
FYI @jonahgraham


laeubi:
@ruspl-afed as discussed in the IDE call yesterday, this is the PR for adjustments in CDT to migrate to the new terminal view, would be great if you can help with this as well.

ruspl-afed:
> Software being installed: org.eclipse.cdt.debug.application.product 12.2.0.qualifier

Please have a look [here ](https://github.com/eclipse-cdt/cdt/blob/main/debug/org.eclipse.cdt.debug.application.product/debug.product#L231)

jld01:
> The feature is still there! So this should not be the problem, it currently fails for some org.bouncycastle stuff, maybe I should try first a PR that uses ibuilds repo to not change too much things at once....

The bouncycastle version issue seems unrelated to this PR. I have been experimenting on #1241. It would be preferable to use a separate PR for releng changes.

ruspl-afed:
@laeubi I plan to continue looking at your changes today, please let me know about your plans.
I also agree that we should split PRs

laeubi:
I think 
- https://github.com/eclipse-cdt/cdt/pull/1241

needs investigation first and if that works I'll rebase my PR to see what else fails (maybe).

ruspl-afed:
> needs investigation first

looking into it. may be just an infra failure

ruspl-afed:
> restarted main `build` action as well to see what is wrong

I think `org.eclipse.cdt.ui.tests.buildconsole.BuildConsoleTests` is not happy @laeubi

laeubi:
See here: https://github.com/eclipse-cdt/cdt/pull/1241 the test seem to timeout currently :-\

jld01:
@laeubi it looks like https://github.com/eclipse-cdt/cdt/blob/main/core/org.eclipse.cdt.ui.tests/src/org/eclipse/cdt/ui/testplugin/EditorTestHelper.java#L376 was broken by the recent platform UI change: https://github.com/eclipse-platform/eclipse.platform.ui/commit/0e4942bca8aad73999ea2a2ea49e8a8e6aef2f38

laeubi:
Renaming a private (!) field of an internal class can merely be seen as breaking change .. I'm not sure why CDT do the same bad hacks as Platform did, but it has to do the same adaption than here then:

https://github.com/eclipse-platform/eclipse.platform.ui/commit/0e4942bca8aad73999ea2a2ea49e8a8e6aef2f38#diff-54c81a16536eba9489797a9d665c195f0908163e2c7b14ad055d78cc62579fdaL193

ruspl-afed:
@laeubi now CDT uses the later Platform I build to simplify Terminal-related changes
Please let me know regarding your further plans

laeubi:
Some more feature cleanups ...

laeubi:
Yes no problem, now the build actually starts and resolves it gives better errors to work on... code cleaniness check seem needs adjustments at least I see some exceptional handling from @jonahgraham  last time something was moved to platform.

ruspl-afed:
for code cleanliness failure the message is
```
Missing about.ini entry in org.eclipse.tm.terminal.connector.ssh.feature/build.properties
```

ruspl-afed:
I'm thinking how we can converge on this. May be we can split this PR and first transfer CDT-DSF to a new Platform Terminal?

laeubi:
> major version bumps, because it raises a lot of further questions

 I think that is the cleanest solution here, as it depends on new API that is leaked into the bundles own API. An alternative would be the rename / reversion of the whole thing. In the end I think no one should actually API wise depend on a connector anyways, but it is API at the moment sadly.

> removal of anything, because CDT codebase is very conservative

If one depends on anything from the old world, it is still there in the old releases an/or branches, it does not seem useful to retain the sources here if they have evolved/moved to another project I think. Also I think usual users/extenders of CDT should never notice that particular thing:

- Everything from CDT is only referenced internally and do not surface to API
- The Connectors are actually extensions to the terminal view and nothing tied to CDT itself

laeubi:
I now squashed everything, so this is ready for review, will try to extract some aspects into separate PRs now.

laeubi:
GDB Migration:
- https://github.com/eclipse-cdt/cdt/pull/1246

laeubi:
remote migration:

- https://github.com/eclipse-cdt/cdt/pull/1247

laeubi:
Migration of remaining connectors to new API including major version bump:

- https://github.com/eclipse-cdt/cdt/pull/1248

ruspl-afed:
Thank you for supporting this PR in actual state @laeubi !
I suggest to remove added `.project` manifests from it to make it even more clear

jonahgraham:
@ruspl-afed are you able to merge this? You have done the work to review so far.

thanks @laeubi for this PR.

laeubi:
@jonahgraham I think I should rebase the PR and have already creates smaller chunks that should be merged first:

- https://github.com/eclipse-cdt/cdt/pull/1248
- https://github.com/eclipse-cdt/cdt/pull/1249

so I will convert this to draft, @ruspl-afed has had some concerns but these are maybe better discussed on the smaller PRs.

PR #1246, opened 2025-07-19
Migrate GDB to new Eclipse Terminal View Support

laeubi:


ruspl-afed (approved):
LGTM

laeubi:
> What do you think about sequenceNumber attribute in .target file from PDE and Tycho perspective? Is it still actual?

It has never worked reliable and PDE does ignore it since a while altogether, Tycho never has used it at all.

PR #1248, opened 2025-07-19
Migrate CDT provided connectors to the new API

laeubi:
Due to the use of new API this requires a major version bump.

FYI @ruspl-afed


ruspl-afed:
I'm not sure how to deal with it correctly.

1) if we have bundle version major bump, what should it mean for its "enclosing" feature `org.eclipse.tm.terminal.connector.cdtserial.feature`?

2) it seems that main CDT feature is not affected by this change, but may be EPP can be affected especially if we bump feature as well

laeubi:
@jonahgraham we would need an decision on the points mentioned by @ruspl-afed !

laeubi:
Yes please merge and I'll proceed with more smaller PRs then and rebase

PR #1249, opened 2025-07-19
Use migrated bundles in existing features and cleanup

laeubi:
This now uses the new names in the existing features for those connectors migrated to platform.


terminal/features/org.eclipse.tm.terminal.connector.ssh.feature/about.ini
ruspl-afed, line 1: Copyright header missed

terminal/features/org.eclipse.tm.terminal.feature/about.mappings
ruspl-afed, line 1: Copyright header missed

terminal/features/org.eclipse.tm.terminal.view.feature/about.ini
ruspl-afed, line 1: Copyright header missed

terminal/features/org.eclipse.tm.terminal.view.feature/about.mappings
ruspl-afed, line 1: Copyright header missed

ruspl-afed (changes requested):
For several files copyright header is absent, this caused -1

In general, I missed the moment when branding info migrated to feature itself. What are the Tycho\Platform requirements to start using this improvement?

I can see `about.html` added to only one feature but not to its `build.properties`, please explain why so.

laeubi:
I just [copied this from existing features in this repo](https://github.com/eclipse-cdt/cdt/blob/main/terminal/features/org.eclipse.tm.terminal.connector.cdtserial.feature/about.ini) to make the check happy. If or not a "copyright" on an ini file is anything relevant... I don't know :-D 
In any case I did not created something new and can add whatever one thinks is useful.

terminal/features/org.eclipse.tm.terminal.connector.ssh.feature/about.mappings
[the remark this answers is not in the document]
jonahgraham: about.mappings doesn't have copyright heeaders in all the other ones in the repo. Perhaps split this out to a new commit later.

ruspl-afed:
Thank you for your comments @jonahgraham !

@laeubi I hope with the change to `check_features` suggested above we can avoid all the branding-related changes.
If you can update this PR accordingly that would be great.

ruspl-afed:
> the features will have no branding at all 

Even if so, I'll better fix it with further PRs, since we have other troubles with license headers.

I believe that feature branding is defined by referenced branding bundle, via `plugin` attribute of feature element. And it will be easier to reference/add them after finalizing this migration work.

laeubi:
I think I should now better do it one by one for each feature, so I started here (with a hopefully easy one):

- https://github.com/eclipse-cdt/cdt/pull/1253

terminal/features/org.eclipse.tm.terminal.connector.ssh.feature/about.mappings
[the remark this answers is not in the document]
ruspl-afed: ack, let's fix it all after completion of the terminal story

laeubi:
The last batch is here:
- https://github.com/eclipse-cdt/cdt/pull/1258

Issue #1245, opened 2025-07-18
'Before you begin' help page is out of date for macOS


jld01:
The CDT "Before you begin" help page needs improving to a align with recent macOS on Apple silicon.

PR #1250, opened 2025-07-21
Update macOS tool installation instructions

jld01:
Relates to: #1245


jld01:
Requires conclusion on #1225 before merge.

PR #1253, opened 2025-07-24
Cleanup the terminal local feature and update bundle names

laeubi:


ruspl-afed (approved):
LGTM

ruspl-afed:
> license-check failed: Received an error response from ClearlyDefined.

We cannot influence this failure and can ignore it, as no external dependencies were introduced.
Merging.
Please continue @laeubi

PR #1259, opened 2025-07-25
Let old terminal view delegate to the new API and add deprecation

laeubi:
This makes the old terminal bundles delegate to the new platform API.

@jonahgraham this is not complete yet, but would it be ok for CDT to host such legacy adapter?


jonahgraham:
> @jonahgraham this is not complete yet, but would it be ok for CDT to host such legacy adapter?

Yes it is ok.

jonahgraham:
If they want to test the use case then yes that would be great. But I think that is on you to decide how much effort you want to support other projects single-sourcing their dependencies on the terminal.

laeubi:
@jonahgraham as we have other consumers as well (mentioned [here](https://github.com/eclipse-simrel/simrel.build/pull/955#issuecomment-3128884498)) I think it will help in the migration.

This would mean that we have to restore also code from the terminal.control here, but this should be a different topic.

So given the build succeeds I would suggest to merge this so that consumers can start testing if they find any issues.

jdneo:
Will this be included in 2025-09?

laeubi:
@jdneo this is currently only a draft (and it seems @jonahgraham has already found some issues). Also as described I'm not completely convinced it is the right aproach here.

So the main question would be if you are interested in helping improving and testin this aproach to make it more bullet proof or if we more should aim to getting things using the modernized API.

PR #1258, opened 2025-07-25
Cleanup remaining terminal features

laeubi:

PR #1260, opened 2025-07-25
Remove `org.eclipse.terminal.view.core.utils`

ruspl-afed:
Currently `Activator` exposes `TraceHandler` for public usage that looks more like a copy-paste from other similar bundles, where it was already removed.

Major bump is a great opportunity to remove this public method.


ruspl-afed:
FYI @laeubi , this should solve the problem with https://github.com/eclipse-cdt/cdt/pull/1258

PR #1264, opened 2025-07-26
Function resolution: prefer definition over declaration (#1263)

### LINKED ISSUE #1263: A constexpr function definiton in a header is ignored if declaration visible in different file
**Describe the bug**
A `constexpr` function definiton in a header is ignored if a declaration of the same function (without body) is visible in  primary source file.

**To Reproduce**

In `foo.h`:

    constexpr int getConstVal() { return 1; }

In `foo.cc`:

    #include <ectest8.h>
    
    constexpr int getConstVal();
    
    constexpr int v = getConstVal();
    
    template <int V> struct Foo;
    template <> struct Foo<1> {
        constexpr static int val = 55;
    };
    
    int x = Foo<v>::val;

The final line in `foo.cc` will be highlighted as an error ("symbol 'val' could not be resolved"). This is due to a failure to correctly resolve `v` as the constant `1` via the call to `getConstVal()` (which returns `1`).

If the definition is moved from the header to the top of foo.cc (or anywhere else in foo.cc) the error goes away.

**Expected behavior**
No error should be marked. Whether the definition is in a header or in the main source file should make no difference.

**Screenshots**
See source samples above.

 - OS and OS Version/extra details: Linux but N/A
 - Eclipse Version: 2025-06 (4.36)
 - CDT Version:   12.1.0.202506041907

**Additional context**
N/A

davmac314:
For constexpr functions, the definition may be found in a separate header to a declaration. For evaluation purposes the definition is needed; therefore, when resolving a function call, select the definition in favour of the declaration, even if it comes from the index.

Fixes #1263.

### DIFF
--- a/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/internal/index/tests/IndexCPPBindingResolutionBugsTest.java
+++ b/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/internal/index/tests/IndexCPPBindingResolutionBugsTest.java
@@ -1447,4 +1447,18 @@ public abstract class IndexCPPBindingResolutionBugsTest extends IndexBindingReso
 	public void testStructNameIntroducedInNamespace() throws Exception {
 		checkBindings();
 	}
+
+	// constexpr int getNum(int) { return 3; }
+	// constexpr int getNum(long) { return 6; }
+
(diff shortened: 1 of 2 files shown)


i-garrison:
I confirm this change fixes the issue indicated.

Issue #1267, opened 2025-07-26
Primitive template arguments not evaluated when matching partial specialisation


davmac314:
**Describe the bug**
Non-literal primitive template arguments not evaluated when matching partial specialisation.

**To Reproduce**

    template <typename T> constexpr bool consume_r1() { return true; }
    template <class T1, bool EN = true> class A { };
    template <class T1> class A <T1, (bool)consume_r1<T1>() > {
        constexpr static int val = 99;
    };
    
    A<int> a1; // should use specialisation
    
    int f = a1.val; // error marked (incorrect!)

**Expected behavior**
The line marked with an error should not be marked, it has no error. The issue is that CDT does not use the partial specialisation  of A which defines the `val` member, though it should.

**Version Information (please complete the following information):**
 - OS: Linux (N/A)
 - Eclipse Version: 2025-06 (4.36)
 - CDT Version: 12.1.0.202506041907

PR #1268, opened 2025-07-26
Improve template specialisation selection (#1267)

davmac314:
A numeric or boolean template argument in a partial specialisation needs to be resolved, which isn't currently being handled.


i-garrison:
I confirm this change is valid and fixes the issue indicated.

Issue #1271, opened 2025-07-26
Parser: various mixings of attributes/alignas are not accepted, but should be


davmac314:
**Describe the bug**
Various mixings of `[[attr]]`-style and `__attribute__((attr))` attributes, and the `alignas(n)` specifier, should be accepted by the parser, but are rejected (highlighted with yellow underline and "syntax error" alert).

**To Reproduce**

    int v1 alignas(16); // should be ok but rejected
    alignas(16) int v2; // (ok)
    int v3 [[maybe_unused]] __attribute__((section(".foo"))) [[maybe_unused]] __attribute__((section(".foo"))); // (ok)
    [[maybe_unused]] __attribute__((section(".foo"))) int v5; // (ok)
    __attribute__((section(".foo"))) [[maybe_unused]] int v5; // (ok)
    [[maybe_unused]] __attribute__((section(".foo")))
            [[maybe_unused]] __attribute__((section(".foo"))) int v4; // should be ok but rejected
    alignas(16) __attribute__((section(".foo"))) [[maybe_unused]] int v5; // should be ok but rejected
    alignas(16) [[maybe_unused]] __attribute__((section(".foo"))) int v6; // (ok)

**Expected behavior**
All lines above should be accepted by the parser

**Version Information (please complete the following information):**
 - OS: Linux
 - Eclipse Version: 2025-06 (4.36)
 - CDT Version: 12.1.0.202506041907

PR #1272, opened 2025-07-26
Allow general mixing of attributes and alignas (#1271)

davmac314:
Regular C++ attributes (eg [[maybe_unused]]), older GCC-style attributes (__attribute__((unused))), and alignas() specifications can be intermingled.

Fix the parser to handle several cases that it disallowed previously.


i-garrison:
I confirm this change is valid and fixes the issue indicated.

PR #1273, opened 2025-07-28
Add old org.eclipse.tm.terminal bundles p2 update site

jonahgraham:
See https://github.com/eclipse-simrel/simrel.build/pull/955


jonahgraham:
This is a temporary solution to get M2 out until we can resolve all dependents in a way we are happy.

fyi: @laeubi

PR #1274, opened 2025-07-29
Delete the terminal test module that was migrated to platform

laeubi:
@jonahgraham @ruspl-afed I don't think it is useful here anymore.


ruspl-afed (commented):
Since we are going to return some old bundles for transition period, shouldn't we test them (may be in reduced scope)?

laeubi:
@jonahgraham @ruspl-afed could we then proceed on this?

jonahgraham:
Explicit +1 - the 2025-09 branch is created and the main branch has been updated now that https://github.com/eclipse-cdt/cdt/pull/1297 is merged

PR #1276, opened 2025-08-01
Fix handling of multiple MBS primary input types

jld01:
Required to accommodate MBS tools that process input files of multiple input types. Previously, the internal builder was processing input files of the first declared input type only.


ruspl-afed:
Do we have any example to demonstrate `before` and `after`? 
Although this change looks like improvement it may have an impact on already existing MBS toolchains

jld01:
Hi @ruspl-afed, we don't have any examples of MBS tools that declare multiple input types in the CDT repo, but MBS tool objects already hold an array of primary input types and MBS is clearly intended to support this. Perhaps there are some downstream MBS tools that make use of this feature. I discovered the bug that this PR fixes while working on #1277. Without this fix and when using the internal builder, only those input files described by the first input type that the tool declares are built. This is the only example I can offer at present.

Issue #1290, opened 2025-09-02
multiple MBS primary input types fix may have regressed something


jonahgraham:
@jld01 sorry for not realizing this earlier - but this change may actually caused a regression. I thought when I first saw the failure it was an intermittent failure in the old test suite. But now it looks like this test reliably fails as each build since it was merged has had this test fail.

https://ci.eclipse.org/cdt/job/cdt/job/main/594/testReport/junit/org.eclipse.cdt.managedbuilder.core.tests/BuildDescriptionModelTests/testDesTest30_2_asm_only/

However, I don't know why it fails on Jenkins and not on GitHub. Perhaps it isn't running on GitHub, or the tools available on the PATH on GitHub affect the behaviour.

I am going to pull this comment into a new issue so it can be discussed in a non-closed PR - lets continue discussion there.

_Originally posted by @jonahgraham in https://github.com/eclipse-cdt/cdt/issues/1276#issuecomment-3246292023_

jonahgraham:
> It does fail on GitHub - it just isn't failing the build 😢

This is because https://github.com/actions/runner-images/issues/11101 - PR on the way.

PR #1291, opened 2025-09-02
Update test result reporting to use latest ubuntu

jonahgraham:
While the builds have been running, for a few months the job that collates the results and reports them hasn't run due to Ubuntu 20.04 EOL on GitHub actions (https://github.com/actions/runner-images/issues/11101)

Fixes the lack of reporting part of https://github.com/eclipse-cdt/cdt/issues/1290

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   build:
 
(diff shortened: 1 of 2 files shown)


jonahgraham:
Note that the test report workflow update that is included in this PR is not used to run the PR. The PR triggers the main branch version of this workflow. So the effects of it will only be seen once merged.

PR #1292, opened 2025-09-02
Pin dependencies to released items for 2025-09 RC2

jonahgraham:


jonahgraham:
Hi @ghentschke - it wasn't actually this commit - see history in https://github.com/eclipse-cdt/cdt/issues/1290

PR #1297, opened 2025-09-07
Prepare repo for CDT 12.3.0 development

jld01:


pom.xml
jld01, line 1: Should we also update `<help-docs-eclipserun-repo>` on each release, @jonahgraham?

releng/org.eclipse.cdt.target/cdt-baseline.target
```diff
@@ -51,7 +51,7 @@
       <unit id="org.eclipse.tools.templates.ui" version="0.0.0"/>
     </location>
     <location includeAllPlatforms="true" includeConfigurePhase="true" includeMode="slicer" includeSource="false" type="InstallableUnit">
-      <repository location="https://download.eclipse.org/tools/cdt/releases/cdt-lsp-3.1/cdt-lsp-3.1.0/"/>
+      <repository location="https://download.eclipse.org/tools/cdt/releases/cdt-lsp-3.2/cdt-lsp-3.2.0/"/>
```
jld01, line 54: I can see no cdt-lsp-3.3 RC builds at present. Are we sticking with CDT-LSP 3.2.0 for 2025-09?

jld01:
Test failure is unrelated and a known issue #1290.

releng/org.eclipse.cdt.target/cdt-baseline.target
ruspl-afed: Personally I do not support this cyclic reference, but they say this is to make possible a scenario where someone want to import only one CDT project to workspace and have it compiled. Well, it can work for some cases, but not for everything.

Answering your question I would align the whole target to one SimRel:
1) "latest released" for baseline target
2) "latest" for current target

pom.xml
jonahgraham: I also added it to the TODO list in https://github.com/eclipse-cdt/cdt/pull/1190#issuecomment-3271693815

Issue #1318, opened 2025-09-25
org.eclipse.cdt.core.build.CBuildConfigurationManagerTests.testNoConfigs is a flaky test


ewaterlander:
org.eclipse.cdt.core.build.CBuildConfigurationManagerTests.testNoConfigs failed in #1317

PR #1320, opened 2025-09-27
Allow NaryTypeId in template variable initializer

i-garrison:
Resolving libstdc++ template variables like `std::is_constructible_v<>` which uses Nary typeid `__is_constructible` as initializer value currently fails because `ValueFactory` does not know about `IAstNaryTypeIdExpression`.

Add missing bits + test.

This fixes resolving lots of e.g. current GCC `std::optional<T>.emplace()` that need Nary builtin value for `std::is_constructible<T>` selector of `emplace()` variant.


i-garrison:
test failure `org.eclipse.cdt.core.build.CBuildConfigurationManagerTests.testNoConfigs` is already known #1318

PR #1337, opened 2025-10-03
Move connector.cdtserial and native.serial to new serial subfolder

laeubi:
It seems to be only used there... WDYT @jonahgraham @ruspl-afed ?


jonahgraham:
Debug.product is the CDT standalone debugger - the other main places it is referenced from are:

- https://github.com/eclipse-packaging/packages/blob/4bb50017c3ef23ecf420372b417da81a14563513/packages/org.eclipse.epp.package.embedcpp.product/epp.product#L280
- https://marketplace.eclipse.org/content/eclipse-terminal

Probably best place is native/ because that is where its primary dependency lives. native/ could just as easily be renamed serial/

PR #1358, opened 2025-10-14
Remove most org.eclipse.tm.terminal.* features

jonahgraham:
With the TM Terminal migrated to Eclipse Terminal, remove the fine grained terminal features. The Eclipse Terminal feature will update from all of these features so users will have a continuous experience.

This requires Eclipse Platform change https://github.com/eclipse-platform/eclipse.platform/pull/2199 which adds support for org.eclipse.tm.terminal features updating to org.eclipse.terminal feature. Please see that PR for details of how it behaves and testing done on this PR.


jonahgraham:
> I have a follow up PR that cleans up the remaining terminal/ folder too

See https://github.com/eclipse-cdt/cdt/pull/1360 for that part

laeubi:
Seems reasonable to me, but EPP would need adjustments as they are still using the old ones as far as I know

jonahgraham:
Thanks for having a look, I have did the epp update yesterday in https://github.com/eclipse-packaging/packages/pull/370

PR #1360, opened 2025-10-15
Cleanup the terminal/ folder in CDT's source

jonahgraham:
Now that the terminal has migrated to the Eclipse Platform this commit moves around the remaining parts of the terminal to their correct location in the CDT source tree

- Move terminal connector features next to their defining bundles
- Remove terminal-only profile from CDT's build system.
- Remove terminal/ references in build scripts
- Rename remaining connectors (cdtserial/remote) so their display name reflects new naming scheme.
- Renaming bundles and packages for remaining connectors is excluded, a readme has been added to each explaining name


jonahgraham:
I think this would be the last step in migrating the terminal to Eclipse Platform. Once this, the parent (#1358) and the dependent platform https://github.com/eclipse-platform/eclipse.platform/pull/2199 is merged then the terminal cleanup is done on the CDT project.

PR #1361, opened 2025-10-15
Fix broken references in CDT.setup's working sets

merks:


akurtakov:
Merging on behalf of @merks . Not waiting for the build as it will not validate this change.

PR #1366, opened 2025-10-19
Index distinct typedef in header file

i-garrison:
GDB sources contain private portable regex implementation which provides own regmatch_t typedef. If another project file references system regex.h which provides regmatch_t typedef too, indexer would not adapt that binding and instead reuse first one it encountered.

Later when that other file is parsed, attempt to resolve binding for ASTName referring to regmatch_t would fail because found PDOM binding is reachable only via private implementation header which is not included.

Fix this by adapting unique typedef name to make both typedefs availabe to lookup procedure which will then filter by reachable index fragment.

This change is related to https://bugs.eclipse.org/bugs/show_bug.cgi?id=214146 `Typedef in a .cc file affects other .cc files`

I found very similar issue first in unrelated workspace where one project carry a header file and another one uses that same file via symlinks on the filesystem, resulting in lots of name resolution errors. To me it looks like the strategy of adapting global names for classes and other entities has it's limits but cannot suggest anything better yet.

### DIFF
--- a/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/pdom/dom/PDOMBinding.java
+++ b/core/org.eclipse.cdt.core/parser/org/eclipse/cdt/internal/core/pdom/dom/PDOMBinding.java
@@ -396,6 +396,9 @@ public abstract class PDOMBinding extends PDOMNamedNode implements IPDOMBinding 
 				IString s0 = b0.getDBName(), s1 = b1.getDBName();
 				cmp = s0.compare(s1, true);
 				if (cmp == 0) {
+					cmp = compareSignatures(b0, b1);
+				}
+				if (cmp == 0) {
 					long l1 = b0.getLocalToFileRec();
 					long l2 = b1.getLocalToFileRec();
 					if (l1 != l2) {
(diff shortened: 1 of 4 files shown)


jonahgraham:
Thanks for trying to tackle this @i-garrison - let me know when it is ready for review + merge.

If you need help resolving this, I'm not sure I can help, but maybe some others may have some ideas, so please ask and we'll see what we can do.

i-garrison:
Hi @jonahgraham I'm a bit confused about which of JUnit variants is needed to run unit tests. IIRC a few indexer tests do not work under JUnit5 because single project / multi project variants either are not created or not working because if changes in JUnit5. I see transition to JUnit5 started some time ago, but what is the current state?

jonahgraham:
JUnit5 can be used to *run* the whole suite - but it is a lot of work to update the tests themselves to JUnit5 API. I have updated tests to JUnit5 when I have had other work to do in that area and when I decide the effort required is worth it / the benefit from new JUnit5 features valuable enough.

> IIRC a few indexer tests do not work under JUnit5 because single project / multi project variants

I am not aware of any tests having a problem with running with JUnit5 runner. If there is a specific test(s) that you see an issue with, please let me know and I can look closer.

For clarity I mean running with JUnit5 means the JUnit5 dropdown in the launch configuration. This picture is from [TESTING.md](https://github.com/eclipse-cdt/cdt/blob/main/TESTING.md)




-- 

I'm reviewing the change now.

Issue #1405, opened 2025-12-21
build  cdt source code failed


duanli1985:
I  import  cdt  source code 12.3.0 in  eclipse  for  committers version,
import as Maven project ,it occurs error,it apperars dialog
your ide is missing natures to properly support your projects
some extensions on the eclipse markplace can be installed to support those natures
com.soyatec addtitional  Nature
org.eclipse.cdt.core.ccnature
org.eclipse.cdt.ore.cnature

jonahgraham:
I have this preference off so I didn't get that prompt:

<img width="734" height="402" alt="Image" src="https://github.com/user-attachments/assets/41722467-23a8-4c9f-8528-3e3c43a0073c" />

However, we should make it work by including CDT, and removing the no longer used soyatec reference. PRs incoming.

PR #1414, opened 2026-01-06
Remove out of date UML (files and natures) from LLVM plug-in

jonahgraham:
Part of https://github.com/eclipse-cdt/cdt/issues/1405


jonahgraham:
@jld01 I am pretty sure this is ok, but as you have been in the llvm bundles recently perhaps you know of some use here that I am not aware of.

Issue #602, opened 2023-10-24
__attribute__ in function declarators missing from AST


PhilippWendler:
**Describe the bug**

I am using Eclipse CDT as a library in our project to parse C code and access the AST nodes. I am having trouble accessing some `__attribute__` nodes that are part of declarators, and if I did not overlook anything, it is because these attributes are missing from CDT's AST. If it is my mistake, I would be glad about a pointer about how to access the attributes.

**To Reproduce**

1. Parse the following code with CDT:
```c
extern int ( __attribute__((__artificial__)) func)();
```
2. Let the following `ASTVisitor` run over it:
```java
new ASTVisitor(true) {
  @Override
  public int visit(IASTAttribute pAttribute) {
    System.err.println("IASTAttribute: " + pAttribute.getRawSignature());
    return super.visit(pAttribute);
  }

  @Override
  public int visit(IASTAttributeSpecifier pSpecifier) {
    System.err.println("IASTAttributeSpecifier:" + pSpecifier.getRawSignature());
    return super.visit(pSpecifier);
  }

  @Override
  public int visit(IASTDeclarator pDeclarator) {
    String className = pDeclarator.getClass().getSimpleName();
    System.err.println(className + ": " + pDeclarator.getRawSignature());
    System.err.println(
        className
            + " attribute specifier count: "
            + pDeclarator.getAttributeSpecifiers().length);
    System.err.println(
        className + " attribute count: " + pDeclarator.getAttributes().length);
    return super.visit(pDeclarator);
  }
};
```
3. This produces the following output:
```
CASTFunctionDeclarator: ( __attribute__((__artificial__)) func)()
CASTFunctionDeclarator attribute specifier count: 0
CASTFunctionDeclarator attribute count: 0
CASTDeclarator: __attribute__((__artificial__)) func
CASTDeclarator attribute specifier count: 0
CASTDeclarator attribute count: 0
```
The attributes are visible in the raw signature of the function declarator and the nested declarator, but `getAttributes()` and `getAttributeSpecifiers()` of both declarators return empty arrays. The visitor also never encounters any `IASTAttribute`s or `IASTAttributeSpecifier`.

**Expected behavior**

The nested `CASTDeclarator` should have the `artificial` attribute attached to it. The `ASTVisitor` should encounter this attribute during traversal.

**Version Information (please complete the following information):**
 - OS and OS Version/extra details: Ubuntu 22.04
[...]

PR #1454, opened 2026-05-06
fix: __attribute__ of declarator missing from AST (issues #706 and #602)

### LINKED ISSUE #706: __attribute__ of declarator is missing from AST
**Description**

If a declaration such as 
```
int target;
extern int other, __attribute__((alias("target"))) error_decl;
```
consists of multiple declarators and a declarator has an attribute prepended, this attribute does not show up in the CDT AST.
The attribute does occur in the declarator's `getRawSignature()`, though.

If the attribute is appended as in 
```
int target;
extern int other, decl __attribute__((alias("target")));
```
everything is fine and the attribute shows up correctly in the AST.

**Related bugs**

#602 may be related, but I did not look into the inner workings of CDT.

**Version Information**
- Eclipse 2023-06 (4.28.0)
- CDT 11.2.0

strimo378:
Fixes [eclipse-cdt/cdt#706](https://github.com/eclipse-cdt/cdt/issues/706) and [eclipse-cdt/cdt#602](https://github.com/eclipse-cdt/cdt/issues/602):

## Root Cause

In `GNUCSourceParser.consumePointerOperators()`, `IToken mark = mark()` was called *after* `__attribute_decl_seq()` consumed any attributes. When no pointer operator (`*`) followed the attribute, `backup(mark)` restored only to *after* the attribute — permanently losing it from the token stream. The `declarator()` caller then tried to re-consume the attribute at line 1407 but it was already gone.

## Fix

Move `IToken mark = mark()` to *before* the `__attribute_decl_seq()` call. Now when no `*` follows, `backup(mark)` correctly restores to *before* the attribute, allowing `declarator()` to pick it up via its own `__attribute_decl_seq()` call.

Also update `startOffset` to use `last.getOffset()` (the `*` token's offset) instead of `mark.getOffset()`, since `mark` now points before any attribute rather than at the `*`.

### DIFF
--- a/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/core/parser/tests/ast2/AST2CPPAttributeTests.java
+++ b/core/org.eclipse.cdt.core.tests/parser/org/eclipse/cdt/core/parser/tests/ast2/AST2CPPAttributeTests.java
@@ -642,4 +642,19 @@ public class AST2CPPAttributeTests extends AST2TestBase {
 		assertSame(parent1, parent3);
 		assertSame(parent1, parent4);
 	}
+
+	//	int target;
+	//	extern int other, __attribute__((alias("target"))) error_decl;
+	@Test
+	public void testGCCAttributeBeforeDeclaratorInList_bug706() throws Exception {
+		IASTTranslationUnit tu = parseAndCheckBindings(getAboveComment(), ParserLanguage.C, ScannerKind.GNU);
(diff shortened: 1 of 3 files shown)


strimo378:
@jonahgraham This is my first contribution to CDT. We have been using it as a front-end for our C source-to-source compiler for about 10 years. During that time, we identified 3–4 issues that I plan to address in follow-up PRs. We also implemented several workarounds for missing API features that might be worth supporting directly in CDT (if of interest).

strimo378:
@jonahgraham @jld01 ping

PR #1462, opened 2026-05-27
[releng] Copy build output to final location in promotion script

jonahgraham:
In the past we would download from CI, now we simply copy the latest build of the given branch to download location.

This became necessary because there is no longer anonymous access to CI from download.eclipse.org, so the wget commands no longer work.

The downside of this method is that we can only deploy the latest build on a branch.

Worksaround https://gitlab.eclipse.org/eclipsefdn/helpdesk/-/issues/7520


jonahgraham:
This is not tested by MR flow, so merging now to test on Jenkins CI.

Issue #1455, opened 2026-05-06
[Sticky Scrolling] C/C++ Language Support in CDT


gsmith-dfc:
**Is your feature request related to a problem? Please describe.**
For code written in C (and, presumably, C++), the sticky scrolling feature seems to use the fallback style based purely on indents. This causes problems when the code is formatted in the Allman style, where curly braces are on their own line.

It does grab the `case:` label properly, as shown in the screenshots, but the rest of the sticky lines are just opening curly braces (`{`) with no context (lines 45 & 50 in the screenshot).

<img width="854" height="236" alt="Image" src="https://github.com/user-attachments/assets/bf4f91a0-b68d-4682-8c98-9e07daba8a94" />

<br/>

<img width="634" height="116" alt="Image" src="https://github.com/user-attachments/assets/573181ec-3584-47a0-a33b-4f333a667e8c" />

<br/>

**Describe the solution you'd like**
I would like the sticky scrolling to work like it apparently does in JDT now, where you get the actual useful line of context for the function, control statements, etc. based on syntax and context, rather than just indentation. In this example, we should see lines 44, 46, 48, and 49, showing the function signature, `switch` statement, `case` statement, and `if` statement in the "sticky" area.

I believe this was improved in JDT at [#2149](https://github.com/eclipse-jdt/eclipse.jdt.ui/pull/2149), which could be used as a reference for the CDT implementation since Java syntax is similar to C/C++.

**Describe alternatives you've considered**
Besides the JDT implementation, I do not know of any alternatives for Eclipse CDT. But VSCode also has a similar feature. Sometimes it seems like it works with Allman formatting and sometimes not, so I'm not sure its feature status.

**Additional context**
I had originally raised this issue in the Eclipse Platform UI repo ([#3952](https://github.com/eclipse-platform/eclipse.platform.ui/issues/3952)) referencing the [Sticky Scroll Umbrella (#2152)](https://github.com/eclipse-platform/eclipse.platform.ui/issues/2152) issue. I was referred to create the issue here, since they have now introduced an [enhancement that allows languages to provide their own implementation (#2756)](https://github.com/eclipse-platform/eclipse.platform.ui/pull/2756).

jld01:
@gsmith-dfc this would be a good improvement for CDT and, as you say, not too hard to implement given the JDT example. If someone would like to pick this work up, I can offer some support. Otherwise, I may find time to look at it myself in the next few months.

jld01:
[...]

PR #1467, opened 2026-06-05
Add sticky lines provider for the C/C++ editor

jld01:
Relates to: #1455


jld01:
Hi @jjohnstn, I wonder if you have any time to look over this PR since you authored the equivalent Java extension? I am thinking it would be good to separate the AST analysis from the UI-dependent code to accommodate unit testing but perhaps in a separate PR.

jonahgraham (approved):
We reviewed this in the CDT Call today and it looks great.

PR #1471, opened 2026-06-10
[releng] p2 composites for CDT 12.5 and CDT LSP 3.6

jonahgraham:


jonahgraham:
This can't be merged until https://github.com/eclipse-cdt/cdt/issues/1435#issuecomment-4671373198 is resolved.

PR #1483, opened 2026-06-24
Added API to manually run CMake

DangMinhTam382:
Allows ISV to manually run CMake without having to build the project.
Moved code to separate method and make a convenience protected for the derived to call.

### DIFF
--- a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
+++ b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
@@ -31,6 +31,7 @@ import org.eclipse.cdt.cmake.core.properties.CMakeGenerator;
 import org.eclipse.cdt.cmake.core.properties.CMakePropertiesFactory;
 import org.eclipse.cdt.cmake.core.properties.ICMakeGenerator;
 import org.eclipse.cdt.cmake.core.properties.ICMakeProperties;
+import org.eclipse.cdt.core.CCorePlugin;
 import org.eclipse.cdt.core.CommandLauncherManager;
 import org.eclipse.cdt.core.ConsoleOutputStream;
 import org.eclipse.cdt.core.ErrorParserManager;
@@ -60,8 +61,11 @@ import org.eclipse.core.resources.IResourceDelta;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
+import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.NullProgressMonitor;
 import org.eclipse.core.runtime.Platform;
+import org.eclipse.core.runtime.Status;
+import org.eclipse.core.runtime.SubMonitor;
 import org.eclipse.core.runtime.jobs.Job;
 import org.eclipse.debug.core.ILaunchManager;
 import org.eclipse.launchbar.core.target.ILaunchTarget;
@@ -195,45 +199,10 @@ public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMa
 			}
 			CommandDescriptorBuilder cmdBuilder = new CommandDescriptorBuilder(cmakeProperties);
 			if (runCMake) {
-				CMakeBuildConfiguration.deleteCMakeErrorMarkers(project);
-
-				infoStream.write(String.format(Messages.CMakeBuildConfiguration_Configuring, buildDir));
-				CommandDescriptor command = cmdBuilder
-						.makeCMakeCommandline(toolChainFile != null ? toolChainFile.getPath() : null);
-				// tell cmake where its script is located..
-				IContainer srcFolder = project;
-				command.getArguments().add(new File(srcFolder.getLocationURI()).getAbsolutePath());
-
-				infoStream.write(String.join(" ", command.getArguments()) + '\n'); //$NON-NLS-1$
-
-				org.eclipse.core.runtime.Path workingDir = new org.eclipse.core.runtime.Path(
-						getBuildDirectory().toString());
-				// hook in cmake error parsing
-				try (CMakeErrorParser errorParser = new CMakeErrorParser(new CMakeExecutionMarkerFactory(srcFolder))) {
-					ParsingConsoleOutputStream errStream = new ParsingConsoleOutputStream(console.getErrorStream(),
-							errorParser);
-					IConsole errConsole = new CMakeConsoleWrapper(console, errStream);
-					Process p = startBuildProcess(command.getArguments(), new IEnvironmentVariable[0], workingDir,
-							errConsole, monitor);
-					String arg0 = command.getArguments().get(0);
-					if (p == null) {
-						// process start failed
-						String msg = String.format(Messages.CMakeBuildConfiguration_Failure, ""); //$NON-NLS-1$
-						addMarker(new ProblemMarkerInfo(srcFolder.getProject(), -1, msg,
-								IMarkerGenerator.SEVERITY_ERROR_BUILD, null, new org.eclipse.core.runtime.Path(arg0)));
-						return null;
-					}
-
-					// check cmake exit status
-					final int exitValue = watchProcess(errConsole, monitor);
-					if (exitValue != 0) {
-						// cmake had errors...
-						String msg = String.format(Messages.CMakeBuildConfiguration_ExitFailure, arg0, exitValue);
-						addMarker(srcFolder.getProject(), -1, msg, IMarkerGenerator.SEVERITY_ERROR_BUILD, null);
-						return null;
-					}
+				IStatus result = configureCMakeBuildFiles(cmdBuilder, console, infoStream, monitor);
+				if (!result.isOK()) {
+					return null;
 				}
-				cmakeListsModified = false;
 			}
 
 			// parse compile_commands.json file
@@ -294,6 +263,65 @@ public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMa
 			throw new CoreException(Activator
 					.errorStatus(String.format(Messages.CMakeBuildConfiguration_Building, project.getName()), e));
 		}
+	}
+
+	/**
+	 * @since 2.1
+	 */
+	protected IStatus configureCMakeBuildFiles(IProgressMonitor monitor) throws CoreException, IOException {
+		IProject project = getProject();
+		project.deleteMarkers(ICModelMarker.C_MODEL_PROBLEM_MARKER, false, IResource.DEPTH_INFINITE);
+		// Setup console
+		IConsole console = CCorePlugin.getDefault().getConsole();
+		console.start(project);
+		ICMakeProperties cmakeProperties = getCMakeProperties();
+		CommandDescriptorBuilder cmdBuilder = new CommandDescriptorBuilder(cmakeProperties);
+		return configureCMakeBuildFiles(cmdBuilder, console, console.getInfoStream(), monitor);
+	}
+
+	private IStatus configureCMakeBuildFiles(CommandDescriptorBuilder cmdBuilder, IConsole console,
+			ConsoleOutputStream infoStream, IProgressMonitor monitor) throws CoreException, IOException {
+		SubMonitor subMonitor = SubMonitor.convert(monitor, 4);
+		CMakeBuildConfiguration.deleteCMakeErrorMarkers(getProject());
+		infoStream.write(String.format(Messages.CMakeBuildConfiguration_Configuring, getBuildDirectory()));
+		CommandDescriptor command = cmdBuilder
+				.makeCMakeCommandline(toolChainFile != null ? toolChainFile.getPath() : null);
+		// tell cmake where its script is located..
+		IContainer srcFolder = getProject();
+		command.getArguments().add(new File(srcFolder.getLocationURI()).getAbsolutePath());
+
+		infoStream.write(String.join(" ", command.getArguments()) + '\n'); //$NON-NLS-1$
+
+		org.eclipse.core.runtime.Path workingDir = new org.eclipse.core.runtime.Path(getBuildDirectory().toString());
+		// hook in cmake error parsing
+		try (CMakeErrorParser errorParser = new CMakeErrorParser(new CMakeExecutionMarkerFactory(srcFolder))) {
+			ParsingConsoleOutputStream errStream = new ParsingConsoleOutputStream(console.getErrorStream(),
+					errorParser);
+			IConsole errConsole = new CMakeConsoleWrapper(console, errStream);
+			Process p = startBuildProcess(command.getArguments(), new IEnvironmentVariable[0], workingDir, errConsole,
+					subMonitor.split(1));
+			String arg0 = command.getArguments().get(0);
+			if (p == null) {
+				// process start failed
+				String msg = String.format(Messages.CMakeBuildConfiguration_Failure, ""); //$NON-NLS-1$
+				addMarker(new ProblemMarkerInfo(srcFolder.getProject(), -1, msg, IMarkerGenerator.SEVERITY_ERROR_BUILD,
+						null, new org.eclipse.core.runtime.Path(arg0)));
+				return Status.error(msg);
+			}
+			// check cmake exit status
+			final int exitValue = watchProcess(errConsole, subMonitor.split(1));
+			if (exitValue != 0) {
+				// cmake had errors...
+				String msg = String.format(Messages.CMakeBuildConfiguration_ExitFailure, arg0, exitValue);
+				addMarker(srcFolder.getProject(), -1, msg, IMarkerGenerator.SEVERITY_ERROR_BUILD, null);
+				return Status.error(msg);
+			}
+		}
+		cmakeListsModified = false;
+		// parse compile_commands.json file
+		getCompileCommandsFile().refreshLocal(IResource.DEPTH_ZERO, subMonitor.split(1));
+		processCompileCommandsFile(console, subMonitor.split(1));
+		return Status.OK_STATUS;
 	}
 
 	@Override
--- a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
+++ b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
@@ -268,7 +268,7 @@ public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMa
 	/**
 	 * @since 2.1
 	 */
-	protected IStatus configureCMakeBuildFiles(IProgressMonitor monitor) throws CoreException, IOException {
+	public IStatus configureCMakeBuildFiles(IProgressMonitor monitor) throws CoreException, IOException {
 		IProject project = getProject();
 		project.deleteMarkers(ICModelMarker.C_MODEL_PROBLEM_MARKER, false, IResource.DEPTH_INFINITE);
 		// Setup console
--- a/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
+++ b/cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
@@ -206,8 +206,7 @@ public class CMakeBuildConfiguration extends CBuildConfiguration implements ICMa
 			}
 
 			// parse compile_commands.json file
-			getCompileCommandsFile().refreshLocal(IResource.DEPTH_ZERO, monitor);
-			processCompileCommandsFile(console, monitor);
+			refreshAndProcessCompileCommandsFile(console, monitor);
 
 			infoStream.write(String.format(Messages.CMakeBuildConfiguration_BuildingIn, buildDir.toString()));
 			// run the build tool...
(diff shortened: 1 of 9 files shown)


DangMinhTam382:
Hi Mr. @betamaxbandit ,
CC: Mr. @Kummallinen ,

Before making this PR official, could you take a quick look and let me know what you think?

FYI, it built failed due to the removal of `IMG_DLCL_COPY_VIEW_TO_CLIPBOARD` in [IInternalDebugUIConstants.class](https://github.com/eclipse-platform/eclipse.platform/blob/master/debug/org.eclipse.debug.ui/ui/org/eclipse/debug/internal/ui/IInternalDebugUIConstants.java).
Fix for this build should be included in #1485

Thanks.

betamaxbandit:
> Hi Mr. @betamaxbandit , CC: Mr. @Kummallinen ,
> 
> Before making this PR official, could you take a quick look and let me know what you think?

Hi @DangMinhTam382 ,
Thanks for this. I've started reviewing this, but need some more time. I'll probably finish the review tomorrow (Fr).
Cheers John

DangMinhTam382:
Hi Mr. @betamaxbandit ,

I have made some modification to allow manual testing for this PR.
This is now ready to be review now!

Many thanks,
Tam.

cmake/org.eclipse.cdt.cmake.ui.tests/manualTests/README.md
```diff
@@ -41,6 +41,16 @@ Note, the Build Settings tab settings are stored separately for Run mode and Deb
 #### 8.1) Perform build, clean and open build settings
   Expected: Settings are persisted
 
+### 9) CMake configuration
+
+Verifies that API for configuring CMake could work individually.
```
betamaxbandit: Verifies that the API for configuring CMake can be invoked independently.

cmake/org.eclipse.cdt.cmake.ui.tests/manualTests/README.md
```diff
@@ -41,6 +41,16 @@ Note, the Build Settings tab settings are stored separately for Run mode and Deb
 #### 8.1) Perform build, clean and open build settings
   Expected: Settings are persisted
 
+### 9) CMake configuration
+
+Verifies that API for configuring CMake could work individually.  
+Note, API could only be test in Developer environment with plug-in **org.eclipse.cdt.cmake.example** included.
```
betamaxbandit: Note: this API can only be tested in a CDT development environment that includes the **org.eclipse.cdt.cmake.example** plug-in.

cmake/org.eclipse.cdt.cmake.ui.tests/manualTests/README.md
```diff
@@ -41,6 +41,16 @@ Note, the Build Settings tab settings are stored separately for Run mode and Deb
...
 
+### 9) CMake configuration
+
+Verifies that API for configuring CMake could work individually.  
+Note, API could only be test in Developer environment with plug-in **org.eclipse.cdt.cmake.example** included.  
+
+  1. Remove existing "**/build/**" folder.
+  2. Right click project > select "Configure CMake Project".  
+
+Expected: CMake Configuration process start with active launch settings.
```
betamaxbandit: 1. Remove any existing `**/build/**` folder.
  2. Right-click the project and select **Configure CMake Project**.

Expected: The CMake configuration process starts using the active launch settings.

betamaxbandit (changes requested):
LGTM, just a few minor grammar issues.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
```diff
@@ -296,6 +265,65 @@ public IProject[] build(int kind, Map<String, String> args, IConsole console, IP
 		}
 	}
 
+	/**
+	 * @since 2.1
+	 */
+	protected IStatus configureCMakeBuildFiles(IProgressMonitor monitor) throws CoreException, IOException {
```
betamaxbandit: This helper now refreshes and processes compile_commands.json, but build() continues to do the existing compile_commands.json refresh/processing immediately afterwards. So a normal build that needs to run CMake appears to parse compile_commands.json twice.

Could we split this into two helpers, e.g. one method that only runs the CMake configure command and another that refreshes/processes compile_commands.json? Then build() can keep its existing single compile_commands.json processing step, while the public configure-only API can call both helpers.
Something like ...
````
private IStatus runCMakeConfigure(CommandDescriptorBuilder cmdBuilder, IConsole console,
		ConsoleOutputStream infoStream, IProgressMonitor monitor)
		throws CoreException, IOException {
	// run CMake only
}

private void refreshAndProcessCompileCommandsFile(IConsole console, IProgressMonitor monitor)
		throws CoreException {
	getCompileCommandsFile().refreshLocal(IResource.DEPTH_ZERO, monitor);
	processCompileCommandsFile(console, monitor);
}
````

File: cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java [hunk 5]
```diff
@@ -296,6 +265,65 @@ public IProject[] build(int kind, Map<String, String> args, IConsole console, IP
...
+		IProject project = getProject();
+		project.deleteMarkers(ICModelMarker.C_MODEL_PROBLEM_MARKER, false, IResource.DEPTH_INFINITE);
```
betamaxbandit: I do not think the configure-only API should delete all CDT problem markers. A manual CMake configure does not run the compiler, so clearing previous compiler diagnostics could hide real errors even though no build has replaced them.

The private helper already clears CMake execution markers via deleteCMakeErrorMarkers(getProject()), which seems appropriate. I suggest removing this broader ICModelMarker.C_MODEL_PROBLEM_MARKER deletion from the configure-only API and leaving broad marker cleanup to the full build path only.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
```diff
@@ -296,6 +265,65 @@ public IProject[] build(int kind, Map<String, String> args, IConsole console, IP
...
+		try (CMakeErrorParser errorParser = new CMakeErrorParser(new CMakeExecutionMarkerFactory(srcFolder))) {
+			ParsingConsoleOutputStream errStream = new ParsingConsoleOutputStream(console.getErrorStream(),
+					errorParser);
+			IConsole errConsole = new CMakeConsoleWrapper(console, errStream);
+			Process p = startBuildProcess(command.getArguments(), new IEnvironmentVariable[0], workingDir, errConsole,
+					subMonitor.split(1));
+			String arg0 = command.getArguments().get(0);
+			if (p == null) {
+				// process start failed
+				String msg = String.format(Messages.CMakeBuildConfiguration_Failure, ""); //$NON-NLS-1$
```
betamaxbandit: Since this message is now returned to callers as an IStatus, it would be useful to include the command that failed to start. Could this use arg0 instead of the empty string?
something like ...
````
String msg = String.format(Messages.CMakeBuildConfiguration_Failure, arg0);
````

File: cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java [hunk 7]
```diff
@@ -296,6 +265,65 @@ public IProject[] build(int kind, Map<String, String> args, IConsole console, IP
 		}
 	}
 
+	/**
+	 * @since 2.1
+	 */
```
betamaxbandit, line 287: Can you add a good comment seeing as this is going to be API.
Something like ...

````
/**
 * Runs the CMake configure step for this build configuration without invoking
 * the build target. On success, refreshes and processes compile_commands.json
 * so scanner information is updated.
 * <p>
 * This method writes to the CDT build console, deletes stale CMake execution
 * markers, may create new CMake execution markers, and updates scanner
 * information for this build configuration. Callers should run this from a
 * background workspace operation, not directly from the UI thread.
 *
 * @param monitor progress monitor, or {@code null}
 * @return {@link Status#OK_STATUS} if CMake completed successfully; otherwise
 *         an error status if the CMake process could not be started or exited
 *         non-zero
 * @throws CoreException if workspace refresh, marker handling, or scanner-info
 *         processing fails
 * @throws IOException if console/process I/O fails
 * @since 2.1
 */
public IStatus configureCMakeBuildFiles(IProgressMonitor monitor)
		throws CoreException, IOException {
````

File: cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java [hunk 8]
```diff
@@ -195,45 +199,10 @@ public IProject[] build(int kind, Map<String, String> args, IConsole console, IP
...
-
-					// check cmake exit status
-					final int exitValue = watchProcess(errConsole, monitor);
-					if (exitValue != 0) {
-						// cmake had errors...
-						String msg = String.format(Messages.CMakeBuildConfiguration_ExitFailure, arg0, exitValue);
-						addMarker(srcFolder.getProject(), -1, msg, IMarkerGenerator.SEVERITY_ERROR_BUILD, null);
-						return null;
-					}
+				IStatus result = configureCMakeBuildFiles(cmdBuilder, console, infoStream, monitor);
```
betamaxbandit, line 202: This helper now refreshes and processes compile_commands.json, but build() continues to do the existing compile_commands.json refresh/processing immediately afterwards. So a normal build that needs to run CMake appears to parse compile_commands.json twice.

Could we split this into two helpers, e.g. one method that only runs the CMake configure command and another that refreshes/processes compile_commands.json? Then build() can keep its existing single compile_commands.json processing step, while the public configure-only API can call both helpers.

Something like ...
````
private IStatus runCMakeConfigure(CommandDescriptorBuilder cmdBuilder, IConsole console,
		ConsoleOutputStream infoStream, IProgressMonitor monitor)
		throws CoreException, IOException {
	// run CMake only
}

private void refreshAndProcessCompileCommandsFile(IConsole console, IProgressMonitor monitor)
		throws CoreException {
	getCompileCommandsFile().refreshLocal(IResource.DEPTH_ZERO, monitor);
	processCompileCommandsFile(console, monitor);
}
````

File: cmake/org.eclipse.cdt.cmake.example/src/org/eclipse/cdt/cmake/example/handler/ConfigureExtendedCMakeProjectHandler.java [hunk 9]
```diff
@@ -0,0 +1,64 @@
...
+				WorkspaceJob job = new WorkspaceJob("Configuring Extended CMake Project...") { //$NON-NLS-1$
+					@Override
+					public IStatus runInWorkspace(IProgressMonitor monitor) throws CoreException {
+						try {
+							return cbc.configureCMakeBuildFiles(monitor);
+						} catch (CoreException | IOException e) {
+							ILog.of(ConfigureExtendedCMakeProjectHandler.class)
+									.error("Failed to configure for extended CMake Project", e); //$NON-NLS-1$
+						}
+						return null;
```
betamaxbandit: runInWorkspace should return an IStatus. Returning null after logging the exception is risky and means the Jobs framework does not receive a proper failure status.

Could this return the CoreException status, or an error status for IOException?

Something like  ...
````
} catch (CoreException e) {
	return e.getStatus();
} catch (IOException e) {
	ILog.of(ConfigureExtendedCMakeProjectHandler.class)
			.error("Failed to configure for extended CMake Project", e); //$NON-NLS-1$
	return Status.error("Failed to configure for extended CMake Project", e); //$NON-NLS-1$
}
````

File: cmake/org.eclipse.cdt.cmake.example/src/org/eclipse/cdt/cmake/example/handler/ConfigureCMakeProjectHandler.java [hunk 10]
```diff
@@ -0,0 +1,64 @@
...
+						try {
+							return cbc.configureCMakeBuildFiles(monitor);
+						} catch (CoreException | IOException e) {
+							ILog.of(ConfigureExtendedCMakeProjectHandler.class)
+									.error("Failed to configure for extended CMake Project", e); //$NON-NLS-1$
+						}
+						return null;
+					}
+				};
+				job.schedule();
```
betamaxbandit, line 59: This job modifies project state: markers, resource refresh, and scanner information. Should it set a scheduling rule before scheduling, probably the project, to avoid running concurrently with other workspace operations on the same project?
Suggestion ...
````
job.setRule(project);
job.schedule();
````

File: cmake/org.eclipse.cdt.cmake.example/plugin.xml [hunk 11]
```diff
@@ -52,4 +52,54 @@
...
+         </command>
+      </menuContribution>
+   </extension>
+   <extension
+         id="org.eclipse.cdt.cmake.example.configureCMake"
+         name="Configure Project"
+         point="org.eclipse.ui.commands">
+      <command
+            id="org.eclipse.cdt.cmake.example.configureCMakeProject"
+            name="Configure Extended CMake Project">
```
betamaxbandit: The menu label says "Configure CMake Project" but the command name says "Configure Extended CMake Project". Could these be made consistent? Unless "Extended" has a specific meaning here, I suggest using "Configure CMake Project" in both places.

betamaxbandit (changes requested):
Hi,
this is looking really good.
There's a few minor issues that should be resolved please.

cmake/org.eclipse.cdt.cmake.ui.tests/manualTests/README.md
DangMinhTam382: Thanks for the suggestion! I updated the README.md file

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
DangMinhTam382: Ahhh, thanks your suggestion.
That's a big help for me.

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
DangMinhTam382: Ahh, you're right there!
It does parse `compile_commands.json` file twice.

Code updated as your suggestion!

cmake/org.eclipse.cdt.cmake.example/src/org/eclipse/cdt/cmake/example/handler/ConfigureCMakeProjectHandler.java
DangMinhTam382: I'm with you on this one.

I'm not quite sure if there's another WS job running during CMake configuration process, so I'll play it safe here and set rule to it as suggested.

cmake/org.eclipse.cdt.cmake.ui.tests/manualTests/README.md
DangMinhTam382: Thanks for the suggestion! I updated the README.md file

cmake/org.eclipse.cdt.cmake.ui.tests/manualTests/README.md
DangMinhTam382: Thanks for the suggestion! I updated the README.md file

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
DangMinhTam382: Agreed!
Code updated, no longer remove `ICModelMarker.C_MODEL_PROBLEM_MARKER` when only configure CMake

cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java
DangMinhTam382: I had some thought on this one.

Since the condition is checking for process starting error, instead of using `arg0`, the message is updated to
`Failure running cmake: Process failed to start`

Error related to `arg0 `with be check/log in [the next condition](https://github.com/eclipse-cdt/cdt/pull/1483/changes#diff-5a7a408bc911439b9da6debb31452221baadcf42a4051f27cae11f22876ea700R333)
```
			if (exitValue != 0) {
				// cmake had errors...
				String msg = String.format(Messages.CMakeBuildConfiguration_ExitFailure, arg0, exitValue);
				addMarker(srcFolder.getProject(), -1, msg, IMarkerGenerator.SEVERITY_ERROR_BUILD, null);
				return Status.error(msg);
			}
```
WDT?

cmake/org.eclipse.cdt.cmake.example/src/org/eclipse/cdt/cmake/example/handler/ConfigureExtendedCMakeProjectHandler.java
DangMinhTam382: Ahh, thanks for pointing this out and the suggestion as well.

I believe the `WorkspaceJob` does log the error if an error status is return.
So, I grouped them together and return a simple
```Status.error(message, exception)```

WDT?

File: cmake/org.eclipse.cdt.cmake.example/plugin.xml (hunk 11 above)
Comment (DangMinhTam382) [reply cmake/org.eclipse.cdt.cmake.example/plugin.xml]: Ah, yes, the initial intent was to allow CMake configuration only for Extended CMake projects.
Later, this was changed to support all CMake projects, and these spots were somehow missed.

Updated the UI and class names to align with the current behavior.

File: cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java (hunk 4 above)
Comment (DangMinhTam382) [reply cmake/org.eclipse.cdt.cmake.core/src/org/eclipse/cdt/cmake/core/CMakeBuildConfiguration.java]: Ahh, you're right there!
It does parse compile_commands.json file twice.

Code updated as your suggestion!

**Review by DangMinhTam382 — COMMENTED** [review]
Hi Mr. @betamaxbandit ,

I fixed your comments and this should be ready to be reviewed again

Many thanks,
Tam

**Review by betamaxbandit — APPROVED** [review]
LGTM
Thanks Tam for reworking this.
I think it's now ready to merge.

File: cmake/org.eclipse.cdt.cmake.example/META-INF/MANIFEST.MF [hunk 12]
```diff
@@ -2,7 +2,7 @@ Manifest-Version: 1.0
 Bundle-ManifestVersion: 2
 Bundle-Name: %pluginName
 Bundle-SymbolicName: org.eclipse.cdt.cmake.example;singleton:=true
-Bundle-Version: 1.0.0.qualifier
+Bundle-Version: 1.0.1.qualifier
```
jld01: Should be 1.0.100

jld01 (changes requested):

cmake/org.eclipse.cdt.cmake.example/META-INF/MANIFEST.MF
DangMinhTam382: Hi Mr. @jld01 ,

Updated the micro version to 100 as suggested.
Just for my understanding, is "increasing micro by 100" always applied? or is it just temporary?

DangMinhTam382:
>[ERROR] download from https://download.eclipse.org/tools/cdt/releases/12.5/cdt-12.5.0/plugins/org.eclipse.cdt.managedbuilder.llvm.ui_1.4.700.202603302001.jar failed: Server returned HTTP code: 502 for URL https://download.eclipse.org/tools/cdt/releases/12.5/cdt-12.5.0/plugins/org.eclipse.cdt.managedbuilder.llvm.ui_1.4.700.202603302001.jar

Looks like [Code Cleanliness Checks / build (pull_request)](https://github.com/eclipse-cdt/cdt/actions/runs/30968238581/job/92186775496?pr=1483) built failed due to server issue?
Will try to rebuild latter.

cmake/org.eclipse.cdt.cmake.example/META-INF/MANIFEST.MF
jld01: Thank you for the update, @DangMinhTam382. The CDT project follows the same [versioning rules](https://github.com/eclipse-platform/eclipse.platform/blob/master/docs/VersionNumbering.md) as Eclipse Platform. The bundle micro version is updated by 1 for occasional CDT bugfix releases from an existing release branch (such as CDT 11.6.1). For the regular minor releases (where a new release branch will be made shortly before release), we update the so-called _service segment_.

jld01 (commented):
Bundle version looks good now.

jld01 (approved):

DangMinhTam382:
Thanks Mr. @betamaxbandit and Mr. @jld01 for reviewing this.

Mr. @jonahgraham, could you have a look and merge this one for me?
Also, there are 5 commits in this PR, do they need to be squash into 1 before merged?

jonahgraham (approved):
> Mr. @jonahgraham, could you have a look and merge this one for me? 

Approved - I will merge momentarily.

> Also, there are 5 commits in this PR, do they need to be squash into 1 before merged?

I will do a squash and merge this time. (There is some inconsistency on what expectations in this area are - there are advantages to leaving individual commits for reviewers to easily see what has changed, and there is advantages to squashing yourself so that exactly the correct commit and message is submitted to main branch)

